PR #8, opened 2019-01-04
add paypal as a user

kentcdodds:
That was easy!

image


website/siteConfig.js
```diff
@@ -10,14 +10,12 @@
...
-  //   // if it is not '/', like: '/test-site/img/docusaurus.svg'.
-  //   image: '/img/docusaurus.svg',
-  //   infoLink: 'https://www.facebook.com',
-  //   pinned: true,
-  // },
+  {
+    caption: 'PayPal',
+    image: '/img/users/paypal.svg',
+    infoLink: 'https://www.paypal.com',
+    pinned: true,
```
kentcdodds: Note: I think I'd prefer well-known/established companies be "pinned" in the long-term.

PR #9, opened 2019-01-04
Improved the selected navigation item indicator, fixes #7

### LINKED ISSUE #7: Make the current page more obvious
Right now it's a little hard to tell which page I'm currently on.

image

Do you think we could make the current page red (maybe the same color as the octopus?)

image

Though that's probably not very accessible, so maybe underlined as well?

image

But that may not be enough either... What about an arrow indicator?

image

Or... maybe... even better... 

image

😂

weyert:


website/static/css/custom.css
```diff
@@ -37,6 +37,21 @@ a {
...
+}
+
+.navListItemActive a {
+    color: red;
+}
+
+.navListItem.navListItemActive:before {
+  padding-left: 0.5rem;
+  padding-right: 0.25rem;
+  content: '🐐';
```
kentcdodds: :joy:

Let's go with 🐙 I think. That'd make more sense for everything I think.

kentcdodds:
Here's what it looks like now:

image


I think I'd like it better this way (so the title link doesn't move):

image

This would need:

```css
.container .docsNavContainer {
  padding-left: 12px
}
.navListItem.navListItemActive:before {
  content: "🐙";
  position: absolute;
  left: -8px;
}
```

What do you think?

website/static/css/custom.css
weyert: Fixed!

PR #10, opened 2019-01-04
add tiller as a user

spartDev:
I had to add a `width: auto` to showcase logo image to avoid having a crapy tiny logo

screenshot 2019-01-04 at 21 32 06


website/static/css/custom.css
```diff
@@ -60,13 +60,19 @@ a {
   left: -8px;
 }
 
+/* Change styling of logos in showcaseSection */
+.productShowcaseSection .logos img,
+.showcaseSection .logos img {
+  width: auto;
```
kentcdodds: I'm :+1: about this

PR #12, opened 2019-01-06
Add Autodesk as a user

matchai:
Thank you for the excellent work documenting the testing-library suite! 😄 
Added a bit of CSS to ensure that wide but short logos don't get too big.

Cheers!


alexkrolick:
Great!

PR #13, opened 2019-01-07
docs: updated the input events doc

weyert:
Called out the user-event library in the input events documentation section


weyert:
Sorry, I will have a look how I can link to a different page

weyert:
Done @alexkrolick

PR #16, opened 2019-01-08
Add Sweepbright to the list

krvajal:
We at Sweepbright are also using react-testing-library and it have changed the way we write tests. This PR adds our company to the list of companies that used `*-testing-library`.


website/siteConfig.js
```diff
@@ -34,6 +34,12 @@ const users = [
     infoLink: 'https://www.autodesk.com',
     pinned: true,
   },
+   {
+    caption: 'Sweepbright',
+    image: '/img/users/sweepbright.svg',
```
kentcdodds: I think you may have forgotten to add this SVG.

krvajal:
@kentcdodds I updated the PR with the icon

PR #18, opened 2019-01-12
Add Velo Payments as a user

robcaldecottvelo:
`react-testing-library` has had an enormous impact at Velo Payments.


alexkrolick:
Awesome!

PR #22, opened 2019-01-28
setupTestFrameworkScriptFile => setupFilesAfterEnv

Vpr99:
With the Jest 24 release, they've deprecated `setupTestFrameworkScriptFile` in favor of [`setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array), so I thought it would be helpful to update the docs to match.


alexkrolick:
We should note how to do it in both Jest <23 and 24, because it's a breaking change

Vpr99:
Sounds great. Updating now!

Vpr99:
@kentcdodds super cool! Updated the PR with docs for Jest 23 and 24.

PR #26, opened 2019-02-05
Fix link to the Update Props page

arthurdenner:
Hi! :wave: 

Currently, the `See examples page` [here](https://testing-library.com/docs/react-testing-library/api#rerender) is linking to the wrong document. This PR fixes it.

Cheers and thanks for the docs. :smiley:


alexkrolick:
Thanks ✔️

PR #28, opened 2019-02-09
add Parsley Health to users

phucanhapril:
[We](https://www.parsleyhealth.com)'re using react-testing-library for our internal tools. Thanks for all the hard work y'all!


alexkrolick:
Cool!

PR #36, opened 2019-02-15
feat: add pptr-testing-library docs

patrickhulce:
Adds docs for pptr-testing-library

ref https://github.com/patrickhulce/pptr-testing-library/issues/9

New homepage look
image

Puppteer page look
image


website/pages/en/pptr.js
```diff
@@ -0,0 +1,159 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const React = require('react')
```
patrickhulce: this was pretty much just copied from the others, I think that's what we're doing here? 🤷‍♂️

alexkrolick: Yeah, but a homepage is optional. You can link straight to the doc page from the root.

alexkrolick:
Thank you!

PR #45, opened 2019-02-27
Add Facebook Open Source as user

JoelMarcey:
screenshot 2019-02-26 19 00 48


JoelMarcey:
I was not going to be the presumptuous 😄

kentcdodds:
I'd there a reason the logo is Facebook open source and not just Facebook itself? I'm pretty sure product teams are using it at Facebook.

JoelMarcey:
I debated this actually, and I made this as a conscious decision. Because I help run our open source program at Facebook, and believe the program helped enable projects like React to use your great project, I went that route. Nothing nefarious - just wanted to tout our open source program.

Is that a big sticking point for you?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kentcdodds:
Not a big sticking point, but I would prefer the more readily recognizable and arguably more official looking facebook logo if you're amiable :)

JoelMarcey:
Ok. I will look into doing that. One thing I might ask - why not both? 😄 Maybe merge this and then I will see about sending a PR with the Facebook logo too?

kentcdodds:
Sounds good to me! What do you think Alex?

alexkrolick:
I was going to suggest both, too 😉

PR #58, opened 2019-03-20
Pr/docs React-Intl

viniciusavieira:
Hello, I'm not sure if this is the correct way to contribute to this project, but I've created an example on react-testing-library repo, and I realized that it should be probably be updated here too.

So I basically replicated here the example I've created on the examples folder. 

Reference: https://github.com/kentcdodds/react-testing-library/pull/329


docs/example-react-intl.md
```diff
@@ -0,0 +1,51 @@
...
+        month="2-digit"
+        year="numeric"
+      />
+    </div>
+  )
+}
+
+const renderWithReactIntl = component => {
+  return {
+    ...render(<IntlProvider locale="pt">{component}</IntlProvider>),
```
alexkrolick: Nit: do you need the `...` here? I think returning the result of render is fine.

docs/example-react-intl.md
```diff
@@ -0,0 +1,51 @@
...
+
+```jsx
+import React from 'react'
+import 'jest-dom/extend-expect'
+import { render, cleanup, getByTestId } from 'react-testing-library'
+import { IntlProvider, FormattedDate } from 'react-intl'
+import IntlPolyfill from 'intl'
+import 'intl/locale-data/jsonp/pt'
+
+const setupTests = () => {
```
alexkrolick: Can this link to the setup section that discusses setting up globals?

alexkrolick:
Thanks, a few small comments.

docs/example-react-intl.md
viniciusavieira: You are absolutely correct, it can be removed. I'll update it.

I have a codesandbox with a functional example of it here: https://codesandbox.io/s/yq2jz85jx1

docs/example-react-intl.md
viniciusavieira: Move the setup description of react-intl to the /docs/react-testing-library/setup or the /docs/setup?

alexkrolick: Just add a link here to `/docs/react-testing-library/setup` to clarify how this can be combined with other global setup functions

viniciusavieira: I've done that and created a section at the setup with a small explanation about why is necessary to load locales.
Do you think it's ok?

alexkrolick: Can you move the description in setup to this file instead to keep it all in one place, and keep setup generic?

viniciusavieira: Done.

alexkrolick:
Cool thanks!

viniciusavieira:
Glad to help!

PR #59, opened 2019-03-21
document the published module

### LINKED ISSUE #56: Document bundle types
Document UMD, CJS, ESM exports and when to use them

https://github.com/kentcdodds/dom-testing-library/issues/212

kentcdodds:
Closes #56

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -26,6 +26,28 @@ wrapper:
 - [user-event](ecosystem-user-event.md) browser event simulation
 - [jest-dom](ecosystem-jest-dom.md) custom Jest matchers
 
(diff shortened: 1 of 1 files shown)


kentcdodds:
I tried to write this to be evergreen in case we change some of the configuration in the future. So I referenced the `package.json` found on unpkg and invited people to look at that rather than write out the files directly in the docs.

kentcdodds:
Hmmm... Hadn't considered that. I don't think that I'd do it personally, but we could 🤷‍♂️

alexkrolick:
Will look into it later

Iterate

PR #60, opened 2019-03-21
✨ Add ExpediaGroup as a user ✈️

sirMerr:
I forgot I had this forked, finally adding ✈️ to the list! Thank you for all the great work 🙌. Slowly but surely converting everyone here to the guiding principle...

### Preview
Screen Shot 2019-03-21 at 1 48 46 PM


website/siteConfig.js
```diff
@@ -64,6 +64,11 @@ const users = [
     infoLink: 'https://opensource.facebook.com',
     pinned: true,
   },
+  {
+    caption: 'Expedia Group',
+    image: '/img/users/expediagroup.svg',
+    infoLink: 'https://www.expediagroup.com',
```
alexkrolick: you can add `pinned` here, we'll sort out the pin rotation later

alexkrolick:
💥 cool

PR #64, opened 2019-03-22
Updated react hooks example page to match react-hooks-testing-library's documentation

ab18556:


mpeyper:
@alexkrolick, I think you can close this now. `react-hooks-testing-library` recently merged [a change to stop using `react-testing-library` for rendering](https://github.com/mpeyper/react-hooks-testing-library/pull/40), so I'm not sure it's still appropriate for it to appear in these docs.

We still follow the same principals, but the say we're part of the `testing-library` ecosystem is not technically true anymore.

alexkrolick:
I don't think not having a hard dependency on RTL is a requirement; I would say it belongs in the same "Ecosystem" category as jest-dom and native-testing-library -- libs that share the overall approach and complementary features.

https://testing-library.com/blog/2019/04/25/new-org

kentcdodds:
Agreed Alex.

mpeyper:
Ok cool.  I'm happy for this to continue then :)

bcarroll22:
I also let @mpeyper know that the change to no longer require `react-testing-library` was actually a huge help to native because now we can leverage his library as well for writing hooks tests 🎉

kentcdodds:
Oh nice!

alexkrolick:
I'm not sure about the current state of this library - is this PR accurate or do we need a reboot?

mpeyper:
The library is still active, and the examples shown here are still accurate (our API hasn't changed much), but some of the descriptions and links are not quite right anymore. We aren't using react-testing-library any more and the links point to the the old location before it was brought into the testing-library org.

docs/example-react-hooks.md
```diff
@@ -3,9 +3,7 @@ id: example-react-hooks
 title: React Hooks
 ---
 
-`react-testing-library` provides the
-[`testHook`](/docs/react-testing-library/api#testhook) utility to test custom
-hooks.
+[`react-hooks-testing-library`](https://github.com/mpeyper/react-hooks-testing-library) is built on top of [`react-testing-library`](/react) to create a simple test harness for [React hooks](https://reactjs.org/docs/hooks-intro.html) that handles running them within the body of a function component, as well as providing various useful utility functions for updating the inputs and retrieving the outputs of your [custom hook](https://reactjs.org/docs/hooks-custom.html).
```
rawat07102, line 6: we should replace this description with react-hooks-testing-library ["The solution"](https://github.com/testing-library/react-hooks-testing-library#the-solution) section with an updated link for [the react-hooks-testing-library](https://github.com/testing-library/react-hooks-testing-library) and [react-testing-library](https://github.com/testing-library/react-testing-library)

PR #76, opened 2019-04-10
add link to api-configuration in the sidebar

saltycrane:


alexkrolick:
good catch, thanks again!

PR #87, opened 2019-04-25
docs: add blog post about new stuff

kentcdodds:


kentcdodds:
I think we should add a section about native-testing-library. @bcarroll22, do you want to do that?

bcarroll22:
Sorry @kentcdodds I just saw this. I see you've got a section in there about it now, do you still want me to add anything to it? I can't really think of anything, the only thing I could maybe see being helpful is the technical deep dive article I wrote a while back?

kentcdodds:
I'll add the link. Thanks @bcarroll22!

kentcdodds:
Sweet. This is ready to go!

PR #89, opened 2019-04-27
update siteconfig with new org info

alexkrolick:


alexkrolick:
preview looks good to me I think...

PR #91, opened 2019-04-29
Update NTL docs

bcarroll22:
This is a first pass at getting more about NTL and its ecosystem on the main docs site. I also tried to change as much as I could to point to the new org and remove "DOM" stuff from the main landing page.

Let me know if I missed anything. I think this probably gets enough on the main docs site to make it feel "first class" but for now I still think its beneficial to leave the native site up for more comprehensive docs on native.


website/pages/en/index.js
```diff
@@ -105,9 +105,6 @@ class Index extends React.Component {
               the more confidence they can give you.
             </i>
           </p>
-          <MarkdownBlock>
```
alexkrolick: It would be cool to have this animate a transition between the different libraries

website/pages/en/index.js
```diff
@@ -138,7 +135,7 @@ class Index extends React.Component {
             imageAlign: 'right',
             imageAlt: 'The solution (picture of a star)',
             content:
-              "## The Solution \n `dom-testing-library` is a very light-weight solution for testing DOM nodes (whether simulated with [JSDOM](https://github.com/jsdom/jsdom) in [Jest](https://jestjs.io) or in the browser). The main utilities it provides involve querying the DOM for nodes in a way that's similar to how the user finds elements on the page. In this way, the library helps ensure your tests give you confidence in your UI code.",
+              '## The Solution \n Testing Library is a very light-weight solution for testing without all the implementation details. The main utilities it provides involve querying for nodes similarly to how users would find them. In this way, testing-library helps ensure your tests give you confidence in your UI code.',
```
alexkrolick: I think this is a bit too generic since `testing-library` isn't something you can install. How about rewording to reference "the testing-library family of libraries" or something like that?

website/pages/en/index.js
```diff
@@ -218,9 +215,9 @@ class Index extends React.Component {
             title: '[Puppeteer Testing Library](./pptr)',
           },
           {
-            image: `${baseUrl}img/construction-128x128.png`,
+            image: `${baseUrl}img/react-128x128.png`,
             imageAlign: 'top',
-            title: '[And more...](./docs/ecosystem-user-event)',
+            title: '[Native Testing Library](./react-native)',
```
alexkrolick: Can you leave the ecosystem link here and make a standard thumbnail for react native?

bcarroll22: Yep I can leave the ecosystem one in there. 

Disclosure, the reason I took it out was that it was the last non-implementation package so it seemed out of place, and it looks maaad weird when there's two items in this last row. Do you think there's one more thing we could put in this table so it's all 3 across?

Also react native and react have the same logo weirdly. Do you want me to just copy the react logo and name it react-native?

bcarroll22:
I'd gravitate toward option 2, I'll be honest until I did this I don't think I even knew the landing pages existed 😬

website/pages/en/index.js
alexkrolick: You can change it to 2, 3, or 4 columns, might look better

alexkrolick: Looks like the RN site uses a white-on-black React logo, how about using a black one for RN?

website/pages/en/index.js
alexkrolick: Can do this as a followup

website/pages/en/index.js
bcarroll22: I'll try to do this this evening, as part of it do you want me to remove the link to each landing page? I think we should probably delete the pages separately because some of the libraries (at least RTL) link to their landing page from their readme.

alexkrolick: You can place redirects in netlify.toml from the landing pages to the docs pages

bcarroll22:
Alrighty, react-native thumbnail added, ecosystem link added back. landing pages removed, and netlify redirects set up. I think we're good to go on this one 👍

website/pages/en/index.js
```diff
@@ -181,43 +181,47 @@ class Index extends React.Component {
...
 
     const Ecosystem = () => (
-      <Block layout="threeColumn" background={null}>
+      <Block layout="fourColumn" background={null}>
         {[
           {
             image: `${baseUrl}img/react-128x128.png`,
             imageAlign: 'top',
-            title: '[React Testing Library](./react)',
+            title: 'React Testing Library',
```
alexkrolick: Wait, let's not remove the links, just point them to the intro.md pages?

netlify.toml
```diff
@@ -40,3 +40,38 @@
   to = "/docs/dom-testing-library/:splat"
   status = 301
   force = false # do not redirect if route is not 404
+
+# React landing page to home page
+[[redirects]]
+  from = "https://testing-library.com/"
```
alexkrolick: Ha, not quite what I meant 😉 

```toml
# Cypress landing page to home page
[[redirects]]
  from = "https://testing-library.com/cypress/"
  to = "https://testing-library.com/docs/cypress-testing-library/intro"
  status = 301
  force = true
```

alexkrolick:
@allcontributors[bot] please add @bcarroll22 for code

bcarroll22:
Addressed 👍

alexkrolick:
This is awesome

PR #97, opened 2019-04-30
Added Testsigma to the list of companies

renjujv:
Added Testsigma to the list of companies and also uploaded the logo for display.


alexkrolick:
Neat!

PR #104, opened 2019-05-03
add Svelte docs & a known issue in testcafe.

benmonro:


benmonro:
@alexkrolick updated w/ logos for test cafe & svelte.  Thanks for the heads up, i didn't even see it below the scoll.

PR #105, opened 2019-05-05
add documentation for waitforElementToBeRemoved (#103)

Pklong:
Addresses #103 

This is my first open-source contribution in many years so please let me know if there's anything I can do to improve. Thanks for the opportunity!


docs/dom-testing-library/api-async.md
```diff
@@ -187,3 +187,69 @@ The default `timeout` is `4500ms` which will keep you under
...
+[`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
+is more efficient than polling the DOM at regular intervals with `wait`. This
+library sets up a
+[`'mutationobserver-shim'`](https://github.com/megawac/MutationObserver.js) on
+the global `window` object for cross-platform compatibility with older browsers
+and the [`jsdom`](https://github.com/jsdom/jsdom/issues/639) that is usually
+used in Node-based tests.
+
+The default `container` is the global `document`. Make sure the elements you
+wait for will be attached to it, or set a different `container`.
```
kentcdodds: Would be good to reword this a bit because right now it sounds like it should be a direct child.

docs/guide-disappearance.md
```diff
@@ -43,6 +43,21 @@ test('movie title goes away', async () => {
 })
 ```
 
+The `waitForElementToBeRemoved` [async helper][async-api] function uses a
```
kentcdodds: Let's move this to be the first example in this section (before the `wait` example) because it's probably the better option.

docs/dom-testing-library/api-async.md
```diff
@@ -187,3 +187,69 @@ The default `timeout` is `4500ms` which will keep you under
...
+  .catch(err => console.log(`Error you need to deal with: ${err}`))
+
+el.setAttribute('data-neat', true)
+// other mutations are ignored...
+
+el.parentElement.removeChild(el)
+// logs 'Element no longer in DOM'
+```
+
+The callback must return the pre-existing element or array of elements that are expected to be removed.
```
alexkrolick: I think the explanation that the element(s) must be in the DOM when you call the function should be at the top

Pklong: No problem, I'll make that change now!

PR #115, opened 2019-05-21
Learning Material - My Experience moving from Enzyme to react-testing-library

boyney123:
Adding a new learning resource for developers to understand a point of view coming from years of experience with Enzyme going onto react-testing-library.


boyney123:
Thanks @kentcdodds 🙇 👍

PR #117, opened 2019-05-22
Fix `getByAltText` examples

### LINKED ISSUE #116: ByAltText examples seem incorrect
**Describe the bug**
The example for `ByAltText` uses the following example markup:

```html
<img alt="Incredibles 2 Poster" src="/incredibles-2.png" />
```

The examples pass `/incredibles.*png$/i` to `getByAltText`, which seems incorrect. If we're trying to find the element by the value of the `alt` attribute, why would we include `png$` in the regular expression, since that's part of the `src` attribute and not the `alt` attribute?

**To Reproduce**
1. Go to https://testing-library.com/docs/dom-testing-library/api-queries#byalttext
2. Scroll down to the `ByAltText` example, if it isn't already visible
3. Look at the example code

**Expected behavior**
I would expect the example to use "Incredibles 2 Poster", `/incredibles/i`, or something without the "png" as the argument to `getByAltText`.

If `/incredibles.*png$/i` is somehow correct, then perhaps a note explaining why that works would be helpful.

**Screenshots**
image

N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
I haven't used this library yet, so maybe `/incredibles.*png$/i` is perfectly valid here. If it is, it is confusing, though.

bjohn465:
Fixes #116

I replaced `/incredibles.*png$/i` with `'Incredibles 2 Poster'`, which should match the `alt` attribute in the example markup.

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -308,7 +308,7 @@ as it's deprecated).
 import { getByAltText } from 'dom-testing-library'
 
 const container = document.body
-const incrediblesPosterImg = getByAltText(container, /incredibles.*png$/i)
+const incrediblesPosterImg = getByAltText(container, 'Incredibles 2 Poster')
 ```
 
 <!--React-->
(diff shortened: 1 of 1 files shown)


bjohn465:
Will do

kentcdodds:
@all-contributors please add @bjohn465 for docs

PR #120, opened 2019-05-28
Update vue-testing-library page

afontcu:
Hi! 

I noticed a few outdated links and also quick wins in vue-testing-library page, so here's a PR.

btw, this page and `vue-testing-library` README are so similar. Does it make sense? Should the page on testing-library.com contain different information from the README on GitHub?

Thank you all for your work in this project.


kentcdodds:
My thoughts: the README of the vue one should just direct to the website.

PR #125, opened 2019-05-31
Update Testing Libraries naming in docs

afontcu:
Hi! 👋  As suggested in #124:

- [x] Update `dom-testing-library`.
- [x] Update `vue-testing-library`.
- [x] Update `react-testing-library`.
- [x] Update `cypress-testing-library`.
- [x] Update `native-testing-library`.
- [x] Update `jest-native`.

Also:
- [x] Angular is being updated in #126.
- [x] Svelte is being updated in #127.
- [x] Testcafe is being updated in #128.

I didn't update the `2019-04-25-new-org` blog post where libraries are mentioned, because... well, it is an "old" post. I guess a newer blog post will be published after the migration to scoped packages has finished, so.. 🤷‍♀ 

Feel free to suggest any modification!

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ testing-library-docs
 [![Code of Conduct][coc-badge]][coc]
 <!-- prettier-ignore-end -->
 
-Documentation site for [react-testing-library](https://github.com/testing-library/react-testing-library), [DOM Testing Library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
+Documentation site for [React Testing Library](https://github.com/testing-library/react-testing-library), [DOM Testing Library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
 
 See [./website/README.md](./website/README.md) for instructions on building the site
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ testing-library-docs
 [![Code of Conduct][coc-badge]][coc]
 <!-- prettier-ignore-end -->
 
-Documentation site for [react-testing-library](https://github.com/testing-library/react-testing-library), [dom-testing-library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
+Documentation site for [react-testing-library](https://github.com/testing-library/react-testing-library), [DOM Testing Library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
 
 See [./website/README.md](./website/README.md) for instructions on building the site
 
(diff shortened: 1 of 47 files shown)


docs/cypress-testing-library/intro.md
```diff
@@ -45,9 +45,9 @@ cy.get('form').then(subject => {
 ```
 
 `cypress-testing-library` supports both jQuery elements and DOM nodes. This is
-necessary because Cypress uses jQuery elements, while `dom-testing-library`
+necessary because Cypress uses jQuery elements, while `DOM Testing Library`
 expects DOM nodes. When you pass a jQuery element as `container`, it will get
 the first DOM node from the collection and use that as the `container` parameter
-for the `dom-testing-library` functions.
+for the `DOM Testing Library` functions.
```
kentcdodds: Let's remove the backtics around these. It only _sorta_ made sense before, but I don't think it makes any sense now.

README.md
```diff
@@ -8,7 +8,7 @@ testing-library-docs
 [![Code of Conduct][coc-badge]][coc]
 <!-- prettier-ignore-end -->
 
-Documentation site for [react-testing-library](https://github.com/testing-library/react-testing-library), [dom-testing-library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
+Documentation site for [react-testing-library](https://github.com/testing-library/react-testing-library), [DOM Testing Library](https://github.com/testing-library/dom-testing-library), and [related projects](https://github.com/testing-library)
```
kentcdodds: Could we fix the `react-testing-library` reference here as well?

kentcdodds:
Thanks a ton! Maybe we could add a note to the old blogpost?

README.md
afontcu: Sure! I only focused on DOM Testing Lib, but as suggested in the initial issue, I'll also update RTL, Cypress, and Vue (at least).

docs/cypress-testing-library/intro.md
afontcu: Done! 👍 

Yet, backticks are used in several other places ([here](https://github.com/testing-library/testing-library-docs/pull/125/files#diff-4b86fa7efa07bfd9ab506f697f6999daR8), for instance, or a couple of lines above your comment).

Shoud we remove all backticks when talking about the library, now that it doesn't look like "code" anymore? Or there's another reason to modify this particular line I'm failing to see?

afontcu:
So, I think I updated all references to ported libraries that are not being handled in other PRs!

The blog post is still missing: would you like to add a footnote saying that names have changed, or update the blog post so it uses the new ones?

kentcdodds:
I've merged this and if there are necessary changes we can make them in another PR.

On the blog post, do what you think makes the most sense 👍

kentcdodds:
@all-contributors please add @afontcu for docs

PR #126, opened 2019-05-31
docs: point to @testing-library/angular

timdeschryver:
😅 Seems like I was a bit to enthusiastic, but this is the Angular's part for #124


kentcdodds:
@all-contributors please add @timdeschryver for docs

PR #127, opened 2019-06-01
updated testcafe docs for new name

benmonro:


kentcdodds:
@all-contributors please add @benmonro  for docs

PR #132, opened 2019-06-01
Add improved Vue docs

afontcu:
As discussed [here](https://github.com/testing-library/vue-testing-library/issues/33), I'm submitting a PR to extend Vue Testing Library docs.

There's a lot of content, so feel free to suggest any modification, clarification or addition! I tried to use as many bits from DOM/React Testing Library as possible.

Thanks!


afontcu:
Can't suggest reviewers, so I'm manually pinging @dfcook ^_^

PR #138, opened 2019-06-05
Document using with Jest mock functions

Shelob9:
This was something I needed to do today, and it wasn't documented, so I wanted to share.


kentcdodds:
Thanks for this. I think it's an important thing to include in the docs. However I want to change a few things. Here's how I would write it:

```jsx
// button.js
const Button = ({onClick, children}) => <button onClick={onClick}>{children}</button>

// __tests__/button.js
test('calls onClick prop when clicked', () => {
  const handleClick = jest.fn()
  const {getByText} = render(<Button onClick={handleClick}>Click Me</Button>)
  fireEvent.click(getByText(/click me/i))
  expect(handleClick).toHaveBeenCalledTimes(1)
})
```

I'll add comments to the lines in your original sample to explain why I made the changes that I did.

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
 fireEvent(node, myEvent)
 // myEvent.timeStamp can be accessed just like any other properties from myEvent
 ```
+
+## Using Jest Function Mocks
+[Jest's Mock functions](https://jestjs.io/docs/en/mock-functions) can be used to test that a callback passed to the function was called, or what it was called when the event that **should** trigger the callback function does trigger the bound callback.
+
+```jsx
+const Button = ({onClick,children}) => <button onClick={onClick}>{children}</button>
+describe('Button component', () => {
```
kentcdodds: I prefer to [avoid nesting](https://www.briefs.fm/3-minutes-with-kent/27). We already know that this is about the button component because we're in the file that's testing the button component.

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
...
+    let onClick;
```
kentcdodds: Sharing this variable through a bunch of tests makes it harder to maintain long-term. This isn't quite the same thing, but it's similar to what I talk about in [Test Isolation with React](https://kentcdodds.com/blog/test-isolation-with-react).

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
...
+    beforeEach(() => {
+        onClick = jest.fn();
+    });
+    afterEach(cleanup);
```
kentcdodds: We can just assume that people have the cleanup stuff working in their setup.

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
...
+
+    it('Fires on click event', () => {
+        const { container } = render(
+            <Button
+                label={'With Callback'}
```
kentcdodds: The Button component does nothing with the `label` prop so I removed that.

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
...
+                onClick={onClick}
+            >
+                Trigger Alert
+            </Button>
+        );
+        const input = container.querySelector('button');
```
kentcdodds: We're not looking for an input, and we should avoid using and suggesting `querySelector` and instead go with a more semantic selector like `getByTest`.

docs/dom-testing-library/api-events.md
```diff
@@ -93,3 +93,31 @@ const myEvent = createEvent.click(node, { button: 2 })
...
+        fireEvent.click(input);
+        expect(onClick.mock.calls[0].length).toBe(1);
```
kentcdodds: There's a built-in assertion for this, so I switched to that one.

Shelob9:
I think that the docs could use a similar example for onChange. Should this be a new PR?

These are the examples abstracted out from what I'm actually working on.

An input component with some logic inside of it:
```
// input.js
const Input = ({ onChange, value }) => {
  const changeHandler = event => onChange(event.target.value);
  return (
    <input data-testid="input-test" onChange={changeHandler} value={value} />
  );
};
```

Tests that change handler is called and that is called with the right value:

// __tests__/input.js
describe("text input", () => {
  afterEach(cleanup);
  test("calls onChange prop when changed", () => {
    const onChange = jest.fn();
    const { getByTestId } = render(<Input onChange={onChange} value={""} />);
    fireEvent.change(getByTestId("input-test"), {
      target: { value: "New Value" }
    });
    expect(onChange).toHaveBeenCalledTimes(1);
  });

  test("When input is changed, the new value, not the event is returned by", () => {
    const onChange = jest.fn();
    const { getByTestId } = render(<Input onChange={onChange} value={""} />);
    fireEvent.change(getByTestId("input-test"), {
      target: { value: "New Value" }
    });
    expect(onChange).toHaveBeenCalledWith("New Value");
  });
});
```

>https://github.com/testing-library/testing-library-docs/pull/138#discussion_r290899979
These are nested with `cleanup()`.  I got tripped up very badly when moving tests from Enzyme to react-testing-library, beacuse I totally missed why it was important in groups of tests.

kentcdodds:
I think that's a great idea. Let's do a different PR for the `onChange`. One thing I'll mention is we really want to avoid using `TestId` queries in any of the examples if we can help it! thanks!

Shelob9:
Sounds great. I'll copy that into a real PR, and update it to query by label or text once this is merged.

afontcu:
Hi @Shelob9! Any chance you could work on Kent's feedback and update the PR? :)

afontcu:
Hi @Shelob9! Any chance you could work on Kent's feedback and update the PR? :)

Shelob9:
@afontcu Thanks for the nudge, this got lost in all things. I'm updating now.

afontcu:
Now that I look at the issue with fresh eyes, I'm wondering if we should use a React-based example to illustrate an example in DOM Testing Library docs?

We could either (1) move the example to React docs, and add demos for other frameworks, or (2) use plain DOM example with `document.createElement` and so on.

Thoughts? 🙌

kentcdodds:
> we should use a React-based example to illustrate an example in DOM Testing Library docs?

Yeah. I don't think we should use the react stuff with DOM Testing Library. Let's make it as vanilla as possible.

> move the example to React docs, and add demos for other frameworks, or (2) use plain DOM example with document.createElement and so on.

I like both ideas. We should have demos/examples for all frameworks, including vanilla (document.createElement) stuff for DOM Testing Library.

alexkrolick:
If someone wants to add a generic version, add a codetab for it. I wrapped the existing example in a React tab.

Screen Shot 2020-10-01 at 12 13 20 PM

PR #146, opened 2019-06-12
Fix native-testing navigation

michalak111:


bcarroll22:
This works for me. Eventually I'd like to get a docs page cheat sheet similar to the ones for DOM and React Testing Library, but we can address that later. Thanks! 

@allcontributors please add @michalak111 for doc.

PR #148, opened 2019-06-14
📝 Add correct tsconfig instructions for cypress

omerose:
Edited README only.

Why?
I followed the instructions and it didn't work with recent changes to official instructions from Cypress-typescript I installed last week. So these documentation will help out anyone doing this today.

I also found the types file to be handy for all library definitions, so I added it into the README


omerose:
@alexkrolick  Can you take a look at my equivalent PR on the actual REPO instead of the documentation?

[Here](https://github.com/testing-library/cypress-testing-library/pull/54)

PR #151, opened 2019-06-20
docs: fix import in render function queries option example

kierenhughes:
I was looking at adding a custom data attribute query and found these docs, but RTL doesn't contain a default export so importing queries as in the example doesn't work. Just added a quick fix for the import.


kentcdodds:
@all-contributors, please add @kierenhughes for docs

PR #152, opened 2019-06-20
docs: add custom queries example to custom render docs

kierenhughes:
Following on from my previous PR (#151) and having noticed the discussion in #80:

* Adding an example for adding custom queries globally via creating a custom render function.

In the example, I've create a data attribute selector for `data-cy` as this is one of the attributes recommended by Cypress. I think in reality I would favour replacing `data-cy` globally throughout my projects with `data-testid`, but this may be useful for some people not willing to refactor out `data-cy`, or similar, at the moment.

### DIFF
--- a/docs/react-testing-library/setup.md
+++ b/docs/react-testing-library/setup.md
@@ -147,31 +147,76 @@ module.exports = {
 <details>
 <summary>Customize queries globally with custom render</summary>
 
-You can override and append queries via the render function by passing a [`queries`](api.md#render-options) option.
+In the example below, a new set of query variants are created for getting
+elements by `data-cy`, an attribute used and recommended by [Cypress.io](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements).
 
-You can define your own custom queries as described in the example in the [Helpers API](api-helpers.md) documentation.
+It is worth noting that generally you should not need to use this feature of
+react-testing-library and, where you do use it, you should really consider whether
+your new queries encourage you to test in a user-centric way, without testing
+implementation detail.
+
+`data-testid` is supported by Cypress.io, so even in this case, you would not technically
+need to create a new set of queries and you could just replace `data-cy` with
+`data-testid` attributes throughout your project. 
+
+You can define your own custom queries as described in the example in the 
+[Helpers API](/doc/dom-testing-library/api-helpers.md) documentation, or via the
+ [`buildQueries`](/doc/dom-testing-library/api-helpers#buildQueries) helper:
+
+```js
+// custom-queries.js
+import { queryHelpers, buildQueries } from '@testing-library/react';
+
+const queryAllByDataCy = (...args) => 
+  queryHelpers.queryAllByAttribute('data-cy', ...args);
+
+const getMultipleError = (c, dataCyValue) => (
+  `Found multiple elements with the data-cy attribute of: ${dataCyValue}`
+);
+const getMissingError = (c, dataCyValue) => (
+  `Unable to find an element with the data-cy attribute of: ${dataCyValue}`
+);
+
+const [
+  queryByDataCy,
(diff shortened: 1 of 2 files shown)


docs/react-testing-library/setup.md
```diff
@@ -144,6 +144,35 @@ module.exports = {
...
+You can override and append queries via the render function by passing a [`queries`](api.md#render-options) option.
+
+If you want to add custom queries globally, you can do this by defining a custom render method:
+
+```js
+// test-utils.js
+import { render, queries, queryHelpers } from '@testing-library/react'
+
+const customQueries = {
+  getByDataCy: queryHelpers.queryByAttribute.bind(
```
alexkrolick: It would be nice to show how to use the `buildQueries` function to create a full set of custom queries that conforms to the `get`/`query`/`find` convention

https://github.com/testing-library/dom-testing-library/blob/master/src/query-helpers.js#L103

docs/react-testing-library/setup.md
```diff
@@ -144,6 +144,35 @@ module.exports = {
 
 </details>
 
+<details>
+<summary>Modify queries globally with custom render</summary>
```
alexkrolick:
```suggestion
<summary>Customize queries globally with custom render</summary>
```

docs/react-testing-library/setup.md
```diff
@@ -144,6 +144,35 @@ module.exports = {
...
+const customRender = (ui, options) =>
+  render(ui, { queries: { ...queries, ...customQueries } })
+
+// re-export everything
+export * from '@testing-library/react'
+
+// override render method
+export { customRender as render }
+```
+</details>
```
afontcu: Looks great! It'd be amazing if, after the config example, we provided a demo usage (as simple as querying a single `<div>` or whatever).
Some people might have a hard time understanding the `queryHelpers`/`buildQueries` API, so a final example showing how to use the newly availalbe `getByDataCy` would come in handy. What do you think about that?

docs/react-testing-library/setup.md
kierenhughes: Great suggestion - I'll look into it.

docs/react-testing-library/setup.md
kierenhughes: Will update! Even if my British instincts make me want to correct it to 'Customi**s**e' 😂

docs/react-testing-library/setup.md
kierenhughes: @alexkrolick do you reckon using `buildQueries` in this example specifically would be useful, or shall I get an example added to the 'Helpers' docs for `dom-testing-library`?

docs/react-testing-library/setup.md
kierenhughes: Fair point! Added a `getByDataCy` usage example 👍

docs/react-testing-library/setup.md
alexkrolick: The function should be documented in the API section and it can point to a usage example here

kierenhughes: @alexkrolick added example usage, along with a brief description in the API section

kierenhughes: @alexkrolick any further comments following my updates?

alexkrolick:
Should probably explain a little what "cy" stands for but other than that LGTM

kierenhughes:
@alexkrolick 

> Should probably explain a little what "cy" stands for but other than that LGTM

I've added a bit about that along with the link to the relevant Cypress.io docs regarding selecting by data-*.

I've also added a recommendation around not needing to use your own queries and only doing so if your queries encourage testing in a user-centric way without focusing on implementation details.

alexkrolick:
@all-contributors please add @kierenhughes for docs

PR #155, opened 2019-06-20
Added docs for getRoles and logRoles

michaellasky:
These functions are proposed in dom-testing-library [PR#285](
https://github.com/testing-library/dom-testing-library/pull/285)


docs/dom-testing-library/api-helpers.md
```diff
@@ -148,6 +148,30 @@ cy.get('form').within(() => {
...
 <!--END_DOCUSAURUS_CODE_TABS-->
 
+## `getRoles`
+
+This function allows iteration over the implicit aria roles represented in a
+given tree of DOM nodes. It returns an object indexed by role name with each
+value being an array of elements which have that implicit aria role.
+
+```javascript
+import { getRoles } from 'dom-testing-library'
```
afontcu:
```suggestion
import { getRoles } from '@testing-library/dom' 
```

docs/dom-testing-library/api-helpers.md
```diff
@@ -213,3 +237,48 @@ console.log(prettyDOM(div))
...
+
+### `logRoles`
+
+This helper function can be used to print out a list of all the implicit aria
+roles within a tree of DOM nodes, each role containing a list of all of the
+nodes which match that role. This can be helpful for finding ways to query the
+DOM under test with [getByRole](api-queries.md#byrole)
+
+```javascript
+import { logRoles } from 'dom-testing-library'
```
afontcu:
```
import { logRoles } from '@testing-library/dom' 
```

docs/dom-testing-library/api-helpers.md
```diff
@@ -148,6 +148,30 @@ cy.get('form').within(() => {
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
+## `getRoles`
+
+This function allows iteration over the implicit aria roles represented in a
```
afontcu: It'd be great if we could provide a link to an exhaustive list of implicit ARIA roles. Couldn't find anything better than [this](https://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html), though.

Thoughts?

afontcu:
```suggestion
This function allows iteration over the implicit ARIA roles represented in a
```

(and same suggestion in other places where ARIA is lowercased!)

docs/dom-testing-library/api-helpers.md
```diff
@@ -148,6 +148,30 @@ cy.get('form').within(() => {
...
+given tree of DOM nodes. It returns an object indexed by role name with each
+value being an array of elements which have that implicit aria role.
+
+```javascript
+import { getRoles } from 'dom-testing-library'
+
+const nav = document.createElement('nav')
+nav.innerHTML = `
+<ul name="menu">
```
afontcu: Since `name` isn't really a valid attribute for ul/li elements, I'd suggest removing it, so there's also less code to read in the example.

What do you think about it?

michaellasky: Oops, I didn't realize name isn't valid for ul/li.  

I think we should have some attribute in there to illustrate how it prints out the attributes.  Thoughts?

Assuming the above, I'll probably change name="menu" to type="circle" on the ul and give the li elements a 'value' attribute in place of 'name'.

docs/dom-testing-library/api-helpers.md
michaellasky: Done [3390faa](https://github.com/testing-library/testing-library-docs/pull/155/commits/3390faabbafc4bc3c232c40b41f7c8a931099665)

michaellasky: I couldn't find much either.  That looks pretty decent though, just wish it was a bit more terse.  

Added a link to it here: [b026d9f](https://github.com/testing-library/testing-library-docs/pull/155/commits/b026d9fd87aaff435d6918a5ff1462cff28afd6e)

docs/dom-testing-library/api-helpers.md
michaellasky: Changed here: [da985145](https://github.com/testing-library/testing-library-docs/pull/155/commits/da9851456fff50320a83c7c268e43f722264bda4)

afontcu: Didn't know that until I checked right now, but `type` attr on `<ul>` elements is discouraged: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#attr-type

Also, looks like `value` on `<ul>` list items doesn't make much sense: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#attr-value

michaellasky: Hmm.. swing and a miss, haha  

nav->ul->li was nice because it is three distinct roles in a real world kind of situation.  

Maybe it doesn't matter so much if the example illustrates how the output prints attributes, so there's always the option of getting rid of the attributes in the example (as your initial suggestion).  

Though I guess we could always add class and/or id attributes to illustrate a more real-world output.

I'm gonna push one more commit this evening.

afontcu: ;) no prob.

> Maybe it doesn't matter so much if the example illustrates how the output prints attributes, so there's always the option of getting rid of the attributes in the example.

This still makes sense to me. The main goal of the example is to illustrate what a grouping-by-role looks like, not the attribute themselves.

Thanks for your work!

michaellasky: Ok, I've pulled the attributes off the elements completely.  I agree that it's the correct call.  Sorry for the delay, distracted by work/holiday.

kentcdodds:
@all-contributors please add @michaellasky for docs

PR #159, opened 2019-06-21
docs: add marko testing library

DylanPiercey:
This PR adds documentation for the new [`@marko/testing-library`](https://github.com/marko-js/testing-library) and closes https://github.com/testing-library/dom-testing-library/issues/290.

The above package is unreleased, but is ready to go hopefully along side this additional documentation!
Any feedback is much appreciated 😄

### DIFF
--- a/docs/marko-testing-library/api.md
+++ b/docs/marko-testing-library/api.md
@@ -1,0 +1,177 @@
+---
+id: api
+title: API
+---
+
+`Marko Testing Library` re-exports everything from `DOM Testing Library` as well
+as these methods:
+
+- [`render`](#render)
+- [`cleanup`](#cleanup)
+
+---
+
+## `render`
+
+```javascript
+function render(
+  template, // A Marko template to render
+  input, // Input for the above template
+  options // You won't often use this, expand below for docs on options
+):
+```
+
+Render into a container which is appended to `document.body`.
+
+```javascript
+import { render } from '@marko/testing-library'
+import MyTemplate from './my-template.marko'
+
+render(MyTemplate)
+```
+
+```javascript
+import { render, cleanup } from '@marko/testing-library'
+import 'jest-dom/extend-expect'
+import Greeting from './greeting.marko'
+
+afterEach(cleanup)
+
+test('renders a message', async () => {
+  const { container, getByText } = await render(Greeting, { name: 'Marko' })
+  expect(getByText(/Marko/)).toBeInTheDocument()
+  expect(container.firstChild).toMatchInlineSnapshot(`
+    <h1>Hello, Marko!</h1>
+  `)
+})
+```
+
+> Note
+>
+> The [cleanup](#cleanup) function should be called between tests to remove the
+> created DOM nodes and keep the tests isolated.
+
+## `render` Options
+
+You won't often need to specify options, but if you ever do, here are the
+available options which you could provide as the third argument to `render`.
+
+### `container`
+
+By default for client side tests, `Marko Testing Library` will create a `div`
+and append that `div` to the `document.body` and this is where your component
+will be rendered. If you provide your own HTMLElement `container` via this
+option, it will not be appended to the `document.body` automatically.
+
+For example: If you are unit testing a `tablebody` element, it cannot be a child
+of a `div`. In this case, you can specify a `table` as the render `container`.
+
+```javascript
+const table = document.createElement('table')
+
+const { container } = await render(MyTableBody, null, {
+  container: document.body.appendChild(table),
+})
+```
+
+## `render` Result
+
+The `render` method returns a promise which resolves with an object that has a
+few properties:
+
+### `...queries`
+
+The most important feature of `render` is that the queries from
+[DOM Testing Library](dom-testing-library/api-queries.md) are automatically
+returned with their first argument bound to the results of rendering your
+component.
+
+See [Queries](dom-testing-library/api-queries.md) for a complete list.
+
+**Example**
+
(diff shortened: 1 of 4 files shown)


docs/marko-testing-library/api.md
```diff
@@ -0,0 +1,177 @@
...
+---
+
+## `render`
+
+```javascript
+function render(
+  template, // A Marko template to render
+  input, // Input for the above template
+  options // You won't often use this, expand below for docs on options
+):
```
afontcu:
```suggestion
)
```

docs/marko-testing-library/api.md
```diff
@@ -0,0 +1,177 @@
...
+  `)
+})
+```
+
+> Note
+>
+> The [cleanup](#cleanup) function should be called between tests to remove the
+> created DOM nodes and keep the tests isolated.
+
+## `render` Options
```
afontcu: Docs looks great and solid! 👍 

It's a bit weird, thou, that the "render options" section uses the same heading as the parent "render" heading.
I'd suggest setting "render Options" as a third-level heading, and "container" as a fourth-level one.

😃 


```suggestion
### `render` Options
```

DylanPiercey: Excellent catches!
Updated the PR based on your comments.

Thanks for the review! 😄

DylanPiercey: I realized I missed this part:

> Same thing with "render result" and its children.

I think having those as a higher heading is still useful as it causes it to show up in the side nav (and these are probably the most important parts of the lib). This is also what the `react-testing-library` docs have setup (https://github.com/testing-library/testing-library-docs/blob/79a0c1b8127d30b04182fbc528ebcb34adee3a0e/docs/react-testing-library/api.md#L116).

afontcu: Totally right, I was thinking about that while typing, actually 🤔 Let's follow RTL example, then! If we ever feel the urge to change it we can make sure we update each framework docs :)

DylanPiercey: I still like moving `render Options` as a sub heading, since in this package there is only one option and honestly it would be very rare that people use it. Could promote it to a higher heading (as is done in the react docs) if more options become available, but I don't see that happening.

In my last commit I updated the PR to do as I said above, but I left the `render Result` heading alone.

I think I'm happy with it now 😄

docs/marko-testing-library/api.md
```diff
@@ -0,0 +1,177 @@
...
+// <h1>Hello World</h1>
+// you can also pass an element: debug(getByTestId('messages'))
+```
+
+This is a simple wrapper around `prettyDOM` which is also exposed and comes from
+[`DOM Testing Library`](https://github.com/testing-library/dom-testing-library/blob/master/README.md#prettydom).
+
+### `rerender`
+
+It'd probably be better if you test the component that's doing the prop updating
```
alexkrolick: I think this could be worded a bit more conservatively - rerender behavior is a valid thing to test for reusable components

docs/marko-testing-library/setup.md
```diff
@@ -0,0 +1,105 @@
+---
+id: setup
+title: Setup
+sidebar_label: Setup
+---
+
+`Marko Testing Library` is not dependent on any test runner, however it is
+dependent on the test environment. These utilities work for testing both server
```
alexkrolick:
```suggestion
dependent on browser globals. These utilities work for testing both server

alexkrolick: What are "these utilities" in this paragraph?

docs/marko-testing-library/api.md
DylanPiercey: @alexkrolick I agree. This is a remnant of basing the docs off of the react ones.

Let me try to word smith this better.

docs/marko-testing-library/setup.md
DylanPiercey: Maybe `test environment` is not the most clear here. Basically I am trying to express that if your tooling does not pick up on the `browser` field in the package.json then your tests will be running in the SSR mode. For jest this just means enabling the `browser` option. Just having a global JSDOM for example (as is the default for jest) will not work by itself. It's a bit of a strange thing to explain to new comers though, do you have some thoughts on how to better describe this?

`these utilities` really just mean the api's exposed by the package. Maybe `these apis` would be better?

alexkrolick: Does that mean you can only test built files and not individual components? AFAIK most tests that require from the src folder aren't really aware of the packaging system.

DylanPiercey: You can test individual components, the key is that by default the server side version of the component is resolved. For a client side version of a component to be resolved, the bundler, or test runner, needs to be aware of the [browser field](https://github.com/defunctzombie/package-browser-field-spec).

In Marko components are compiled to an optimized SSR streaming API or a VDOM api. The browser field is used to provide an appropriate runtime based on the environment.

DylanPiercey:
Let's hold on merging this for now. Going to be addressing the comments above, and I also want to have a good solution for working with events emitted from Marko components.

Will update this PR hopefully today 😄

docs/marko-testing-library/api.md
DylanPiercey: Updated this section of the docs, let me know what you think 😄

DylanPiercey:
Alright, I've improved the `rerender` documentation and added a section about testing emitted events from components. I think it's in a pretty good state 😄

PR #162, opened 2019-06-23
change text cursor to pointer cursor

iqbal125:
hovering over each question in the faq, results in the cursor changing to a text cursor. I have added a simply style attribute of pointer: cursor and id to each summary.

### DIFF
--- a/docs/react-testing-library/faq.md
+++ b/docs/react-testing-library/faq.md
@@ -8,13 +8,9 @@ to React testing
 
 <details>
   
-<style>
- #summary {
-    pointer: cursor;
-  }
-</style>
 
-<summary id="summary">How do I test input onChange handlers?</summary>
+
+<summary>How do I test input onChange handlers?</summary>
 
 TL;DR:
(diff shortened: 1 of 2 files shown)


alexkrolick:
There are several more usages of summary/details tags on different pages in the docs - can you explain why the default is confusing and/or suggest a fix that works everywhere?

iqbal125:
@alexkrolick So hovering over the summary links turns the cursor to a text pointer, which makes it seem like its not clickable, but it is clickable. 

so adding a simply style to make the cursor a pointer will make it easy for everyone to see that it is clickable. 

Only for this page I cant speak for the rest of the docs

afontcu:
I kind of agree with the idea of the PR - `<summary>` provides a text cursor style, and it might be confusing.

However, it is true that we should find a way to style all `<summary>` elements of the docs, not only the ones in React section. Maybe a simple element styling in `/website/static/css/custom.css`?

```css
details summary {
  cursor: pointer;
}
```


What do you think about that?

alexkrolick:
@afontcu maybe scope that to the article class if there is one

afontcu:
Hi @iqbal125! Did you get a chance to look into this?

iqbal125:
hello again @afontcu, I made changes based on feedback. I also updated my other PRs, do you mind taking a look?

alexkrolick:
Thank you for the update!

PR #165, opened 2019-06-26
Remove non-existent parameter from getByText

afontcu:
Hi! 👋 

AFAIK, RTL automatically binds the queries' first argument to baseElement (`document.body` by default), so there's no need to specify a container.

thx!


alexkrolick:
good catch

PR #166, opened 2019-06-27
Fix a capitalization error

huyenltnguyen:
Changed the TextMatch from `username` to `Username` in the `getByLabelText()` example.

I believe the method uses exact match by default, which would make the test fail as the label it is looking for has a value of `Username`.


afontcu:
Nice catch!

PR #167, opened 2019-06-28
Break React-Redux example into multiple "files" (code blocks)

SeanMcP:
The previous example put all the necessary variables into one "file" (code block). However, this is unrealistic, and I found it confusing when first reading through the docs.

Instead, I've broken each chunk of this example into its own "file", which I think reads better and will be more applicable to a user's code base.

### DIFF
--- a/docs/example-react-redux.md
+++ b/docs/example-react-redux.md
@@ -4,12 +4,10 @@ title: React Redux
 ---
 
 ```jsx
-import React from 'react'
-import { createStore } from 'redux'
-import { Provider, connect } from 'react-redux'
-import { render, fireEvent, cleanup } from '@testing-library/react'
-
 // counter.js
+import React from 'react'
+import { connect } from 'react-redux'
+
 class Counter extends React.Component {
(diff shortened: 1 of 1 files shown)


docs/example-react-redux.md
```diff
@@ -33,14 +31,18 @@ class Counter extends React.Component {
...
-// export default connect(state => ({count: state.count}))(Counter)
-// but for this test we'll give it a variable name
-// because we're doing this all in one file
-const ConnectedCounter = connect(state => ({ count: state.count }))(Counter)
+export default connect(state => ({count: state.count}))(Counter)
+```
+
+For this example, we'll have a simple reducer that tracks and updates `count`:
+
+```
```
kentcdodds: Let's specify a language for this codefence for syntax highlighting

````
```js
````

SeanMcP: Good catch.

kentcdodds:
@all-contributors please add @SeanMcP for docs

PR #169, opened 2019-06-29
create example-react-hooks-useState.md

iqbal125:
react testing with react hooks and usestate,

you guys have some fairly advanced examples, I thought I would add a simple React Hooks and useState example for people who may not be as experienced.  

This is a simple test that checks to see if text in state has changed based on a button click.


afontcu:
Hi! 👋 Content looks good! I'd suggest the same modifications we've been discussing in your other PRs: avoiding semicolons, adding backticks (` `` `) around code, etc.

by the way, I'd say that having a simpler example when starting reading the docs would benefit new users. So I suggest using this simple `setState` test as the initial example provided by the docs in [Example](https://testing-library.com/docs/react-testing-library/example-intro) (we could provide both). What do you think about it @alexkrolick?

alexkrolick:
@afontcu adding a simple example sounds like a good idea

alexkrolick:
@iqbal125 FYI if you check the "allow maintainers to make edits" box on the PR we can commit suggestions directly to the branch rather than bugging you about it.

docs/example-react-hooks-useState.md
```diff
@@ -0,0 +1,81 @@
...
+
+This will be a simple example using useState to test a React Hooks functional component.
+
+Here is our component, it is a simple component that changes some text when a user clicks a button: 
+
+```
+import React, { useState } from 'react';
+
+
+const TestHook = (props) => {
```
alexkrolick: One thing I'd like to avoid is too many examples using "TestXYX" as the name. That makes it sound like this is a special test harness or wrapper for testing. Names like `Example`, `MyComponent`, `App`, or a descriptive name like `ToggleButton` avoid this confusion.

iqbal125:
@alexkrolick I actually have it the checkboxed. weird why its not letting you edit 

image

iqbal125:
@alexkrolick @afontcu  Hey guys made the updates, let me know if I missed something

afontcu:
Hi @iqbal125! 

Regarding [this comment](#issuecomment-515703373):

> Having a simpler example when starting reading the docs would benefit new users. So I suggest using this simple `setState` test as the initial example provided by the docs in [Example](https://testing-library.com/docs/react-testing-library/example-intro) (we could provide both)


would you mind "promoting" this test to the "main" test in the Example page to provide an easier onboarding to new users?

afontcu:
I've updated the example to match React docs' first example on useState: https://reactjs.org/docs/hooks-state.html

I feel this is the best way to provide a pleasant onboarding.

iqbal125:
@afontcu done, I deleted the old example.

afontcu:
Hi @iqbal125! Thanks for this. I think it is better to "add" this simple example to the page, instead of removing the current one.

```
## Example

// the simple button example


## A more complex example

// the actual <Fetch> example
```

also, the useState example page would go away.

thanks!

iqbal125:
@afontcu gotcha will do.

iqbal125:
@afontcu made the updates

afontcu:
Looks quite good to me, I'd love to know @alexkrolick and even @kentcdodds opinion on this (since we are messing with the entry point for new RTL users).

docs/react-testing-library/example-intro.md
```diff
@@ -4,25 +4,77 @@ title: Example
...
+// import the component to test
+import Counter from '../Counter.js'
+
+test('increments value on click', () => {
+  // The render method returns a collection of utilities to query your component
+  const { getByText } = render(<Counter />)
+
+  // getByText returns the first matching node for a query, and throws an error
+  // if no elements match or if more than one match is found
+  getByText('Times clicked: 0')
```
kentcdodds: I think some people don't like the way the implicit assertion looks, so it may be nice to add an extra comment:

```javascript
// if you prefer to make an explicit assertion, you can do so like this:
// expect(getByText('Times clicked: 0')).toBeInTheDocument()
```

docs/react-testing-library/example-intro.md
```diff
@@ -4,25 +4,77 @@ title: Example
...
+
+  const button = getByText('Click me')
+
+  // Fires a click DOM event to the specified target
+  fireEvent.click(button)
+  fireEvent.click(button)
+
+  getByText('Times clicked: 2')
```
kentcdodds: I prefer to use case-insensitive regex rather than strings because they're more resilient to changes. Could this (and others like it) be changed to: `getByText(/times clicked: 2/i)`

docs/react-testing-library/example-intro.md
```diff
@@ -96,7 +140,7 @@ const { getByText, getByTestId, container, asFragment } = render(
 )
 ```
 
-### Act
+#### Act
 
 The [`fireEvent`](dom-testing-library/api-events.md) method allows you to fire
 events to simulate user actions.
```
kentcdodds: The `axiosMock` stuff below here should probably be part of "Arrange"

docs/react-testing-library/example-intro.md
```diff
@@ -118,7 +162,39 @@ const greetingTextNode = await waitForElement(() =>
 )
 ```
 
-### Assert
+#### Assert
+
+```jsx
+// fetch.js
+
+import React,{useState} from 'react'
```
kentcdodds: Add a space after `,`

docs/react-testing-library/example-intro.md
```diff
@@ -118,7 +162,39 @@ const greetingTextNode = await waitForElement(() =>
 )
 ```
 
-### Assert
+#### Assert
+
+```jsx
+// fetch.js
```
kentcdodds: Should this be above "Arrange"? It seems odd to include it in the middle here with "Assert"

docs/react-testing-library/example-intro.md
```diff
@@ -127,7 +203,12 @@ expect(getByTestId('greeting-text')).toHaveTextContent('hello there')
 expect(getByTestId('ok-button')).toHaveAttribute('disabled')
 
 // snapshots work great with regular DOM nodes!
```
kentcdodds: Could we add a link to my blog post about snapshots: https://kcd.im/snapshots

PR #171, opened 2019-06-30
create example-react-hooks-useContext

iqbal125:
simple test example to test if a child component can update the context state.

### DIFF
--- a/docs/example-react-hooks-useContext.md
+++ b/docs/example-react-hooks-useContext.md
@@ -1,0 +1,94 @@
+This is a simple example to test a component using the useContext hook. 
+
+
+The context object
+
+```
+import React from 'react';
+
+const Context = React.createContext()
+
+export default Context
+```
+
+the root parent component
+
+```
+import React, { useState } from 'react';
+import Context from './components/store/context';
+
+
+const App = () => {
+  const [state, setState] = useState("Some Text")
+
+  const changeText = () => {
+    setState("Some Other Text")
+  }
+
+   <h1> Basic Hook useContext</h1>
+     <Context.Provider value={{changeTextProp: changeText,
+                               stateProp: state
+                                 }} >
+        <TestHookContext />
+     </Context.Provider>
+    </div>
+  );
+}
+
+export default App;
+
+```
+
+
+the child component using the context
+
+```
+import React, { useContext } from 'react';
+
+import Context from '../store/context';
+
+const TestHookContext = () => {
+  const context = useContext(Context)
+
+  return (
+    <div>
+    <button onClick={context.changeTextProp}>
+        Change Text
+    </button>
+      <p>{context.stateProp}</p>
+    </div>
+  )
+}
+
+
+export default TestHookContext;
+```
+
+and the tests 
+
+```
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestHookContext from '../test_hook_context.js';
+import {render, fireEvent, cleanup} from '@testing-library/react';
+import App from '../../../App'
+
+import Context from '../../store/context';
+
+afterEach(cleanup)
+
+it('Context value is updated by child component', () => {
+
+   const { container, getByText } = render(<App>
+                                            <Context.Provider>
+                                             <TestHookContext />
+                                            </Context.Provider>
+                                           </App>);
+
+   expect(getByText(/Some/i).textContent).toBe("Some Text")
+
+   fireEvent.click(getByText("Change Text"))
+
+   expect(getByText(/Some/i).textContent).toBe("Some Other Text")
+})
+```
--- a/docs/example-react-hooks-useContext.md
+++ b/docs/example-react-hooks-useContext.md
@@ -3,7 +3,7 @@ This is a simple example to test a component using the useContext hook.
 
 The context object
 
-```
+```js
 import React, { createContext } from 'react'
 
 const Context = createContext()
@@ -13,7 +13,7 @@ export default Context
 
 the root parent component
 
-```
+```jsx
 import React, { useState } from 'react'
 import Context from './context'
 
@@ -25,6 +25,7 @@ const App = () => {
     setState("Some Other Text")
   }
 
+return (
    <h1> Basic Hook useContext</h1>
      <Context.Provider value={{changeTextProp: changeText,
                                stateProp: state
@@ -42,7 +43,7 @@ export default App
 
 the child component using the context
 
-```
+```jsx
 import React, { useContext } from 'react'
 
 import Context from '../context'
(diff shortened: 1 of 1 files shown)


alexkrolick:
One comment - naming the inner component `TestComponent` could make people think that is a special for component just for the test instead of part of the App. Can we change that and add a note along the lines of:

> The basic idea is to test your components as they are used. Instead of testing the context-providing component separately, your main focus should be testing that the features connected to the context work correctly.

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,94 @@
+This is a simple example to test a component using the useContext hook. 
+
+
+The context object
+
+```
+import React from 'react';
```
afontcu: Hi! Some parts of the code are using semicolons, and some others aren't. I just checked and I'd say all demo code from the docs are written without semicolons, so I'd suggest removing them from the examples too!

alexkrolick: If you pull down the repo locally the git hooks should format the docs on save

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,94 @@
...
+import App from '../../../App'
+
+import Context from '../../store/context';
+
+afterEach(cleanup)
+
+it('Context value is updated by child component', () => {
+
+   const { container, getByText } = render(<App>
+                                            <Context.Provider>
```
eps1lon: Missing a value here. I'm not familiar with this usage. Does this set the value to `undefined` or the default context value? If it sets it to `undefined` we should make it explicit. If it uses the default value I would rather omit the provider component.

alexkrolick: The default value is only used when there is no Provider at all above a Consumer

iqbal125:
@kentcdodds  @eps1lon  @alexkrolick  @afontcu Made changes based on feedback let me know what you think.

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+const App = () => {
+  const [state, setState] = useState("Some Text")
+
+  const changeText = () => {
+    setState("Some Other Text")
+  }
+
+   <h1> Basic Hook useContext</h1>
```
afontcu: isn't a `return` statement missing here? I'd run all code in docs, to make sure it is working properly. We don't want to provide broken demos to our readers! 😉

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,95 @@
+This is a simple example to test a component using the useContext hook. 
+
+
+The context object
+
+```
```
afontcu: You should add the language type for the syntax highlighter: ` ```js ` or ` ```jsx ` or similar

docs/example-react-hooks-useContext.md
iqbal125: Wow your right cant believe I missed something so obvious

docs/example-react-hooks-useContext.md
iqbal125: I included it because I thought it would make the code easier to read, I'll change to render(<App/>)

iqbal125:
@afontcu Made changes based on feedback let me know what you think.

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,92 @@
...
+
+```jsx
+import React, { useContext } from 'react'
+
+import Context from '../context'
+
+const ChildComponent = () => {
+  const context = useContext(Context)
+
+  return (
```
afontcu: Looks better! Thank you 🙌  could you please fix the indentation of the JSX elements?

```html
<div>
  <button ...>
    Change text
  </button>
  <p>...</p>
</div>
```

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+const App = () => {
+  const [state, setState] = useState("Some Text")
+
+  const changeText = () => {
+    setState("Some Other Text")
+  }
+
+return (
```
afontcu:
```suggestion
  return (
```

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,92 @@
...
+
+the root parent component
+
+```jsx
+import React, { useState } from 'react'
+import Context from './context'
+
+
+const App = () => {
+  const [state, setState] = useState("Some Text")
```
afontcu:
```suggestion
  const [state, setState] = useState('Some Text')
```

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,92 @@
...
+import {render, fireEvent, cleanup} from '@testing-library/react'
+import App from './App'
+
+import Context from './context'
+
+afterEach(cleanup)
+
+it('Context value is updated by child component', () => {
+
+   const { container, getByText } = render(<App />);
```
afontcu:
```suggestion
   const { getByText } = render(<App />)
```

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,92 @@
+This is a simple example to test a component using the useContext hook. 
+
+
+The context object
```
afontcu: I just noticed that other examples add the filename on top of the code snippet block, as in

```js
// context.js

import React from 'react'

....
```

I feel that could be seen as the "title" of the snippet. It'd be great to keep that strategy, to make things consistent. What do you think about it?

iqbal125: yeah makes sense

iqbal125:
@afontcu was able to make the changes let me know what you think.

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,96 @@
...
+import {render, fireEvent, cleanup} from '@testing-library/react'
+import App from './App'
+
+import Context from './context'
+
+afterEach(cleanup)
+
+it('Context value is updated by child component', () => {
+
+   const { getByText } = render(<App />);
```
afontcu:
```suggestion
   const { getByText } = render(<App />)
```

docs/example-react-hooks-useContext.md
```diff
@@ -0,0 +1,96 @@
...
+
+  return (
+   <h1> Basic Hook useContext</h1>
+     <Context.Provider value={{changeTextProp: changeText,
+                               stateProp: state
+                                 }} >
+        <ChildComponent />
+     </Context.Provider>
+    </div>
+  );
```
afontcu:
```suggestion
  )
```

afontcu:
I've cleaned up the file a bit (whitespacing, naming, etc), but I'd love someone else to check that the _actual_ example test makes sense (from a React/useContext point of view). Not sure if using it with `useState` makes the example a bit cumbersome.

Your call! :)

alexkrolick (commented):
It would be good to link this page to the `wrapper` docs to show how to deal with Context providers when you aren't rendering a whole app.

PR #172, opened 2019-07-01
Add guide for events

eps1lon:
Closes testing-library/dom-testing-library#293, testing-library/dom-testing-library#292

[View rendered text](https://deploy-preview-172--testing-library.netlify.com/docs/guide-events)

### DIFF
--- a/docs/guide-events.md
+++ b/docs/guide-events.md
@@ -22,13 +22,16 @@ order):
 - fireEvent.mouseDown(element)
 - element.focus() (if that element is focusable)
 - fireEvent.mouseUp(element)
-- fireEvent.click(element) And then, if that element happens to be a child of a
-  `label`, then it will also move focus to the form control that the label is
-  labeling. So even though all you really are trying to test is the click
-  handler, by simply using `fireEvent.click` you're missing out on several other
-  potentially important events the user is firing along the way. Again, most of
-  the time this isn't critical for your tests and the trade-off of simply using
-  `fireEvent.click` is worth it.
+- fireEvent.click(element)
+
+And then, if that element happens to be a child of a `label`, then it will also
+move focus to the form control that the label is labeling. So even though all
+you really are trying to test is the click handler, by simply using
+`fireEvent.click` you're missing out on several other potentially important
+events the user is firing along the way.
+
+Again, most of the time this isn't critical for your tests and the trade-off of
+simply using `fireEvent.click` is worth it.
 
 ## Alternatives
 
--- a/docs/guide-events.md
+++ b/docs/guide-events.md
@@ -16,7 +16,7 @@ element that should receive a keydown. It will only receive that event if the bu
 is focused. If it is focused an additional click event is fired if the Enter key
 was pressed.
 
-You don't need to know every possible combinations of interactions but you shouldn't
+You don't need to know every possible combination of interactions but you shouldn't
 have to much confidence in the interactive behavior of your components if you only
 test them with mocked sequences of events.
 
@@ -44,7 +44,7 @@ This will also test that the element in question can even receive keyboard event
 
 If an element is focused not only a focus event is dispatched. The active element
 in the document also changes as well as the previously focused element getting blurred.
-To simulate this behavior you simply replace `fireEvent` with imperative focus:
+To simulate this behavior you can simply replace `fireEvent` with imperative focus:
 
 ```diff
 fireEvent.focus(getByText('focus me');
--- a/docs/guide-events.md
+++ b/docs/guide-events.md
@@ -7,30 +7,41 @@ title: Considerations for fireEvent
 
 Based on [the Guiding Principles](guiding-principles.md), your test should
 resemble how users interact with your code (component, page, etc.) as much as
-possible. With this in mind, you should be aware that `fireEvent` is not actually
-how the user interacts with your application.
+possible. With this in mind, you should know that `fireEvent` isn't _exactly_
+how the user interacts with your application, but it's close enough for most
+scenarios.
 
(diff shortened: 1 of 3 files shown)


docs/guide-events.md
```diff
@@ -0,0 +1,55 @@
...
+id: guide-events
+title: Considerations for fireEvent
+---
+
+## Interactions vs. events
+
+Based on [the Guiding Principles](guiding-principles.md), your test should
+resemble how users interact with your code (component, page, etc.) as much as
+possible. With this in mind, you should be aware that `fireEvent` is not actually
+how the user interacts with your application.
```
kentcdodds: This is too strongly worded. In fact, none of what our tests are doing is how the user interacts with our app. We're just making trade-offs. Instead, you could say this:

```
With this in mind, you should know that `fireEvent` isn't _exactly_ how the user interacts with your application, but it's close enough for most scenarios.
```

docs/guide-events.md
```diff
@@ -0,0 +1,55 @@
...
+
+Depending on the interaction multiple events can be fired in the browser and sometimes
```
kentcdodds: Could we replace this paragraph and the next with this instead?

```
Consider `fireEvent.click` which creates a click event and dispatches that event on the given DOM node. This works properly for most situations when you simply want to test what happens when your element is clicked, but when the _user_ actually clicks your button, these are the events that are typically fired (in order):

- fireEvent.mouseOver(element)
- fireEvent.mouseMove(element)
- fireEvent.mouseDown(element)
- element.focus()
- fireEvent.mouseUp(element)
- fireEvent.click(element)

And then, if that element happens to be a child of a `label`, then it will also move focus to the form control that the label is labeling.

So even though all you really are trying to test is the click handler, by simply using `fireEvent.click` you're missing out on several other potentially important events the user is firing along the way. Again, most of the time this isn't critical for your tests and the trade-off of simply using `fireEvent.click` is worth it.
```

docs/guide-events.md
```diff
@@ -0,0 +1,55 @@
...
+
+You don't need to know every possible combination of interactions but you shouldn't
+have to much confidence in the interactive behavior of your components if you only
+test them with mocked sequences of events.
+
+## Alternatives
+
+We will describe a couple of simple adjustments to your tests that will increase your confidence
+in the interactive behavior of your components. For other interactions you should
+either consider using `user-event` or testing your components in a real environment
```
kentcdodds: Could we make `user-event` a link to the `user-event` docs page?

docs/guide-events.md
```diff
@@ -0,0 +1,55 @@
...
+was pressed.
+
+You don't need to know every possible combination of interactions but you shouldn't
+have to much confidence in the interactive behavior of your components if you only
+test them with mocked sequences of events.
+
+## Alternatives
+
+We will describe a couple of simple adjustments to your tests that will increase your confidence
+in the interactive behavior of your components. For other interactions you should
```
kentcdodds: Replace `should` with `may want to`.

I really don't want people reading this page and thinking: "Well, what the heck, why should I even bother with `fireEvent` at all then?!"

`fireEvent` is sufficient for most use cases, especially for beginners.

docs/guide-events.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Focus/Blur
+
+If an element is focused not only a focus event is dispatched. The active element
+in the document also changes as well as the previously focused element getting blurred.
+To simulate this behavior you can simply replace `fireEvent` with imperative focus:
+
+```diff
+fireEvent.focus(getByText('focus me');
+getByText('focus me').focus();
```
kentcdodds: Don't forget to add the `-` and `+` for this diff.

eps1lon: Done

docs/guide-events.md
eps1lon: Agreed. Super valuable do have an experienced, native speaking technical writer review this. I tend to come over a bit over-aggressive :smile:

docs/guide-events.md
eps1lon: done

docs/guide-events.md
eps1lon: Great stuff. Though we start by mentioning a button and then transition to the child of a label. I replaced button with a generic element since I never encountered buttons as children of labels.

docs/guide-events.md
eps1lon: done

docs/guide-events.md
```diff
@@ -0,0 +1,69 @@
...
+want to test what happens when your element is clicked, but when the _user_
+actually clicks your element, these are the events that are typically fired (in
+order):
+
+- fireEvent.mouseOver(element)
+- fireEvent.mouseMove(element)
+- fireEvent.mouseDown(element)
+- element.focus() (if that element is focusable)
+- fireEvent.mouseUp(element)
+- fireEvent.click(element) And then, if that element happens to be a child of a
```
kentcdodds: Could you put the paragraph on a new line rather than as part of the list?

docs/guide-events.md
```diff
@@ -0,0 +1,69 @@
...
+  `label`, then it will also move focus to the form control that the label is
+  labeling. So even though all you really are trying to test is the click
+  handler, by simply using `fireEvent.click` you're missing out on several other
+  potentially important events the user is firing along the way. Again, most of
```
kentcdodds: Could you put "Again" on a new line so it's more of a prominent summary?

docs/guide-events.md
eps1lon: Done

docs/guide-events.md
eps1lon: done

PR #193, opened 2019-07-25
refactoring asFragment example to use hooks

pedroapfilho:
All other examples are using hooks, so I refactored this one to use hooks too.


kentcdodds:
@all-contributors please add @pedroapfilho for docs

PR #197, opened 2019-07-27
run format command before build for consistency

alexkrolick:
Since a lot of people are committing on GitHub directly, there are lots of pages that haven't had the format command run over them. This PR adds the format command before pushing to netlify. This won't check in changes, but it will catch formatting inconsistency in code blocks when published.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -11,7 +11,7 @@
   publish = "website/build/testing-library-docs"
 
   # Default build command.
-  command = "cd website; npm install; npm run build"
+  command = "cd website; npm install; npm run format-docs; npm run build"
 
 # React root to React landing page
 [[redirects]]


netlify.toml
```diff
@@ -11,7 +11,7 @@
   publish = "website/build/testing-library-docs"
 
   # Default build command.
-  command = "cd website; npm install; npm run build"
+  command = "cd website; npm install; npm run format-docs; npm run build"
```
afontcu: Question! Won't this create a diferent version of deployed docs than the one on git/GitHub?

alexkrolick: Yeah, but it's pretty harmless I think. The alternative is to put the format command as a CI check, but it's hard to actually fix up prettier differences on GitHub.

alexkrolick: I figure we can periodically run the `format-docs` command to clean up.

afontcu: Yep, agree.. at the end of the day, people should just fork and edit docs locally, so git hooks autofix everything. But yeah, I can see how a "safety check" right before deploying could help. The only risk is that formatting ends up breaking the docs, but this feels kinda unlikely.

alexkrolick: Yeah, I would think the formatting would break the build rather than breaking the docs silently after the build.

PR #200, opened 2019-07-31
docs: add typescript config update

tanguyantoine:
add extra block for typescript users


kentcdodds:
@all-contributors please add @tanguyantoine for docs

PR #202, opened 2019-08-01
updated example code per issue

Agentkma:
**Please Review**

Searched for the ' _expect(container.firstChild).toMatchSnapshot()_ ' in the docs and found 2 examples.  Replaced both with ' _expect(container).toMatchSnapshot()_ '

### DIFF
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -127,7 +127,7 @@ expect(getByTestId('greeting-text')).toHaveTextContent('hello there')
 expect(getByTestId('ok-button')).toHaveAttribute('disabled')
 
 // snapshots work great with regular DOM nodes!
-expect(container.firstChild).toMatchSnapshot()
+expect(container).toMatchSnapshot()
 
 // you can also use get a `DocumentFragment`,
 // which is useful if you want to compare nodes across render
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -120,14 +120,50 @@ const greetingTextNode = await waitForElement(() =>
 
 ### Assert
 
+fetch.js
+
 ```jsx
+import React,{useState} from 'react'
+
+export default function Fetch({url}) {
+  const {greeting, setGreeting} = useState('')
+  const {buttonClicked, setButtonClicked} = useState(false)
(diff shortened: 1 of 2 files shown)


docs/react-testing-library/example-intro.md
```diff
@@ -127,7 +127,7 @@ expect(getByTestId('greeting-text')).toHaveTextContent('hello there')
 expect(getByTestId('ok-button')).toHaveAttribute('disabled')
 
 // snapshots work great with regular DOM nodes!
-expect(container.firstChild).toMatchSnapshot()
+expect(container).toMatchSnapshot()
```
kentcdodds: Actually, while we're at it, maybe we should also swap `.toMatchSnapshot` with `.toMatchInlineSnapshot`. This will help to encourage people to keep snapshots smaller. This will require that we figure out what the snapshot will really look like so we can inline it. You up for that @Agentkma? If you need help let us know.

Agentkma: Yes, I'm up for it. I will look into and let you know if I need help.  I've just started learning and using this library, so this is a great opportunity to learn while helping improve things.  Win-Win in my book :)

Agentkma: Ok, so would we want to show a code snippet/or full component of the <Fetch/> component(like the Svelt Testing example does?) to show how we find what the string arg to .toMatchInlineSnapshot() ? 
Screen Shot 2019-08-01 at 8 18 54 AM

kentcdodds: Yeah, that sounds good. It'll probably need to be something like:

```javascript
expect(container).toMatchInlineSnapshot(`
  <div>
    <h1>
      Hello
      {name}
      !
    </h1>
  </div>
`)
```

I'm not 100% certain that's how it will be serialized, but I'm pretty sure that's what it would look like.

Agentkma:
Is this in line with what you are looking for @kentcdodds ?  Thanks

kentcdodds:
@all-contributors please add @Agentkma for docs

osinjoku:
Is this how this page is supposed to look? It looks like it interpreted the HTML in the snapshot.
image

kentcdodds:
Whoops!

PR #208, opened 2019-08-08
Remove unnecessary whitespace in example-intro.md

chaficnajjar:
Replace `<button disabled >` with `<button disabled>`.


afontcu:
Hi! Thank you for your PR 😄 

Have you checked that the inline snapshot doesn't actually render that whitespace (can't check it myself right now)?

PR #209, opened 2019-08-08
Add Textlint

afontcu:
As suggested in #207, I'm opening up a draft PR to showcase `textlint` results.

There are several things I wanted to point out.

First of all, you can use commit history to understand the impact of each rule. Notice how some of the modifications are not automatic, but have been fixed manually.

Secondly, I've added "stop-words" and "write-good" rules as warnings. These are not autofixable, and I don't want to overcomplicate things for new contributors. Even though they require manual action, I believe the suggestions they yield are quite good. 

imatge

Finally, I've created a custom terminology list based on common mistakes.


docs/dom-testing-library/api-events.md
```diff
@@ -85,8 +85,8 @@ createEvent[eventName](node: HTMLElement, eventProperties: Object)
 
 Convenience methods for creating DOM events that can then be fired by
 `fireEvent`, allowing you to have a reference to the event created: this might
-be useful if you need to access event properties that cannot be initiated
-programmatically (such as `timeStamp`).
+be useful to access event properties that cannot be initiated programmatically
```
alexkrolick: Not sure I understand the removal of "if you need" and similar statements throughout. What rule is causing that?

afontcu: It is due to `textlint-rule-stop-words`

source: https://github.com/sapegin/textlint-rule-stop-words/blob/master/dict.txt

alexkrolick: Hmm, can we hold off on that rule for now? The "if you need/want" phrases seem to be the only things getting picked up by the rule so far and the usages would need to be rephrased a bit more carefully than what autocorrect implied.

cc @kentcdodds

website/.textlintrc.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    'write-good': {
+      severity: 'warning',
+    },
+    terminology: {
+      defaultTerms: false,
+      terms: [
+        ['built ?in', 'built-in'],
+        ['end[-]to[-]end', 'end-to-end'],
+        ['environemnt(s?)', 'environment$1'],
```
alexkrolick: 🤣

website/.textlintrc.js
```diff
@@ -0,0 +1,41 @@
...
+        ['e2e', 'end-to-end'],
+        ['dom[-]testing[-]library', 'DOM Testing Library'],
```
alexkrolick: 👍

website/package.json
```diff
@@ -10,18 +10,30 @@
...
+    "textlint-rule-en-capitalization": "^2.0.2",
+    "textlint-rule-stop-words": "^1.0.17",
+    "textlint-rule-terminology": "^1.1.30",
+    "textlint-rule-write-good": "^1.6.2"
   },
   "lint-staged": {
     "linters": {
       "*.js": ["prettier --write", "git add"],
-      "../docs/**/*.md": ["prettier --write --parser=markdown", "git add"],
+      "../docs/**/*.md": [
```
alexkrolick: Can this also run in CI?

docs/dom-testing-library/api-events.md
afontcu: Notice that only the rules with "`>`" are autofixed. The other ones remain as warnings (check out the screenshot on the PR description).

But of course no problem on taking it back

website/.textlintrc.js
afontcu: been there, done that 😇 😂

afontcu:
This didn't gain much traction and I don't have strong opinions on this, so let's put it on hold for now 👍

PR #213, opened 2019-08-09
Remove cleanup references

dustinmyers:
This pull request removes any instances of `cleanup-after-each` and `afterEach(cleanup)` in the react docs and examples. From issue #212

### DIFF
--- a/docs/react-testing-library/api.md
+++ b/docs/react-testing-library/api.md
@@ -254,10 +254,14 @@ Unmounts React trees that were mounted with [render](#render).
 > using supports the `afterEach` global (like mocha, Jest, and Jasmine). If not,
 > you will need to do manual cleanups after each test.
 
+For example, if you're using the [ava](https://github.com/avajs/ava) testing
+framework, then you would need to use the `test.afterEach` hook like so:
+
 ```jsx
 import { cleanup, render } from '@testing-library/react'
+import test from 'ava'
 
-afterEach(cleanup) // <-- add this
+test.afterEach(cleanup)
 
 test('renders into document', () => {
   render(<div />)
--- a/docs/react-testing-library/api.md
+++ b/docs/react-testing-library/api.md
@@ -275,11 +275,6 @@ Failing to call `cleanup` when you've called `render` could result in a memory
 leak and tests which are not "idempotent" (which can lead to difficult to debug
 errors in your tests).
 
-**If you don't want to add this to _every single test file_** then we recommend
-that you configure your test framework to run a file before your tests which
-does this automatically. See the [setup](./setup) section for guidance on how to
-set up your framework.
-
 ---
 
 ## `act`
--- a/docs/react-testing-library/api.md
+++ b/docs/react-testing-library/api.md
@@ -246,6 +246,38 @@ expect(firstRender).toMatchDiffSnapshot(asFragment())
 
 ---
 
+## `cleanup`
+
+Unmounts React trees that were mounted with [render](#render).
+
+> Please note that this is done automatically if the testing framework you're
+> using supports the `afterEach` global (like mocha, Jest, and Jasmine). If not,
+> you will need to do manual cleanups after each test.
+
+```jsx
+import { cleanup, render } from '@testing-library/react'
+
+afterEach(cleanup) // <-- add this
+
+test('renders into document', () => {
+  render(<div />)
+  // ...
+})
+
+// ... more tests ...
+```
+
+Failing to call `cleanup` when you've called `render` could result in a memory
+leak and tests which are not "idempotent" (which can lead to difficult to debug
+errors in your tests).
+
+**If you don't want to add this to _every single test file_** then we recommend
+that you configure your test framework to run a file before your tests which
+does this automatically. See the [setup](./setup) section for guidance on how to
+set up your framework.
+
+---
+
 ## `act`
 
 This is a light wrapper around the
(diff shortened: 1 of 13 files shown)


docs/react-testing-library/api.md
```diff
@@ -252,34 +246,6 @@ expect(firstRender).toMatchDiffSnapshot(asFragment())
 
 ---
 
-## `cleanup`
```
dustinmyers: I removed this full section as it didn't seem pertinent anymore.I think there may be an argument for keeping it here though, just to show how to use `afterEach(cleanup)`

kentcdodds: Let's keep this around, just modify it to explain why it exists and when you should use it (like if you're using a testing framework like `tape` or `ava`).

docs/react-testing-library/setup.md
```diff
@@ -328,3 +273,13 @@ mocha --require jsdom-global/register
 Note, depending on the version of Node you're running, you may also need to
 install @babel/polyfill (if you're using babel 7) or babel-polyfill (for babel
 6).
+
+### Skipping Auto Cleanup
+
+[`Cleanup`](./api#cleanup) is called after each test automatically by default.
```
kentcdodds:
```suggestion
[`Cleanup`](./api#cleanup) is called after each test automatically by default if the testing framework you're using supports the `afterEach` global (like mocha, Jest, and Jasmine).
```

docs/react-testing-library/setup.md
```diff
@@ -328,3 +273,13 @@ mocha --require jsdom-global/register
...
+However, you may choose to skip the auto cleanup by add setting the
+RTL_SKIP_AUTO_CLEANUP env variable to 'true'.
```
kentcdodds:
```suggestion
`RTL_SKIP_AUTO_CLEANUP` env variable to 'true'.
```

docs/react-testing-library/setup.md
```diff
@@ -328,3 +273,13 @@ mocha --require jsdom-global/register
...
+
+```javascript
```
kentcdodds:
````suggestion
```
````

File: docs/react-testing-library/setup.md [hunk 5]
```diff
@@ -328,3 +273,13 @@ mocha --require jsdom-global/register
...
+RTL_SKIP_AUTO_CLEANUP = 'true'
```
kentcdodds: Let's add a reference to [`cross-env`](https://github.com/kentcdodds/cross-env) here and suggest that they can put this in their test script.

Then this line can be:

```suggestion
cross-env RTL_SKIP_AUTO_CLEANUP=true jest
```

dustinmyers: Okay. How would you word that?

dustinmyers: Can you help me out with the wording? Not sure how this would be done, so it will be hard to suggest :)

docs/react-testing-library/setup.md
```diff
@@ -328,3 +273,14 @@ mocha --require jsdom-global/register
 Note, depending on the version of Node you're running, you may also need to
 install @babel/polyfill (if you're using babel 7) or babel-polyfill (for babel
 6).
+
+### Skipping Auto Cleanup
+
+[`Cleanup`](./api#cleanup) is called after each test automatically by default if
+the testing framework you're using supports the `afterEach` global (like mocha,
+Jest, and Jasmine). However, you may choose to skip the auto cleanup by setting
+the `RTL_SKIP_AUTO_CLEANUP` env variable to 'true'.
```
kentcdodds:
```suggestion
the `RTL_SKIP_AUTO_CLEANUP` env variable to 'true'. You can do this with
[`cross-env`](https://github.com/kentcdodds/cross-env) like so:
```

docs/react-testing-library/api.md
```diff
@@ -256,6 +250,10 @@ expect(firstRender).toMatchDiffSnapshot(asFragment())
 
 Unmounts React trees that were mounted with [render](#render).
 
+> Please note that this is done automatically if the testing framework you're
+> using supports the `afterEach` global (like mocha, Jest, and Jasmine). If not,
+> you will need to do manual cleanups after each test.
+
 ```jsx
```
kentcdodds: Could we change this to an example of using AVA? It could look like this:

````md

For example, if you're using the [ava](https://github.com/avajs/ava) testing framework, then you would need to use the `test.afterEach` hook like so:

```jsx
import {cleanup, render} from '@testing-library/react'
import test from 'ava'

test.afterEach(cleanup)

test('your test', () => {})
```
````

File: docs/react-testing-library/api.md [hunk 8]
```diff
@@ -256,10 +250,18 @@ expect(firstRender).toMatchDiffSnapshot(asFragment())
...
 import { cleanup, render } from '@testing-library/react'
+import test from 'ava'
 
-afterEach(cleanup) // <-- add this
+test.afterEach(cleanup)
 
 test('renders into document', () => {
   render(<div />)
```
kentcdodds: I think this bit:

```
**If you don't want to add this to _every single test file_** then we recommend
that you configure your test framework to run a file before your tests which
does this automatically. See the [setup](./setup) section for guidance on how to
set up your framework.
```

can be removed.

dustinmyers: Okay. I was wondering about that. I'll remove it.

kentcdodds:
@all-contributors please add @dustinmyers for docs

PR #223, opened 2019-08-18
Update tsconfig example in cypress docs

simjes:
Update `tsconfig.json` in README to reflect that `@types/testing-library__cypress` is a dependency, and should not have to be installed by the user.

Related: testing-library/cypress-testing-library#76


afontcu:
I was about to write the same comment Kent added [here](https://github.com/testing-library/cypress-testing-library/pull/76#pullrequestreview-276279766) :)

simjes:
Closing, see discussion in testing-library/cypress-testing-library#76

PR #225, opened 2019-08-20
all-contributors/example-function-agentkma

Agentkma:
added anonymous function to onClick call for fetchGreeting and passed in the url prop as an argument

### DIFF
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -122,7 +122,7 @@ export default function Fetch({url}) {
   const {greeting, setGreeting} = useState('')
   const {buttonClicked, setButtonClicked} = useState(false)
   
-  const fetchGreeting = async url => {
+  const fetchGreeting = async () => {
     const data = await fetch(url)
     const {greeting} = data
     setGreeting(greeting)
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/example-intro.md
```diff
@@ -134,7 +134,7 @@ export default function Fetch({url}) {
   return (
     <div>
       <button 
-        onClick={fetchGreeting} 
+        onClick={()=>fetchGreeting(url)}
```
kentcdodds: What if instead, we leave this line as-is, and remove the `url` parameter from `fetchGreeting` above.

Agentkma: Hi there @kentcdodds ,  I added the url b/c I saw in the example tests where the <Fetch url={url} />....so I was trying to reverse engineer the fetch.js component to match up this test example 
image  Let me know what you think. Thanks

kentcdodds: Right, my suggestion will work with that as well :) What you have here is functionally equivalent to what I'm suggesting. The difference is that what's here right now is creating an extra function unnecessarily.

Agentkma: Yes, I misunderstood...I see that we don't need to add url as a param to fetchGreeting because we have access to that prop inside the function already...got it...will update.  Thanks @kentcdodds , overlooked the obvious here :)

kentcdodds:
@all-contributors please add @Agentkma for docs

Agentkma:
You are welcome, and thank you sir! @kentcdodds

PR #226, opened 2019-08-20
add support page

kentcdodds:
@alexkrolick and I have discussed things a bit and we think it'd be better to get rid of the opencollective and instead have this page for people to support the project.

Here's what it looks like:

Screen Shot 2019-08-20 at 9 53 29 AM

Let me know what changes you'd like to make. I'll note that I'm happy to add other courses/etc. on this page for anyone who's contributed a nontrivial amount to any of the Testing Library projects.


website/pages/en/support.js
```diff
@@ -0,0 +1,100 @@
...
+            Thanks! The Testing Library maintainers are happy to maintain this
+            library along with you, the community. We aren't looking for
+            funding, but we need everyone to pitch in to make this project and
+            community successful and improve tests for everyone in the long run.
+            With that in mind, here are a few ways you can help support this
+            effort.
+          </p>
+          <GridBlock
+            contents={supportLinks}
+            layout="threeColumn"
```
alexkrolick: The layout adds some unneeded left-margin on the second row, might be nice to fix as a followup

PR #228, opened 2019-08-22
docs: update dom-testing-library/install.md to state react as library

rehan-sattar:
Since react is a library not a framework, so it should be stated in the docs. Otherwise there is a chance of misconception.


afontcu:
Thanks!

PR #229, opened 2019-08-25
docs: pass options into render function

brycegit:
One of the customRender examples is accepting an options argument but not passing it to the rtl render function. This just adds it in.


alexkrolick:
🥇 thanks

PR #235, opened 2019-09-01
docs: use introduction in the side bar

timdeschryver:
Most of the libraries use `Introduction` in the sidebar, this PR updates the ones that were using `Intro`.


kentcdodds:
Thanks!

PR #239, opened 2019-09-01
Remove empty code block

aryzing:


aryzing:
It appears the file has been this way since its inception, the only modifications are a file rename with no modifications and an update to import statements.

PR #243, opened 2019-09-05
Update example-intro.md example

aayushrajvanshi:


docs/react-testing-library/example-intro.md
```diff
@@ -117,13 +117,15 @@ fetch.js
...
 export default function Fetch({url}) {
-  const {greeting, setGreeting} = useState('')
-  const {buttonClicked, setButtonClicked} = useState(false)
+  const [greeting, setGreeting] = useState('')
+  const [buttonClicked, setButtonClicked] = useState(false)
   
   const fetchGreeting = async () => {
-    const data = await fetch(url)
+    const response = await axios.get(url)
+    const data = await response.data
```
kentcdodds: I don't think we need `await` here.

aayushrajvanshi: Oh yeah! Forget to remove as it had "fetch" before so were using for response.json().

kentcdodds:
@all-contributors please add @aayushrajvanshi for docs

PR #250, opened 2019-09-06
Replace evergreen icon with cypress icon on website

jennifer-shehane:
I get that, maybe this is the logo for the ‘cypress-tesing-library’ itself? But the puppeteer and testcafe testing libraries show the base testing libraries’ logos, so we thought it made since to display Cypress’s logo here.


kentcdodds:
@all-contributors please add @jennifer-shehane for docs

PR #251, opened 2019-09-06
Minor fixes to cypress-testing-library docs

jennifer-shehane:
- Singular entities that end in an “s” character use “’s” when showing
possession, not “‘“
- Add proper casing of TypeScript
- lowercase ‘Library’


kentcdodds:
Thank you!

PR #261, opened 2019-09-16
docs(react-router): examples for react-router

alexkrolick:
- extract out "renderWithRouter" helper and explain when it is needed
- add some "getBy" queries instead of just container.innerHTML assertions, to show more idiomatic RTL usage

### DIFF
--- a/docs/example-react-router.md
+++ b/docs/example-react-router.md
@@ -56,7 +56,7 @@ test('full app rendering/navigating', () => {
 })
 
 test('landing on a bad page shows 404 page', () => {
-  window.history.pushState({}, '', '/some/bad/route');
+  window.history.pushState({}, '', '/some/bad/route')
   const { getByRole } = render(
     <Router>
       <App />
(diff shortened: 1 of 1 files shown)


docs/example-react-router.md
```diff
@@ -32,11 +33,57 @@ function App() {
...
+    <Router>
+      <App />
+    </Router>
+  )
+  expect(getByRole('heading')).toHaveTextContent('404 Not Found')
+})
+
+test('rendering a component that uses withRouter', () => {
+  const route = '/some-route'
+  window.history.pushState({}, '', route);
```
afontcu:
```suggestion
  window.history.pushState({}, '', route)
```

(same thing in the `window.history...` statement above)

alexkrolick:
> Looking good! Could we take this as an opportunity to encourage people to use wrapper?

let's show 
1. inline
2. wrapper
3. helper with wrapper

docs/example-react-router.md
```diff
@@ -32,11 +32,72 @@ function App() {
...
+  expect(getByRole('heading')).toMatch('Home')
+})
+```
+
+2. If you find yourself adding Router components to your tests a lot, you may want to create
+a helper function that wraps around `render`. 
+
+```jsx
+// test utils file
 function renderWithRouter(
```
kentcdodds: Let's use the wrapper for this as well so people get rerender working by default.

PR #266, opened 2019-09-21
feat(queries): Add docs about byRole#hidden

eps1lon:
Documentation for https://github.com/testing-library/dom-testing-library/pull/352

Preview: https://deploy-preview-266--testing-library.netlify.com/docs/dom-testing-library/api-queries#byrole

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -533,6 +533,7 @@ getByRole(
   role: TextMatch,
   options?: {
     exact?: boolean = true,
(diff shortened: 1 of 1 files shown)


eps1lon:
I missed mentioning this in the original PR. Could anybody from @testing-library/dom take a look if the wording is ok?

eps1lon:
> Can you add an example of when an element would be excluded by default and included by hidden: true?

Good idea. Added a more common example with modals and `aria-hidden`

alexkrolick:
💯

PR #270, opened 2019-09-25
Fix Cypress example URL and example in cypress-testing-library/intro.md

HTMLhead:
Hi, I explained the problem at issue #269 

So I tried to change the URL right place and change examples in there. But I'm not sure if my example is correct.

I look forward to hearing from you.

Thanks


kentcdodds:
@all-contributors please add @HTMLhead for docs

PR #279, opened 2019-10-03
Documentation: native-testing-library fix preset name

olethanh:
the preset name was wrong, probably a bad replace  somewhere?


bcarroll22:
@allcontributors please add @olethanh for doc

PR #283, opened 2019-10-06
Fix typo

pavelloz:


afontcu:
Hi! Thanks for this. actually, we ditched “dom-testing-library” in favor of “DOM Testing Library”. Would you mind changing that too? :)

docs/testcafe-testing-library/intro.md
```diff
@@ -69,7 +69,7 @@ test('getByLabelText', async t => {
 ```
 
 ## Configure
-You can customize the testIdAttribute using the [configure function of dom-testing-librarty][config] in a few different ways:
+You can customize the testIdAttribute using the [configure function of dom-testing-library][config] in a few different ways:
```
pavelloz:
```suggestion
You can customize the testIdAttribute using the [configure function of DOM Testing Library][config] in a few different ways:
```

pavelloz:
Of course, nowadays with GitHub inline editing we can do so much with so little effort :)

PR #284, opened 2019-10-07
feat: Add isInaccessible

eps1lon:
Documentation for https://github.com/testing-library/dom-testing-library/pull/364

### DIFF
--- a/docs/dom-testing-library/api-helpers.md
+++ b/docs/dom-testing-library/api-helpers.md
@@ -186,7 +186,7 @@ console.log(getRoles(nav))
 // }
 ```
 
-## `shouldExcludeFromA11yTree`
+## `isInaccessible`
 
 This function will compute if the given element should be excluded from the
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-helpers.md
```diff
@@ -185,6 +186,19 @@ console.log(getRoles(nav))
 // }
 ```
 
+## `shouldExcludeFromA11yTree`
+
+This function will compute if the given element should be excluded from the
+accessibility API by the browser. It implements every **MUST** criteria from
+https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion with the exception of
```
afontcu:
```suggestion
[Accessibility Tree spec](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion) with the exception of
```

eps1lon: Thank you for the review. I've been c&p too many spec links these days and got tunnel vision.

PR #285, opened 2019-10-08
removed the example for getting by label

### LINKED ISSUE #236: Unclear example for `ByLabelText`
**Describe the bug**
The example provided in the [`ByLabelText`](https://testing-library.com/docs/dom-testing-library/api-queries#bylabeltext) section does not seem to match the surrounding description

**To Reproduce**
If you go to the link above, the surrounding description states that 

> The example below will find the input node for the following DOM structures

yet there is no input node in the following part of the code snippet in the example

```html
// The aria-labelledby attribute with non-form elements
<section aria-labelledby="section-one-header">
  <h3 id="section-one-header">Section One</h3>
  <p>some content</p>
</section>
```

**Expected behavior**
Unsure, although was expecting to see some kind of input element in the snippet above.

sdowding:
for a non input html because the javascript example does not use it.

Closes #236

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -87,12 +87,6 @@ The example below will find the input node for the following DOM structures:
 <label id="username-label">Username</label>
 <input aria-labelledby="username-label" />
 
-// The aria-labelledby attribute with non-form elements
-<section aria-labelledby="section-one-header">
-  <h3 id="section-one-header">Section One</h3>
-  <p>some content</p>
-</section>
-
 // Wrapper labels
 <label>Username <input /></label>
 


sdowding:
this should resolve #236

alexkrolick:
Can we fix the explanatory text instead? This is a useful snippet for aria-labelledby usage.

eps1lon:
> Can we fix the explanatory text instead? This is a useful snippet for aria-labelledby usage.

I think for non-form elements we should recommend `ByRole` e.g. `getByRole('section', { name: 'Section one' })`. "Label" is associated with a `<label>` element and is "just" an alias for "name". Since we want to get rid of as many aliases as possible we should phase `ByLabelText` for aria-attributes out. ByLabelText should really only consider actual <label> elements since those have UX implications (clicking label focuses input).

afontcu:
Given the explanation @eps1lon shared above, I think this PR is worth merging

PR #294, opened 2019-10-10
Adding Wayfair to list of users

ErezNagar:


ErezNagar:
Sorry but this might take a while. Closing the PR for now, will open a PR when I hear back.

PR #300, opened 2019-10-17
Update cypress configure documentation

homburg:
`configure` is exported from `@testing-library/cypress` as of https://github.com/testing-library/cypress-testing-library/pull/81 / version 4.2.0

Update documentation to describe `configure` usage with cypress.


alexkrolick:
Thanks!

PR #305, opened 2019-10-20
docs: Prefer byRole over byTestId where possible

eps1lon:
This is a collection of changes that tries to reduce the amount of `byTestId` usages.

`byTestId` is especially problematic for component libraries since the markup is very often not simple (e.g. includes wrapper divs) and users often ask where they should put the testId or claim the library in question is not very well testable with `@testing-library/*`. 

In many cases a `byRole` query would have resolved those questions.

It is probably a good idea to also add a table to the docs explaining which HTML elements map to which ARIA role. I'd like to to do this in a separate PR since there are some unresolved issues with `aria-query` and we need to figure out where the source of truth is (aria-query or html-aam).

### DIFF
--- a/docs/dom-testing-library/faq.md
+++ b/docs/dom-testing-library/faq.md
@@ -76,20 +76,20 @@ in the selector.
 const thirdLiInUl = container.querySelector('ul > li:nth-child(3)')
 ```
 
-Or you could include the index or an ID in your attribute:
+Or you could query the `listitem` role and access the index in question:
 
 ```javascript
-;`<li data-testid="item-${item.id}">{item.text}</li>`
+;`<li>{item.text}</li>`
 ```
 
-And then you could use the `getByTestId` utility:
+And then you could use the `getAllByRole` utility:
 
 ```javascript
 const items = [
   /* your items */
 ]
 const container = render(/* however you render this stuff */)
-const thirdItem = getByTestId(container, `item-${items[2].id}`)
+const thirdItem = getAllByRole(container, 'listitem')[2]
 ```
 
 </details>
--- a/docs/dom-testing-library/faq.md
+++ b/docs/dom-testing-library/faq.md
@@ -76,13 +76,8 @@ in the selector.
 const thirdLiInUl = container.querySelector('ul > li:nth-child(3)')
 ```
 
-Or you could query the `listitem` role and access the index in question:
-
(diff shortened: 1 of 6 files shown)


docs/dom-testing-library/faq.md
```diff
@@ -76,20 +76,20 @@ in the selector.
...
-Or you could include the index or an ID in your attribute:
+Or you could query the `listitem` role and access the index in question:
 
 ```javascript
-;`<li data-testid="item-${item.id}">{item.text}</li>`
+;`<li>{item.text}</li>`
 ```
 
-And then you could use the `getByTestId` utility:
+And then you could use the `getAllByRole` utility:
```
afontcu: I believe this sentence belongs to the one above. It made sense when we first added the `testid` attribute and then showed how to query it, but since the attribute is gone, I believe the whole content could be collapse into a single paragraph:

```
Or you could use the `getAllByRole` utility querying by 
the `listitem` role and access the index in question:

const items = [ 
  /* your items */  
]
const container = render(/* however you render this stuff */) 
const thirdItem = getAllByRole(container, 'listitem')[2] 
```

thoughts?

eps1lon: Absolutely :+1: Thanks.

eps1lon: > Or you could use getAllByRole to query the listitem role and access the index in question:

https://deploy-preview-305--testing-library.netlify.com/docs/dom-testing-library/faq

docs/ecosystem-user-event.md
```diff
@@ -16,12 +16,12 @@ import React from 'react'
 import { render } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
-test('Fill text area', () => {
-  const { getByTestId } = render(<textarea data-testid="email" />)
-
-  userEvent.type(getByTestId('email'), 'Hello, World!')
-  expect(getByTestId('email')).toHaveAttribute('value', 'Hello, World!')
+const { getByRole } = test('click', () => {
+  render(<textarea />)
```
afontcu: Looks like the destructuring statement jumped to the line above:

```js
test('click', () => { 
  const { getByRole } = render(<textarea />) 
```

eps1lon: Done

PR #315, opened 2019-10-30
Fixing modal example to use hooks

pedroapfilho:
I just updated the modal example to use hooks, nothing much. It's the last example using a class component.

### DIFF
--- a/docs/example-react-modal.md
+++ b/docs/example-react-modal.md
@@ -5,7 +5,7 @@ title: Modals
 
 ```jsx
 import React, { useEffect } from 'react'
-import ReactDOM from 'react-dom'
+import { render, fireEvent } from '@testing-library/react'
 
 const modalRoot = document.createElement('div')
 modalRoot.setAttribute('id', 'modal-root')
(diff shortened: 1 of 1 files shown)


docs/example-react-modal.md
```diff
@@ -4,63 +4,64 @@ title: Modals
...
+        <hr />
+        <button onClick={onClose}>Close</button>
+      </div>
+    </div>,
+    el
+  )
 }
 
+const rootElement = document.getElementById('root')
+ReactDOM.render(<App />, rootElement)
```
kentcdodds: Oh, this shouldn't need to be here.

docs/example-react-modal.md
```diff
@@ -4,63 +4,64 @@ title: Modals
 ---
 
 ```jsx
-import React from 'react'
+import React, { useEffect } from 'react'
 import ReactDOM from 'react-dom'
-import { render, fireEvent } from '@testing-library/react'
```
kentcdodds: Why was this removed?

docs/example-react-modal.md
```diff
@@ -4,63 +4,64 @@ title: Modals
...
+      style={{
+        position: 'absolute',
+        top: '0',
+        bottom: '0',
+        left: '0',
+        right: '0',
+        display: 'grid',
+        justifyContent: 'center',
+        alignItems: 'center',
+        backgroundColor: 'rgba(0,0,0,0.3)',
```
afontcu: Since we're tweaking this example, would it make sense to remove stylistic CSS to make it more focused on the problem we're trying to solve?

PR #316, opened 2019-10-31
Add jest-environment-jsdom-fifteen recommendation

trurl-master:
Fix for https://github.com/testing-library/react-testing-library/issues/517

### DIFF
--- a/docs/react-testing-library/setup.md
+++ b/docs/react-testing-library/setup.md
@@ -249,7 +249,7 @@ NODE_PATH=src/utils
 If you're using the Jest testing framework version 24 or lower with the default
 configuration, it's recommended to use `jest-environment-jsdom-fifteen` package
 as Jest uses a version of the jsdom environment that misses some features and
-fixes, required by the `React Testing Library`
+fixes, required by React Testing Library.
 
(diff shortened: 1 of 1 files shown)


trurl-master:
@eps1lon Thank you for your comment! Please check the updated version

docs/react-testing-library/setup.md
```diff
@@ -244,6 +244,29 @@ $ app
 NODE_PATH=src/utils
 ```
 
+### Jest 24 (or lower) and defaults
+
+If you're using the Jest testing framework version 24 or lower with the default
+configuration, it's recommended to use `jest-environment-jsdom-fifteen` package
+as Jest uses a version of the jsdom environment that is not fully compatible
```
eps1lon: It's not about compatibility but about missing fixes/features in earlier versions. At least this applies to other testing-library packages (e.g. `@testing-library/dom`.

trurl-master: Sure. Please check the new version. If you have a better way of describing it, I'll gladly use it as my English skills is relatively limited. Thank you again for the feedback.

kentcdodds:
@all-contributors please add @trurl-master for docs

PR #322, opened 2019-11-08
update Cypress documentation for v5

SavePointSam:
This update brings the Cypress Testing Library documentation to be better in line with the current README of the project. It also removes references to the `get*` queries as there were [removed in v5](https://github.com/testing-library/cypress-testing-library/releases/tag/v5.0.0).

### DIFF
--- a/docs/cypress-testing-library/intro.md
+++ b/docs/cypress-testing-library/intro.md
@@ -26,6 +26,11 @@ You can now use all of `DOM Testing Library`'s `findBy`, `findAllBy`, `queryBy`
 and `queryAllBy` commands off the global `cy` object.
 [See the `DOM Testing Library` docs for reference](https://testing-library.com).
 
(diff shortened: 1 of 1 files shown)


docs/cypress-testing-library/intro.md
```diff
@@ -35,40 +39,30 @@ and should be added as follows in `tsconfig.json`:
 }
 ```
 
-You can now use all of `DOM Testing Library`'s `getBy`, `getAllBy`, `queryBy`
-and `queryAllBy` commands.
-[See `DOM Testing Library` API for reference](dom-testing-library/api-queries.md)
+You can find [all Library definitions here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__cypress/index.d.ts).
```
alexkrolick: Can we keep a link to the queries with an explanation of why `get` isn't exposed? TS defs are community maintained and might not be 100% up to date all the time (plus you need to understand Typescript).

PR #326, opened 2019-11-18
add Quizlet to list of users

turadg:
We started replacing Enzyme with testing-library at Quizlet and it's great!

This PR adds [Quizlet](https://quizlet.com) to the list of users.

---
This change is [Reviewable](https://reviewable.io/reviews/testing-library/testing-library-docs/326)


kentcdodds:
@all-contributors please add @turadg for docs

PR #331, opened 2019-12-02

arturoromeroslc:
The first example test did not pass because we expected the element to be a heading not a div. 
Next the assertion for a no match route is 404 Not Found


alexkrolick:
Something is up with the Docusaurus build

kentcdodds:
Let's just merge it and hope it works 🤞 If not then we can look into fixing things.

kentcdodds:
@all-contributors please add @arturoromeroslc for docs

kentcdodds:
Whoops, made that too soon.

kentcdodds:
@all-contributors please add @arturoromeroslc for docs

PR #349, opened 2019-12-12
add jest.mock() instruction

iswara108:
This code does not work without specifically instructing the manual mock.
It is only the case while automock is set to false (as it is by default, and create-react-app does not support overriding it)


kentcdodds:
@all-contributors please add @iswara108 for docs

PR #355, opened 2020-01-13
Improve usage instructions for jest-dom

PabloDinella:
People not familiar with jest matchers extensions might not know what to do after installing the lib, so this addition should avoid confusion.


PabloDinella:
Right, I changed it to point the user to jest-dom's doc

PR #358, opened 2020-01-15
Fix component name: from App to Modal

stefanwille:
This PR changes the example such that the component has the name that is referenced in the test.


alexkrolick:
Thanks!

PR #363, opened 2020-01-22
Update react-router example to match react-testing-library-examples

jermainezhimin:
To resolve https://github.com/testing-library/testing-library-docs/issues/361

### DIFF
--- a/docs/example-react-router.md
+++ b/docs/example-react-router.md
@@ -4,16 +4,17 @@ title: React Router
 ---
 
 ```jsx
-// app.js
 import React from 'react'
-import { withRouter } from 'react-router'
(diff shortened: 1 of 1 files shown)


alexkrolick:
This page is actually newer than the example from the repo and extended it. Can we fix the specific lines that have issues instead of going backwards?

jermainezhimin:
My intent was to match the format of `react-router` example with `reach-router`. Closing the PR because the page update is newer (_I misunderstood the initial intent_) 😅

alexkrolick:
You can still fix the example to make it work, by comparing the repos. But don't delete the page structure (headings and example text).

PR #366, opened 2020-01-24

pylnata:
React must be in scope when using JSX


kentcdodds:
@all-contributors please add @pylnata for docs

PR #368, opened 2020-01-25
feat(byRole): Add `name` option

eps1lon:
Docs for https://github.com/testing-library/dom-testing-library/pull/408

Preview:
- https://deploy-preview-368--testing-library.netlify.com/docs/dom-testing-library/api-queries#byrole
- https://deploy-preview-368--testing-library.netlify.com/docs/guide-which-query

I moved `byRole` up in the recommendations. I'm not so sure if we should already do this though. Personally I would hope that we can reduce the queries to `byRole` and `byTestId` and then recommend asserting properties such as `placeholder`, `alt`, `value` etc. Choosing the right query is currently not trivial. Though I fear that by only having `byRole` we loose a lot of people because they are not aware of the role names and fall back to testid. While I would hope that everybody knows what roles are it's not mainstream knowledge as far as I can tell. And `<input />` having so many roles with names that are not that common in frontend development. I can only speak for myself though (mentally it's still "`input` of type `text`" instead of just "a `textbox`"). So I'm really looking forward to input from others that have teaching experience or can quickly poll their teams.

I'm fine with leaving the `name` option a bit more obscure for a start and see how  adoption goes.

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -164,8 +164,11 @@ cy.getByLabelText('username').should('exist')
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
-It will NOT find the input node for label text broken up by elements. If it is
-important that you query an actual `<label>` element you can provide a
+It will NOT find the input node for label text broken up by elements. You can
+use `getByRole('textbox', { name: 'Username' })` instead which is robust against
+switching to `aria-label` or `aria-labelledby`.
+
+If it is important that you query an actual `<label>` element you can provide a
 `selector` in the options:
 
 ```html
@@ -178,9 +181,6 @@ const inputNode = getByLabelText(container, 'Username', {
   selector: 'input',
 })
 ```
-
-Otherwise `byRole('textbox', { name: 'Username' })` works as well and is robust
-against switching to `aria-label` or `aria-labelledby`.
 
 ### `ByPlaceholderText`
 
@@ -594,12 +594,12 @@ a specific element if multiple elements with the same role are present on the
 rendered content. For an in-depth guide check out
 ["What is an accessible name?" from ThePacielloGroup](https://developer.paciellogroup.com/blog/2017/04/what-is-an-accessible-name/).
 If you only query for a single element with `getByText('The name')` it's
-oftentimes better to use `byRole(expectedRole, { name: 'The name' })`. The
+oftentimes better to use `getByRole(expectedRole, { name: 'The name' })`. The
 accessible name query does not replace other queries such as `byAlt` or
 `byTitle`. While the accessible name can be equal to these attributes, it does
 not replace the functionality of these attributes. For example
 `<img aria-label="fancy image" src="fancy.jpg" />` will be returned for both
-`getByAlt('fancy image')` and `byRole('image', { name: 'fancy image' })`.
+`getByAlt('fancy image')` and `getByRole('image', { name: 'fancy image' })`.
 However, the image will not display its description if `fancy.jpg` could not be
 loaded. Whether you want assert this functionality in your test or not is up to
 you.
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -164,8 +164,9 @@ cy.getByLabelText('username').should('exist')
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
-It will NOT find the input node for label text broken up by elements. For this
-case, you can provide a `selector` in the options:
+It will NOT find the input node for label text broken up by elements. If it is
+important that you query an actual `<label>` element you can provide a
+`selector` in the options:
 
 ```html
 <label> <span>Username</span> <input /> </label>
@@ -177,6 +178,9 @@ const inputNode = getByLabelText(container, 'Username', {
   selector: 'input',
 })
 ```
+
+Otherwise `byRole('textbox', { name: 'Username' })` works as well and is robust
+against switching to `aria-label` or `aria-labelledby`.
 
 ### `ByPlaceholderText`
 
@@ -567,6 +571,7 @@ getByRole(
   options?: {
     exact?: boolean = true,
     hidden?: boolean = true,
+    name?: TextMatch,
     normalizer?: NormalizerFn,
   }): HTMLElement
 ```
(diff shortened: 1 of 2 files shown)


docs/dom-testing-library/api-queries.md
```diff
@@ -178,6 +179,9 @@ const inputNode = getByLabelText(container, 'Username', {
 })
 ```
 
+Otherwise `byRole('textbox', { name: 'Username' })` works as well and is robust
```
kentcdodds: Maybe move this up above the example and recommend it first. Also, use an actual query (not just `byRole` because that doesn't exist). Maybe `getByRole`?

docs/dom-testing-library/api-queries.md
```diff
@@ -581,6 +586,24 @@ Library uses `aria-query` under the hood to find those elements by their default
...
+[accessible name](https://www.w3.org/TR/accname-1.1/). The accessible name is
+for simple cases equal to e.g. the label of a form element, or the text content
+of a button, or the value of the `aria-label` attribute. It can be used to query
+a specific element if multiple elements with the same role are present on the
+rendered content. For an in-depth guide check out
+["What is an accessible name?" from ThePacielloGroup](https://developer.paciellogroup.com/blog/2017/04/what-is-an-accessible-name/).
+If you only query for a single element with `getByText('The name')` it's
+oftentimes better to use `byRole(expectedRole, { name: 'The name' })`. The
+accessible name query does not replace other queries such as `byAlt` or
+`byTitle`. While the accessible name can be equal to these attributes, it does
```
kentcdodds: Let's make sure that we never have sample code that looks like it could be real code but isn't. So either `getByRole...` or `*ByRole...`

docs/dom-testing-library/api-queries.md
```diff
@@ -581,6 +586,24 @@ Library uses `aria-query` under the hood to find those elements by their default
...
+not replace the functionality of these attributes. For example
+`<img aria-label="fancy image" src="fancy.jpg" />` will be returned for both
+`getByAlt('fancy image')` and `byRole('image', { name: 'fancy image' })`.
```
kentcdodds: Here as well.

docs/dom-testing-library/api-queries.md
eps1lon: Done

docs/dom-testing-library/api-queries.md
eps1lon: Done

docs/dom-testing-library/api-queries.md
eps1lon: Done

docs/dom-testing-library/api-queries.md
```diff
@@ -581,6 +586,24 @@ Library uses `aria-query` under the hood to find those elements by their default
...
+[accessible name](https://www.w3.org/TR/accname-1.1/). The accessible name is
+for simple cases equal to e.g. the label of a form element, or the text content
+of a button, or the value of the `aria-label` attribute. It can be used to query
+a specific element if multiple elements with the same role are present on the
+rendered content. For an in-depth guide check out
+["What is an accessible name?" from ThePacielloGroup](https://developer.paciellogroup.com/blog/2017/04/what-is-an-accessible-name/).
+If you only query for a single element with `getByText('The name')` it's
+oftentimes better to use `getByRole(expectedRole, { name: 'The name' })`. The
+accessible name query does not replace other queries such as `byAlt` or
+`byTitle`. While the accessible name can be equal to these attributes, it does
```
kentcdodds: These should both be `*ByAlt` and `*ByTitle`

docs/guide-which-query.md
```diff
@@ -14,6 +14,11 @@ possible. With this in mind, we recommend this order of priority:
    1. `getByLabelText`: Only really good for form fields, but this is the number
       one method a user finds those elements, so it should be your top
       preference.
+   1. `getByRole`: With the `name` option you can basically query the full
+      accessibility tree. It can query all the elements you can query with the
+      other accessility query. But consider that you still might want to assert
+      if an element has an actual `<label>` element or a `placeholder`
+      attribute.
```
kentcdodds: I think many people use `*ByText` to select `button`s, `div`s, and `span`s. Maybe we should briefly mention what the role for each of those would be and that will cover most use-cases.

eps1lon: I'm not encountering much use cases to select `div` or `span` these days. These wouldn't be queryable with `getByRole` and don't have an accessible name to begin with. 

After considering the feedback in the implementation PR I would actually agree that `byText('some text', { role: 'listitem' })` would be a useful query. Simply because some roles can't have an accessible name and the text (i.e. `.textContent`) is the appropriate fallback for these.

I think I'll revert the change in the recommendation order until this is more mature (adding `byText(text, { role })` etc).

Overall I'd say we should definitely add some guard rails here when people try e.g. `getByRole('listitem', { name: 'first' })` by warning that certain roles don't have an accessible name and suggest using `byText('first', { role: 'listitem' })`. This is even a candidate for a lint rule with auto-fixing

docs/dom-testing-library/api-queries.md
eps1lon: Done

docs/guide-which-query.md
eps1lon: I changed the wording and order again. The rule of thumb is now:
- input? -> byLabelText (simple query that doesn't require much a11y knowledge. Personally I will still use `byRole('textbox', { name: '' })`. The problem is that you need to know all the various input roles but then maybe we should encourage knowing these?)
- interactive? -> byRole
- not-interactive? -> byText

alexkrolick: > The problem is that you need to know all the various input roles but then maybe we should encourage knowing these?

There a lot of aria roles and not necessarily one perfect role for a given use case. For example you might have a list that gets refactored to the new `feed` role later. I think the generic queries like `label` are good for use cases where you don't want to overspecify the exact roles.

eps1lon: > gets refactored to the new feed role later

That isn't just a refactor. You're actually changing semantics. We shouldn't be too overzealous with our testing approach. A test should actually fail some times.

~I'm also not sure how a `feed` relates to this issue~ I just realized why this causes so much confusion. ByLabelText is actually quite problematic with this API. I was under the impression it only considers actual `label` elements. The current implementation of ByLabelText is incomplete. We should discourage ByLabelText for anything other thant `input`.

alexkrolick: > The current implementation of ByLabelText is incomplete

What do you mean by incomplete? It also understands `aria-label` and `aria-labelledby`, which would require a significant amount of work to implement in userspace if you are wanting to traverse labelled elements.

PR #373, opened 2020-02-03
docs: Update Cypress Testing Library scoping example

NicholasBoll:
Recently https://github.com/testing-library/cypress-testing-library/pull/108 added a way to take the previous subject of a previous command to scope the query. Also `find*` queries handle `.should('not.exist')` and should be preferred over `query*` which require additional logic for eventual non-existence queries.

### DIFF
--- a/docs/cypress-testing-library/intro.md
+++ b/docs/cypress-testing-library/intro.md
@@ -27,8 +27,9 @@ and `queryAllBy` commands off the global `cy` object.
 [See the `DOM Testing Library` docs for reference](https://testing-library.com/docs/dom-testing-library/api-queries).
 
 > Note: the `get*` queries are not supported because for reasonable Cypress tests you
-> need retryability and `find*` queries already support that. `query*` queries are not
-> necessary, but supported.
+> need retryability and `find*` queries already support that. `query*` queries are no longer
+> necessary since v5 and will be removed in v6. `find*` fully support built-in Cypress
+> assertions (removes the only use-case for `query*`).
 
 ## With TypeScript
 
@@ -58,9 +59,6 @@ cy.findByLabelText('Label text', {timeout: 7000}).should('exist')
 
 // findAllByText _inside_ a form element
 cy.get('form').findByText('Button Text').should('exist')
-cy.get('form').within(() => {
-  cy.findByText('Button Text').should('exist')
-})
 cy.get('form').then(subject => {
   cy.findByText('Button Text', {container: subject}).should('exist')
 })
--- a/docs/cypress-testing-library/intro.md
+++ b/docs/cypress-testing-library/intro.md
@@ -27,9 +27,8 @@ and `queryAllBy` commands off the global `cy` object.
 [See the `DOM Testing Library` docs for reference](https://testing-library.com/docs/dom-testing-library/api-queries).
 
 > Note: the `get*` queries are not supported because for reasonable Cypress tests you
-> need retryability and `find*` queries already support that. The reason the `query*`
-> variants are supported is to allow you to assert that elements do _not_ appear on
-> the screen.
+> need retryability and `find*` queries already support that. `query*` queries are not
+> necessary, but supported.
 
 ## With TypeScript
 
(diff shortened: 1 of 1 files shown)


docs/cypress-testing-library/intro.md
```diff
@@ -27,9 +27,8 @@ and `queryAllBy` commands off the global `cy` object.
 [See the `DOM Testing Library` docs for reference](https://testing-library.com/docs/dom-testing-library/api-queries).
 
 > Note: the `get*` queries are not supported because for reasonable Cypress tests you
-> need retryability and `find*` queries already support that. The reason the `query*`
-> variants are supported is to allow you to assert that elements do _not_ appear on
-> the screen.
+> need retryability and `find*` queries already support that. `query*` queries are not
```
NicholasBoll: `query*` queries are no longer needed for asserting that elements do not appear in the DOM. A `find*` query can do that instead:

```js
// Doesn't exist now
cy.findByText('Non-existing Button Text').should('not.exist')

// Eventually won't exist
cy.findByText('Eventually will not exist').should('not.exist')
```

docs/cypress-testing-library/intro.md
```diff
@@ -53,9 +52,12 @@ To show some simple examples (from
 
 ```javascript
 cy.findAllByText('Jackie Chan').click()
-cy.queryByText('Button Text').should('exist')
-cy.queryByText('Non-existing Button Text').should('not.exist')
-cy.queryByLabelText('Label text', {timeout: 7000}).should('exist')
+cy.findByText('Button Text').should('exist')
+cy.findByText('Non-existing Button Text').should('not.exist')
+cy.findByLabelText('Label text', {timeout: 7000}).should('exist')
```
NicholasBoll: `find*` should be preferred over `query*`

docs/cypress-testing-library/intro.md
alexkrolick: I'd include a note about why they are "not necessary", including version when the find* change was added, and mention plans for future deprecation here.

kentcdodds: Agreed with @alexkrolick here 👍

docs/cypress-testing-library/intro.md
```diff
@@ -53,9 +52,12 @@ To show some simple examples (from
...
+
+// findAllByText _inside_ a form element
+cy.get('form').findByText('Button Text').should('exist')
 cy.get('form').within(() => {
```
kentcdodds: Maybe we should remove this example in favor of the regular chaining one?

docs/cypress-testing-library/intro.md
kentcdodds: Agreed. I'm glad that we're keeping all of the `query*` stuff off here considering we're going to be removing it. Thanks!

docs/cypress-testing-library/intro.md
NicholasBoll: I think `find` was changed in https://github.com/testing-library/cypress-testing-library/pull/78 and released as `5.0.0`

docs/cypress-testing-library/intro.md
NicholasBoll: I think so. I avoid `cy.within` for many of the same reasons I avoid `with` in JavaScript. At the top level and in the simplest case, it avoids duplicate code, but it can easily get out of hand

kentcdodds: Cool. Let's remove it then 👍

docs/cypress-testing-library/intro.md
NicholasBoll: How does the following sound?

https://github.com/testing-library/testing-library-docs/pull/373/files#diff-6245e2573aa625d927375e2065f25b1cR29-R32

> Note: the `get*` queries are not supported because for reasonable Cypress tests you
> need retryability and `find*` queries already support that. `query*` queries are no longer
> necessary since v5 and will be removed in v6. `find*` fully support built-in Cypress
> assertions (removes the only use-case for `query*`).

docs/cypress-testing-library/intro.md
```diff
@@ -53,12 +53,12 @@ To show some simple examples (from
...
-})
+cy.findByText('Button Text').should('exist')
+cy.findByText('Non-existing Button Text').should('not.exist')
+cy.findByLabelText('Label text', {timeout: 7000}).should('exist')
+
+// findAllByText _inside_ a form element
+cy.get('form').findByText('Button Text').should('exist')
 cy.get('form').then(subject => {
   cy.findByText('Button Text', {container: subject}).should('exist')
 })
```
NicholasBoll: What do people think about removing these lines? It is still supported, but the above example is much cleaner and more natural to other Cypress code.

kentcdodds:
Awesome. Thanks!

kentcdodds:
@all-contributors please add @NicholasBoll for docs

PR #374, opened 2020-02-05
Fixed axios module default import name to match mock

CatsOnFilm:
Why:

The example provides a declarative import name ("axiosMock") in
anticipation for it being mocked but doesn't actually use this name
in the mock.

* ...

This change addresses the need by:
- updates module name from axios -> axiosMock in `Full Example` snippet
- updates module name from axios -> axiosMock in `Step-By-Step/Imports`
  snippet


CatsOnFilm:
Yes, you're right. I thought it behaved differently and this wasn't clear from other examples (including Jest's docs), where the import is assigned the same name as the module name.

PR #385, opened 2020-02-29
docs(api-events): Add example of invalid change events

PaulACoroneos:
…ue (date example)

Edit @eps1lon:
Closes #385

### DIFF
--- a/docs/dom-testing-library/api-events.md
+++ b/docs/dom-testing-library/api-events.md
@@ -57,6 +57,16 @@ fireEvent.change(getByLabelText(/picture/i), {
     files: [new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' })],
   },
 })
+
+// note: When firing change events please be careful to use valid values.
+// For example if changing a target's value to a date you must abide by ISO 8601
+// standards. Otherwise the element will not update with the new value.
+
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-events.md
```diff
@@ -57,6 +57,16 @@ fireEvent.change(getByLabelText(/picture/i), {
     files: [new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' })],
   },
 })
+
+// note: When firing change events please be careful to use valid values.
+// For example if changing a target's value to a date you must abide by ISO 8601
```
eps1lon: This isn't 100% correct. It sounds like e.g. you're changing a value from a number to a date but the DOM doesn't make that inference. It will only check this if the input is already a date input. So I would change the wording here to make it clear that we have a date input (not just any input) and that we want to change its value.

afontcu:
Hi @PaulACoroneos! Sorry this took so long. Are you up to resolve Sebastian's suggestion so we can merge this up? 🙌

PaulACoroneos:
@afontcu Hello there! I don't know what happened but I can't update this PR anyone. I have made proposed changes and have re-opened the PR. Thank you!

PR #392, opened 2020-03-04
Fix arguments typo

dale-french:
Caught the typo in @kentcdodds live stream 😀


kentcdodds:
@all-contributors please add @dale-french for docs

PR #411, opened 2020-03-23
Updates clientScript instructions to use module

theinterned:
Module is preferable to absolute file path as it doesn't assume the location of `node_modules` folder and works better with monorepos using yarn workspaces etc. where dependencies may be hoiseted.

This `{ "module": "npm module" }` syntax is documented here https://devexpress.github.io/testcafe/documentation/using-testcafe/configuration-file.html#clientscripts — see the example of how `"loadsh"` is loaded.

### DIFF
--- a/docs/testcafe-testing-library/intro.md
+++ b/docs/testcafe-testing-library/intro.md
@@ -27,7 +27,7 @@ Add the following to your .testcaferc.json file:
 
 ```json
   "clientScripts": [
-    "module":"@testing-library/dom/dist/@testing-library/dom.umd.js"
+    { "module": "@testing-library/dom/dist/@testing-library/dom.umd.js" }
   ],
 ```
 
--- a/docs/testcafe-testing-library/intro.md
+++ b/docs/testcafe-testing-library/intro.md
@@ -27,7 +27,7 @@ Add the following to your .testcaferc.json file:
 
 ```json
   "clientScripts": [
-    "./node_modules/@testing-library/dom/dist/@testing-library/dom.umd.js"
+    "module":"@testing-library/dom/dist/@testing-library/dom.umd.js"
   ],
 ```
 


docs/testcafe-testing-library/intro.md
```diff
@@ -27,7 +27,7 @@ Add the following to your .testcaferc.json file:
 
 ```json
   "clientScripts": [
-    "./node_modules/@testing-library/dom/dist/@testing-library/dom.umd.js"
+    "module":"@testing-library/dom/dist/@testing-library/dom.umd.js"
```
benmonro: @alexkrolick is there docs on this syntax somewhere? I'm not familiar w/ it. also if this works, that doesn't look like valid json to me... is something off?

theinterned: oh right you are ... I edited this in the GitHub interface ... let me update

theinterned: @benmonro thanks for reviewing this. I updated the description with a link to the TestCafe docs for this syntax. https://devexpress.github.io/testcafe/documentation/using-testcafe/configuration-file.html#clientscripts

benmonro: oh very cool, I didn't realize they added that!  Actually let me hold off on merging this, I want to see if we can simplify it even further and just do 

`{module:"@testing-library/dom"}`

theinterned: I just tried running my repo with `{module:"@testing-library/dom"}` and it didn't work.

benmonro: dang. ok in that case i'll approve. :)

theinterned: This issue seems to be that the `"main"` entry in `@testing-library/dom` points to `"dist/index.js",` as opposed to the mud build — there is a `"umd:main"` entry, but I would guess TestCafe doesn't see that. https://github.com/testing-library/dom-testing-library/blob/master/package.json#L5-L7

I hacked the `main` in my node modules folder:

```diff
-  "main": "dist/index.js",
+  "main": "dist/@testing-library/dom.umd.js",
```

And the tests run with 

```json
{
  "clientScripts": [
    {
      "module": "@testing-library/dom"
    }
  ]
}
```

benmonro: yeah i filed an issue about this w/ testcafe way back: https://github.com/DevExpress/testcafe/issues/4176

looks like it's still open

PR #412, opened 2020-03-23
Update guide-disappearance to mention waitFor and change waitForElement to findBy

danieljcafonso:
Update to mention `waitFor` as it will replace `wait` and change `waitForElement` example to use `findBy` query instead.


kentcdodds:
@all-contributors please add @danieljcafonso for docs

PR #414, opened 2020-03-26
fix: add React import statement

NomanGul:
added React import statement because we are using JSX... otherwise, it was throwing an error:

image


kentcdodds:
@all-contributors please add @NomanGul docs

PR #416, opened 2020-03-28
Replace waitForElement to waitFor to remove warnings 

rafaelangical:
Replace waitForElement to WaitFor to break warnings on console and update example-intro on https://testing-library.com/docs/react-testing-library/example-intro 

And set correctly use guide to o use new release.


kentcdodds:
@all-contributors please add @rafaelangical for docs

PR #421, opened 2020-04-02
Uses && instead of ternary

nainardev:
Please do tests and tell me

### DIFF
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -122,12 +122,12 @@ export default function Fetch({ url }) {
   const buttonText = buttonClicked ? 'Ok' : 'Load Greeting'
 
   return (
-    <div>
+    <>
       <button onClick={fetchGreeting} disabled={buttonClicked}>
         {buttonText}
       </button>
-      {greeting ? <h1>{greeting}</h1> : null}
-    </div>
+      {greeting && <h1>{greeting}</h1>}
+    </>
   )
 }
 ```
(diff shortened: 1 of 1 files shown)


gnapse:
> Uses && instead of ternary

Why? What's wrong with ternaries?

docs/react-testing-library/example-intro.md
```diff
@@ -122,12 +122,12 @@ export default function Fetch({ url }) {
   const buttonText = buttonClicked ? 'Ok' : 'Load Greeting'
 
   return (
-    <div>
+    <>
```
eps1lon, line 125: Let's not use too much syntax here. There are more people familiar with `<div>` than `<>`. What was the purpose of this change?

docs/react-testing-library/example-intro.md
```diff
@@ -122,12 +122,12 @@ export default function Fetch({ url }) {
...
       <button onClick={fetchGreeting} disabled={buttonClicked}>
         {buttonText}
       </button>
-      {greeting ? <h1>{greeting}</h1> : null}
-    </div>
+      {greeting && <h1>{greeting}</h1>}
```
eps1lon, line 129: Similar reason here. Ternaries are perfectly fine.

nainardev:
Ok sorry for opening tis pr

kentcdodds:
No worries @nainardev! Thanks for taking the time :)

eps1lon:
> Ok sorry for opening tis pr

No worries. Every PR is appreciated. It helps including the reason for a change since it wasn't entirely obvious whether we got complacent with a certain pattern (here: conditional rendering) or if there's an actual issue.

docs/react-testing-library/example-intro.md
alexkrolick: I actually prefer ternaries so that you don't actually end up with, say `<div>0</div>` when the value is falsy but not actually `null` or `false`.

nainardev:
>Ok sorry for opening tis pr

Just kidding guys have a great day

PR #424, opened 2020-04-04
Update React Redux recipe to use custom render with wrapper

MatanBobi:
Updated the React Redux recipe to use a custom `render` with the `wrapper` option.
Also refactored the class component to a functional component.
This PR is following the talk with @kentcdodds in [this](https://github.com/reduxjs/redux/pull/3708) PR.

### DIFF
--- a/docs/example-react-redux.md
+++ b/docs/example-react-redux.md
@@ -81,6 +81,12 @@ function render(
   }
   return rtlRender(ui, { wrapper: Wrapper, ...renderOptions })
 }
+
+// re-export everything
+export * from '@testing-library/react'
+
+// override render method
+export { render }
 ```
 
 ```jsx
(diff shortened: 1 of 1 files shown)


docs/example-react-redux.md
```diff
@@ -23,7 +23,7 @@ class Counter extends React.Component {
         <h2>Counter</h2>
         <div>
           <button onClick={this.decrement}>-</button>
-          <span data-testid="count-value">{this.props.count}</span>
+          <h1>{this.props.count}</h1>
```
alexkrolick, line 26: h1 after h2? Let's leave this alone.

MatanBobi: My bad, missed it sorry.. Will fix.

MatanBobi: @alexkrolick Fixed it, thanks..
Also refactored the class component to a functional one in the current time spirit :)

docs/example-react-redux.md
```diff
@@ -58,58 +56,68 @@ export function reducer(state = initialState, action) {
...
+```
+
+```jsx
+// counter.test.js
+import React from 'react'
+import { createStore } from 'redux'
+import { Provider } from 'react-redux'
+import { fireEvent, screen } from '@testing-library/react'
+// We're using our own custom render function and not RTL's render
+import { render } from './test-utils.js
```
kentcdodds, line 93: In accordance with the setup suggestions, this should be:

```suggestion
// We're using our own custom render function and not RTL's render
// our custom utils also re-export everything from RTL
// so we can import fireEvent and screen here as well
import { render, fireEvent, screen } from './test-utils.js
```

MatanBobi: Updated it and also added the export to the `test-utils.js` file..
Thanks!

kentcdodds:
@all-contributors please add @MatanBobi for docs

PR #425, opened 2020-04-05
findBy / findAllBy default timeout

dymafr:
findBy / findAllBy default timeout is 1000ms and not 450ms

see: 
https://github.com/testing-library/dom-testing-library/blob/1b711a4a0143b532dc8258f9ceac1a1b36557ce0/src/config.js#L8
https://github.com/testing-library/dom-testing-library/blob/672f2316d49609572857e7de80251eb052c9cbd1/src/wait-for.js#L15

Thanks for your awesome work!


kentcdodds:
@all-contributors please add @dymafr for docs

PR #428, opened 2020-04-06
Update api-events.md

dymafr:
code are wrong, see:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code/code_values

charCode and keyCode are deprecated and should not be used any more:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode


kentcdodds:
@all-contributors please add @dymafr for docs

PR #432, opened 2020-04-07
explain how to update wait to waitFor

timrobinson33:
added information for anyone trying to migrate an existing parameterless wait call. Also there were 2 headings for wait() and I removed the second one because I thought it was superfluous


kentcdodds:
@all-contributors please add @timrobinson33 for docs

PR #435, opened 2020-04-08
fixed a minor typo

mdugue:
that can still cause a lot of confusion :)

it should be (only one dot)
```js 
"test-utils": ["./utils/test-utils"]
```

instead of 

```js 
"test-utils": ["../utils/test-utils"]
```


kentcdodds:
@all-contributors please add @mdugue for docs

PR #442, opened 2020-04-15
Replace waitForElement with waitFor in api-queries.md

mkermani144:


kentcdodds:
@all-contributors please add @mkermani144 for docs

PR #444, opened 2020-04-15
[docs] add await to example in userEvent.type for page example in testing-library.com/docs/ecosystem-user-event

adeelibr:
Made the change in the demo here https://testing-library.com/docs/ecosystem-user-event

```diff
- userEvent.type(getByRole('textbox'), 'Hello, World!')
+ await userEvent.type(getByRole('textbox'), 'Hello, World!')
```


kentcdodds:
@all-contributors please add @adeelibr for docs

PR #447, opened 2020-04-16
Fix wrapper link

varundey:


docs/react-testing-library/setup.md
```diff
@@ -26,7 +26,7 @@ all your imports. See [below](#configuring-jest-with-test-utils) for a way to
 make your test util file accessible without using relative paths.
 
 The example below sets up data providers using the
-[`wrapper`](api.md#render-options) option to `render`.
+[`wrapper`](api.md#wrapper) option to `render`.
```
kentcdodds, line 29: Thanks for this. But it looks like this is already done? 🤔

https://github.com/testing-library/testing-library-docs/blob/master/docs/react-testing-library/setup.md

That shows the correct link.

varundey: I'm confused. Do you mean that [render-options](https://github.com/testing-library/testing-library-docs/blob/master/docs/react-testing-library/api.md#render-options) should be the hyperlink rather than [wrapper](https://github.com/testing-library/testing-library-docs/blob/master/docs/react-testing-library/api.md#wrapper)?

kentcdodds: ooooooh! Sorry, I was confused. Yes, you're right. Thank you!

kentcdodds:
@all-contributors please add @varundey for docs

PR #453, opened 2020-05-04
Add leaning material link in portuguese

PabloDinella:
As suggested here :) https://twitter.com/kentcdodds/status/1236829600665063424


kentcdodds:
@all-contributors please add @PabloDinella for docs

PR #454, opened 2020-05-04
Update guide-which-query.md

kentcdodds:


alexkrolick:
Might be worth showing a small example with the name option inline since it is a primary use case

PR #456, opened 2020-05-05
doc: order testing library users alphabetically, handle pinned users

oriolpuig:
### Purpose
As I've proposed this morning [vía twitter](https://twitter.com/oriolpuigbuil/status/1257608327451807747) to @afontcu, I suggest to order all Testing Library users alphabetically to make it more readible.

**UPDATE**: I've applied the suggestion to only show 3 ranzom users at home page. Users page stills listing all users ordered alphabetically

### This is how it will look like

image


oriolpuig:
Thank you!

alexkrolick:
We should probably implement the "rotating" (randomized on load) pin approach so that we don't force users to download all these images to read the docs...

kentcdodds:
I agree. Most sites only show 3 or so and they rotate them. I'm a fan of that :)

oriolpuig:
I've updated the PR with more users and showing randomized list at home page. Users page stills listing all items ordered alphabetically.

website/pages/en/index.js
```diff
@@ -261,13 +261,33 @@ class Index extends React.Component {
...
+          taken[x] = --len in taken ? taken[len] : len
+        }
+        return result
+      }
+
+      const randomizedList = () => {
+        const NUMBER_OF_USERS_TO_SHOW = 3
+
+        return getRandom(
+          siteConfig.users.filter(user => user.pinned),
```
alexkrolick, line 281: Now that we have a randomizer, we should restore the meaning of "pinned" to mean "always on the homepage". Right now we don't want to pin anything, so they can all change to false.

oriolpuig: Cool, I'll do it and remove the filter.

oriolpuig:
I've updated the PR to apply last review comments. For now, all job is done. Let me know if you have mor feedback before merge this PR 🤘

alexkrolick:
Interesting... Docusaurus only executes JS at build time, so the users are fixed until a new deploy comes in and randomizes it differently. Are we OK with that?

Screenshot_2020-05-07 Testing Library · Simple and complete testing utilities that encourage good testing practices
Screenshot_2020-05-07 Testing Library · Simple and complete testing utilities that encourage good testing practices(1)

kentcdodds:
What do other doc sites do?

alexkrolick:
I changed it to show the pinned users and then add 3 random unpinned ones each deploy. Reasonable?

kentcdodds:
I think it's perfect 👍

kentcdodds:
@all-contributors please add @oriolpuig for docs

oriolpuig:
Thanks team! ❤️

PR #459, opened 2020-05-07
docs: add dataTransfer event property info

jamsinclair:
Docs for https://github.com/testing-library/dom-testing-library/pull/554

Please let me know if there is anywhere else that might be worth documenting this change. Also feel free to edit and improve.


kentcdodds:
@all-contributors please add @jamsinclair for docs

PR #462, opened 2020-05-11
Update cy examples to use find*

coryhouse:


kentcdodds:
@all-contributors please add @coryhouse for docs

PR #466, opened 2020-05-15
Add link to known aria roles

FredyC:


docs/guide-which-query.md
```diff
@@ -19,7 +19,7 @@ possible. With this in mind, we recommend this order of priority:
       top preference for just about everything. There's not much you can't get
       with this (if you can't, it's possible your UI is inaccessible). Most
       often, this will be used with the `name` option like so:
-      `getByRole('button', {name: /submit/i})`
+      `getByRole('button', {name: /submit/i})`. [Check the list of known roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles).
```
eps1lon, line 22:
```suggestion
      `getByRole('button', {name: /submit/i})`. Check the [list of roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles).
```
which is more exhaustive. The page you propose specifically lists only roles discussed on MDN (whatever that means):

> This page lists reference pages covering all the WAI-ARIA roles discussed on MDN.

"known roles" is a bit vague IMO.

FredyC: Makes sense. I found that link in @kentcdodds article :) https://kentcdodds.com/blog/common-mistakes-with-react-testing-library (in section Not using *ByRole most of the time).

eps1lon: Let's wait for Kent's approval. Maybe this was intentional. Otherwise we should get https://github.com/kentcdodds/kentcdodds.com/pull/462 in as well.

PR #472, opened 2020-05-23
Add example of firing select event

rkennel:
As a new user of this library I had a difficult time determining how to test outcomes of a drop-down selection.  There is an example if you use the react-select library but very little information on the internet if you use a plain select tag.

I found Kent's comment on an issue in the react-testing-library and thought it would be helpful to others to add it to the FAQs.  https://github.com/testing-library/react-testing-library/issues/322


kentcdodds:
@all-contributors please add @rkennel for docs

PR #477, opened 2020-05-26
docs: added showSuggestions

benmonro:


docs/dom-testing-library/api-configuration.md
```diff
@@ -24,6 +24,13 @@ for you to identify the part of your code that resulted in the error (async
 stack traces are hard to debug). If you want to disable this, then set
 `showOriginalStackTrace` to `false`. You can also disable this for a specific
 call in the options you pass to `waitFor`.
+`throwSuggestions`: (experimental) When enabled, if [better queries](https://testing-library.com/docs/guide-which-query) are available 
+the test will fail and provide a suggested query to use instead.  Default is 'false'
```
timdeschryver, line 28: To be consistent with the other config options

```suggestion
the test will fail and provide a suggested query to use instead.  Default to `false`.
```

PR #478, opened 2020-05-29
docs: fix default hidden option value for getByRole query

qkdreyer:
https://github.com/testing-library/testing-library-docs/blob/ea15db89abea14f4710ea6c7484884592a1b37a8/docs/dom-testing-library/api-configuration.md#L18-L19


eps1lon:
@qkdreyer Thanks

PR #479, opened 2020-05-29
Update intro.md

coryhouse:
As I outlined [here](https://github.com/testing-library/cypress-testing-library/issues/135), this example is incorrect. @pranjaljately pointed out the correct code [here](https://github.com/testing-library/cypress-testing-library/issues/135#issuecomment-636000449).


kentcdodds:
@all-contributors please add @coryhouse for docs

PR #483, opened 2020-06-01
docs: replace "axiosMock" with "msw" in React Testing Library example

### LINKED ISSUE #482: Consider showcasing "msw" package for API mocking
## What

I'd like to suggest to use [Mock Service Worker](https://github.com/mswjs/msw) in test examples that concern API mocking.

## Why

- MSW uses a declarative API to describe requests that should be mocked. 
- MSW captures all outgoing network communication on both client and server. It also runs in a fall-through mode, which means unrelated requests are left intact.
- MSW utilizes [Service Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) on the client-side usage, making the mocks completely detached from the application. Unlike most of the other solutions, that means a request interception on the _network_ level, not application level (i.e. making even the mocked requests/responses inspectable in the "Network" tab of your browser).
- MSW is not opinionated over how you construct your mock definitions. You can use plain request-response contracts, or bring in ORM and integrate it closer to your actual server's logic. The beauty if that you _don't have_ to do that.
- MSW can be used on both client and server, being agnostic of frameworks and request-issuing libraries (use React, Vue, `fetch` or `axios`—you're covered).

> Looks like @kentcdodds would [support recommending `msw`](https://twitter.com/kentcdodds/status/1266718046778880001) as a part of Testing Library ❤️ .

## What would it bring?

- Discourage people to stub `window.fetch` for the sake of API mocking. Instead, they would describe how an imaginary server would respond.
- Educate people on how Service Worker can be used for API mocking (the approach is relatively new and not explored).

## Where

Here's the "Example" page that showcases the usage of `axiosMock.get.mockResolvedValueOnce`:

- https://testing-library.com/docs/react-testing-library/example-intro

This example alone brings some of the following pain-points:

- It assumes that developer is using `axios`, making it a test's dependency.
- It stubs `axios.get` method to make it return what you need.
- Asserts of `axiosMock.get` was called, while it's not a part of the functionality we are testing.

Compare the page above with how it may look like with MSW:

```js
// __tests__/fetch.test.js
import React from 'react'
import { render, fireEvent, waitFor, screen } from '@testing-library/react'
import '@testing-library/jest-dom/extend-expect'
import { rest } from 'msw'
import { setupServer } from 'msw/node'
import Fetch from '../fetch'

const server = setupServer(
  rest.get('/greeting', (req, res, ctx) => {
    return res(ctx.json({ greeting: 'hello there'}))
  })
)

beforeAll(() => server.listen())
afterAll(() => server.close())

test('loads and displays greeting', async () => {
  const url = '/greeting'
  render(<Fetch url={url} />)

  fireEvent.click(screen.getByText('Load Greeting'))

  await waitFor(() => screen.getByRole('heading'))

  expect(screen.getByRole('heading')).toHaveTextContent('hello there')
  expect(screen.getByRole('button')).toHaveAttribute('disabled')
})
```

> Notice how test suite itself now knows nothing about the mocking. Because it shouldn't. We also remove the assertion that `axiosMock.get` was called, because it is not related to what we are actually testing.

## How

I would be glad to issue a pull request regarding these changes. Any guidance on that is also highly appreciated!

kettanaito:
## Changes

- Removes the usage of `jest.mock` and `axiosMock.get.mockResolvedValueOnce()` in favor of a more declarative approach (see the reasoning in the GitHub issue).
- Updates the "Step-by-step" section to briefly explain why `msw` has been added, and what each of its usage steps means.

## GitHub

- Closes #482 

## Notes

This was the only usage example I've found that illustrates testing a component that performs requests. Please, if I missed something, would you suggest me where to look for similar examples in the documentation? Thanks!

### DIFF
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -64,7 +64,22 @@ import { render, fireEvent, waitFor, screen } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 // the component to test
 import Fetch from '../fetch'
+```
 
+```jsx
+test('loads and displays greeting', async () => {
+  // Arrange
+  // Act
+  // Assert
+})
+```
+
+### Mock
+
+Use the `setupServer` function from `msw` to mock an API request that our tested
+component makes.
+
+```js
 // declare which API requests to mock
 const server = setupServer(
   // capture "GET /greeting" requests
@@ -78,14 +93,6 @@ const server = setupServer(
 // and clean up once the tests are finished
 beforeAll(() => server.listen())
 afterAll(() => server.close())
-```
-
-```jsx
-test('loads and displays greeting', async () => {
-  // Arrange
-  // Act
-  // Assert
-})
 ```
 
 ### Arrange
--- a/docs/react-testing-library/example-intro.md
+++ b/docs/react-testing-library/example-intro.md
@@ -11,27 +11,29 @@ See the following sections for a detailed breakdown of the test
 ```jsx
 // __tests__/fetch.test.js
 import React from 'react'
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
 import { render, fireEvent, waitFor, screen } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
-import axiosMock from 'axios'
 import Fetch from '../fetch'
 
-jest.mock('axios')
+const server = setupServer(
+  rest.get('/greeting', (req, res, ctx) => {
+    return res(ctx.json({ greeting: 'hello there' }))
+  })
+)
+
+beforeAll(() => server.listen())
+afterAll(() => server.close())
 
 test('loads and displays greeting', async () => {
   const url = '/greeting'
   render(<Fetch url={url} />)
 
-  axiosMock.get.mockResolvedValueOnce({
-    data: { greeting: 'hello there' },
-  })
-
   fireEvent.click(screen.getByText('Load Greeting'))
 
   await waitFor(() => screen.getByRole('heading'))
 
-  expect(axiosMock.get).toHaveBeenCalledTimes(1)
-  expect(axiosMock.get).toHaveBeenCalledWith(url)
   expect(screen.getByRole('heading')).toHaveTextContent('hello there')
   expect(screen.getByRole('button')).toHaveAttribute('disabled')
 })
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/example-intro.md
```diff
@@ -11,27 +11,29 @@ See the following sections for a detailed breakdown of the test
...
 
-jest.mock('axios')
+const server = setupServer(
+  rest.get('/greeting', (req, res, ctx) => {
+    return res(ctx.json({ greeting: 'hello there' }))
+  })
+)
+
+beforeAll(() => server.listen())
+afterAll(() => server.close())
```
kentcdodds, line 27: Should we add an `afterEach(() => server.resetHandlers())` here?

docs/react-testing-library/example-intro.md
```diff
@@ -133,8 +147,6 @@ export default function Fetch({ url }) {
 ```
 
 ```jsx
-expect(axiosMock.get).toHaveBeenCalledTimes(1)
-expect(axiosMock.get).toHaveBeenCalledWith(url)
 expect(screen.getByRole('heading')).toHaveTextContent('hello there')
 expect(screen.getByRole('button')).toHaveAttribute('disabled')
```
kentcdodds, line 151: Could you do me a favor and delete all the snapshot stuff at the bottom. Nobody should be doing that for stuff like this. <3

docs/react-testing-library/example-intro.md
kettanaito: As a safe option—yes, but unless you declare runtime handlers via `server.use()` there’s nothing to reset. I’d not suggest calling `resetHandlers` by default. Developers that need it will find it :)

docs/react-testing-library/example-intro.md
kettanaito: Sure. Will delete them tomorrow.

docs/react-testing-library/example-intro.md
kentcdodds: That's a fair point. Now that you mention it, I think the `server.use` call should be *inside* the test rather than outside it. That's what people should be doing most of the time.

With that in mind, resetHandlers should be included.

And I think it should be used by default 😉

kettanaito: May I please double check on this?

I see your point of using resetHandlers and server.use inside a test suite, but I think I'm confused of what runtime handlers to add via server.use, as the example tests exactly one API communication to "GET /greeting". 

```js
const server = setupServer(
  // centralized place for request handlers,
  // generally encouraged.
  rest.get('/greetings', resolver)
)

beforeAll(...)
afterAll(...)

test('loads and displays greeting', () => {
  // I need `afterEach(() => server.resetHandlers())`
  // only if I use `server.use()` in any of my tests.
  server.use(/* however, what should I add here? */)
})
```

Do you see what I mean?

kentcdodds: This is a fair point. There are two things to consider:

1. Handle all of our application's typical requests so we don't have to think about them on a per-test basis
2. Add a handler for a specific test (to test an error state)

With that in mind, I think we should adjust this example so we have two tests. One for the success case, and one for the error case. That way the first one can rely on the default handlers and one that can add a runtime handler for the error.

Because I see using `afterEach(() => server.resetHandlers())` as an important default whenever someone uses `server.use` and we should probably demonstrate that so people don't add a bunch of complicated single-test-specific handlers to their `setupServer` call.

kettanaito: _That_ is perfect.

kettanaito: I've added the error handling test scenario to this example.

1. Rewrote `useState` to `useReducer` in `fetch.js` to reflect the error and data relation.
1. Added a new test suite, used `server.use()` with permanent override there.
1. Added `afterEach` that calls `server.resetHanders()`.

Please, could you double check if that `fetch.js` looks okay to you? Thanks.

PR #499, opened 2020-06-12
docs: rework disappearance guide to provide more guidance

kimroen:
Makes a number of changes to the disappearance guide, moving it in a direction where it recommends
more of the best practices that have appeared since the last time it was uses, as well as taking a stronger
stance of which of multiple options are recommended.

This PR came about because I was answering some questions related to appearance and disappearance, and the people I was hadn't found the answers they were looking for in this guide.

In more detail, the changes are:

- Recommend `findBy` over `waitFor`
- Recommend `waitForElementToBeRemoved` over `waitFor`
- Use matchers from `jest-dom` in examples instead of checking for `null` and then mentioning `jest-dom` later
- Remove inline setup-instructions for `jest-dom` and instead link to docs
- Smaller code example changes (split up separate examples in the same code fence to multiple fences)
- Use the word `document` instead of `DOM` most places
- Use `screen` in all examples
- Add explicit `expect` statements in all the examples for consistency

### DIFF
--- a/docs/guide-disappearance.md
+++ b/docs/guide-disappearance.md
@@ -3,93 +3,136 @@ id: guide-disappearance
 title: Appearance and Disappearance
 ---
 
-Sometimes you need to test that an element is present and then disappears or
-vice versa.
+Sometimes you need to test that an element is present and then disappears or the
+other way around.
 
 ## Waiting for appearance
 
-If you need to wait for an element to appear, the [async wait
-utilities][async-api] allow you to wait for an assertion to be satisfied before
-proceeding. The wait utilities retry until the query passes or times out.
+If you need to wait for an element to appear, [`findBy`][find-by] and the [async
+wait utility][async-api] `waitFor` allow you to wait for an assertion to be
+satisfied before proceeding. The wait utilities retry until the query passes or
+times out.
 
-```jsx
+Wait for appearance and return the element using `findBy`:
+
+```javascript
 test('movie title appears', async () => {
   // element is initially not present...
 
-  // wait for appearance
-  await waitFor(() => {
-    expect(getByText('the lion king')).toBeInTheDocument()
-  })
-
-  // wait for appearance and return the element
-  const movie = await findByText('the lion king')
+  const movieTitle = await screen.findByText('the lion king')
+  expect(movieTitle).toBeInTheDocument()
 })
 ```
+
+Or wait for appearance using `waitFor` and `getBy`:
+
+```javascript
+test('movie title appears', async () => {
+  // element is initially not present...
+
+  await waitFor(() => {
+    expect(screen.getByText('the lion king')).toBeInTheDocument()
+  })
+})
+```
+
+These do more or less the same thing (`findBy` uses `waitFor` under the hood),
+but the `findBy` version results in simpler code and a better error message.
+
+_Note: `toBeInTheDocument()` comes from [`@testing-library/jest-dom`][jest-dom]_
 
 ## Waiting for disappearance
 
-The `waitForElementToBeRemoved` [async helper][async-api] function uses a
-callback to query for the element on each DOM mutation and resolves to `true`
-when the element is removed.
+When you want to wait until an element has disappeared, you can use
+`waitForElementToBeRemoved` or `waitFor`.
 
-```jsx
-test('movie title no longer present in DOM', async () => {
-  // element is removed
-  await waitForElementToBeRemoved(() => queryByText('the mummy'))
+The `waitForElementToBeRemoved` [async wait utility][async-api] takes an element
+as an argument and waits until that element has been removed from the document.
+
+```javascript
+test('movie title no longer present in document', async () => {
+  // element is present
+  const movie = screen.queryByText('the mummy')
+  await waitForElementToBeRemoved(movie)
+
+  // element has been removed
+  expect(movie).not.toBeInTheDocument()
 })
 ```
 
-Using
+`waitForElementToBeRemoved` works by using a
 [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
-is more efficient than polling the DOM at regular intervals with `waitFor`.
+internally, which means it responds to the element being removed in a very
+efficient way.
 
-The `waitFor` [async helper][async-api] function retries until the wrapped function
-stops throwing an error. This can be used to assert that an element disappears
-from the page.
+You can also pass an array or a callback to `waitForElementToBeRemoved` -
+[see the documentation for more options](dom-testing-library/api-async.md#waitforelementtoberemoved).
 
-```jsx
+Another option is using the `waitFor` helper.
+
+The `waitFor` [async wait utility][async-api] retries until the wrapped function
+stops throwing an error. Because expectations throw errors when they fail,
+putting one in the wrapped function can be used to wait for the element to be
+removed.
+
+```javascript
 test('movie title goes away', async () => {
   // element is initially present...
   // note use of queryBy instead of getBy to return null
   // instead of throwing in the query itself
   await waitFor(() => {
-    expect(queryByText('i, robot')).not.toBeInTheDocument()
+    expect(screen.queryByText('i, robot')).not.toBeInTheDocument()
   })
 })
 ```
 
+Polling like this is not as efficient as observing for mutations using
+`waitForElementToBeRemoved`, but sometimes it's the best option.
+
 ## Asserting elements are not present
 
-The standard `getBy` methods throw an error when they can't find an element, so
-if you want to make an assertion that an element is _not_ present in the DOM,
-you can use `queryBy` APIs instead:
+As opposed to waiting for removal, this is for when you are at a point in your
+test where you know an element shouldn't be present.
+
+You might reach for `getBy` to check that something is not present, but `getBy`
+queries throw an error when they can't find an element, which means you don't
+get the chance to use the result in an expectation. If you want to make an
+assertion that an element is _not_ present in the document, you can use
+`queryBy` queries instead:
 
 ```javascript
 const submitButton = screen.queryByText('submit')
-expect(submitButton).toBeNull() // it doesn't exist
+expect(submitButton).not.toBeInTheDocument() // it doesn't exist in the document
 ```
 
-The `queryAll` APIs version return an array of matching nodes. The length of the
-array can be useful for assertions after elements are added or removed from the
-DOM.
+The `queryAll` query methods return an array of matching elements. The length of
+the array can be useful for assertions after elements are added or removed from
+the document.
 
 ```javascript
 const submitButtons = screen.queryAllByText('submit')
 expect(submitButtons).toHaveLength(2) // expect 2 elements
 ```
 
-### `not.toBeInTheDocument`
+_Note: `not.toBeInTheDocument()` comes from
+[`@testing-library/jest-dom`][jest-dom]_
 
-The [`jest-dom`](ecosystem-jest-dom.md) utility library provides the
+## The `toBeInTheDocument()` matcher
+
+The [`@testing-library/jest-dom`][jest-dom] utility library provides the
 `.toBeInTheDocument()` matcher, which can be used to assert that an element is
 in the body of the document, or not. This can be more meaningful than asserting
-a query result is `null`.
+a query result is `null`, and it also provides more helpful error messages when
+your tests fail.
 
 ```javascript
-import '@testing-library/jest-dom/extend-expect'
 // use `queryBy` to avoid throwing an error with `getBy`
 const submitButton = screen.queryByText('submit')
 expect(submitButton).not.toBeInTheDocument()
 ```
 
+Read about how to set this up [in the documentation](jest-dom).
+
 [async-api]: dom-testing-library/api-async.md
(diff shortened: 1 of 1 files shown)


docs/guide-disappearance.md
```diff
@@ -3,93 +3,136 @@ id: guide-disappearance
...
   // instead of throwing in the query itself
   await waitFor(() => {
-    expect(queryByText('i, robot')).not.toBeInTheDocument()
+    expect(screen.queryByText('i, robot')).not.toBeInTheDocument()
   })
 })
 ```
 
+Polling like this is not as efficient as observing for mutations using
+`waitForElementToBeRemoved`, but sometimes it's the best option.
```
kimroen, line 90: Are there any good examples of when `waitFor` would work an `waitForElementToBeRemoved` wouldn't in this context? If not, then maybe it shouldn't have equal space on this page.

afontcu, line 90: Could we explain when it is the best option?

docs/guide-disappearance.md
```diff
@@ -3,93 +3,136 @@ id: guide-disappearance
...
-if you want to make an assertion that an element is _not_ present in the DOM,
-you can use `queryBy` APIs instead:
+As opposed to waiting for removal, this is for when you are at a point in your
+test where you know an element shouldn't be present.
+
+You might reach for `getBy` to check that something is not present, but `getBy`
+queries throw an error when they can't find an element, which means you don't
+get the chance to use the result in an expectation. If you want to make an
+assertion that an element is _not_ present in the document, you can use
+`queryBy` queries instead:
```
afontcu, line 101: I'd use more strong language here:

```suggestion
You might reach for `getBy` to check that something is not present, but `getBy`
queries throw an error when they can't find an element. If you want to make an
assertion that an element is _not_ present in the document, you need to use
`queryBy` queries instead:
```

docs/guide-disappearance.md
```diff
@@ -3,93 +3,136 @@ id: guide-disappearance
...
-
-  // wait for appearance and return the element
-  const movie = await findByText('the lion king')
 })
 ```
 
+These do more or less the same thing (`findBy` uses `waitFor` under the hood),
+but the `findBy` version results in simpler code and a better error message.
+
+_Note: `toBeInTheDocument()` comes from [`@testing-library/jest-dom`][jest-dom]_
```
afontcu, line 42: Neat!

docs/guide-disappearance.md
afontcu: agree 👍 

If not, I wouldn't even mention a "negated" `waitFor` as an alternative to `waitForElementToBeRemoved`

docs/guide-disappearance.md
kimroen: I'm not sure I understand how removing that clause results in stronger language, but I'm fine with the change.

afontcu: Oh, the change was from "you can use" to "you need to". Then I noticed the clause was kinda redundant so I merged both changes. Sorry if that wasn't clear!

docs/guide-disappearance.md
```diff
@@ -3,93 +3,136 @@ id: guide-disappearance
...
+your tests fail.
 
 ```javascript
-import '@testing-library/jest-dom/extend-expect'
 // use `queryBy` to avoid throwing an error with `getBy`
 const submitButton = screen.queryByText('submit')
 expect(submitButton).not.toBeInTheDocument()
 ```
 
+Read about how to set this up [in the documentation](jest-dom).
```
kimroen, line 134: Oops, used the wrong kind of markdown link here.

```suggestion
Read about how to set this up [in the jest-dom documentation][jest-dom].
```

docs/guide-disappearance.md
kimroen: Ah, I didn't even notice that change from "can" to "need". Makes more sense now!

docs/guide-disappearance.md
```diff
@@ -3,93 +3,135 @@ id: guide-disappearance
...
+wait utility][async-api] `waitFor` allow you to wait for an assertion to be
+satisfied before proceeding. The wait utilities retry until the query passes or
+times out.
 
-```jsx
+Wait for appearance and return the element using `findBy`:
+
+```javascript
+test('movie title appears', async () => {
+  // element is initially not present...
```
MichaelDeBoey, line 20: Maybe add an extra assertion to check that it's not present initially?
```js
// element is initially not present...
expect(movieTitle).not.toBeInTheDocument()
```

docs/guide-disappearance.md
```diff
@@ -3,93 +3,135 @@ id: guide-disappearance
...
+  const movieTitle = await screen.findByText('the lion king')
+  expect(movieTitle).toBeInTheDocument()
+})
+```
+
+Or wait for appearance using `waitFor` and `getBy`:
+
+```javascript
 test('movie title appears', async () => {
   // element is initially not present...
```
MichaelDeBoey, line 31: Same

docs/guide-disappearance.md
```diff
@@ -3,93 +3,135 @@ id: guide-disappearance
...
+When you want to wait until an element has disappeared, you can use
+`waitForElementToBeRemoved` or `waitFor`.
+
+The `waitForElementToBeRemoved` [async wait utility][async-api] takes an element
+as an argument and waits until that element has been removed from the document.
+
+```javascript
+test('movie title no longer present in document', async () => {
+  // element is present
+  const movie = screen.queryByText('the mummy')
```
MichaelDeBoey, line 55: Maybe add an extra assertion to check that it is present initially?

docs/guide-disappearance.md
```diff
@@ -3,93 +3,135 @@ id: guide-disappearance
...
-from the page.
+The `waitFor` [async wait utility][async-api] retries until the wrapped function
+stops throwing an error. Because expectations throw errors when they fail,
+putting one in the wrapped function can be used to wait for the element to be
+removed.
 
-```jsx
+```javascript
 test('movie title goes away', async () => {
   // element is initially present...
```
MichaelDeBoey, line 80: Same

docs/guide-disappearance.md
kimroen: Maybe, but I feel that isn't entirely relevant and detracts a little from what we're trying to show in this example. Also, asserting that things don't exist is covered further down on the page.

If we did do that though, there should also be code here to do something that would make the title appear after asserting it isn't here, and I'm not sure it makes sense to get into that here.

We could add a comment saying as much after the assertion, like this:

```suggestion
  expect(screen.queryByText('the lion king')).not.toBeInTheDocument()
  
  // pretend we do something here that should make the title appear
```

But I think it might be better to alter the comment in the test to something that makes it clearer that this would be part of a longer test:

```suggestion
  // ...code setting up a page that will eventually contain a movie title

```

Do you think the last suggestion would help, or is there maybe something else we could do to alleviate the confusion or lack of clarity we are hoping to fix by adding this assertion?

MichaelDeBoey: I think the first suggestion is much clearer instead of changing the comment tbh

MichaelDeBoey: On the other hand, you're right with not wanting to have things in here that you'll explain later, so maybe the changed comment is indeed a better option to go with here...

Maybe we can make a full example (like the second suggestion) later on the page? 🤔

kimroen: >I think the first suggestion is much clearer instead of changing the comment tbh

Fair! In your mind, what is it that is unclear that adding this assertion solves? Or is there something else that makes it so you want it to be added?

MichaelDeBoey: On the other hand, you're right with not wanting to have things in here that you'll explain later, so maybe the changed comment is indeed a better option to go with here...

Maybe we can make a full example (like the second suggestion) later on the page? 🤔

kimroen:
I've been trying to think of a good example where we would want to assert that something is present, disappears and is not present, and I'm struggling. I think this is change is valuable without an extra example though, so maybe a good way forward could be to merge it and then work on an example separately if we feel that would add clarity?

@MichaelDeBoey Do you have a specific example we could add, and would you be interested in adding that in a separate PR?

@afontcu I think I have addressed all your comments - are you okay with merging this as it is?

kimroen:
Actually, I'll go in and remove the `waitFor` variant for waiting for elements to be removed.

docs/guide-disappearance.md
kimroen: I agree, I've removed the `waitFor` option here ✨

docs/guide-disappearance.md
```diff
@@ -3,93 +3,114 @@ id: guide-disappearance
...
+    expect(screen.getByText('the lion king')).toBeInTheDocument()
   })
-
-  // wait for appearance and return the element
-  const movie = await findByText('the lion king')
 })
 ```
 
+These do more or less the same thing (`findBy` uses `waitFor` under the hood),
+but the `findBy` version results in simpler code and a better error message.
```
afontcu, line 40: Okay, this is interesting.

I'm not sure I agree on the statement (I believe using `waitFor` is more explicit, and helps the test conveying meaning) but that's entirely personal. The question here is whether if docs should showcase two ways of doing the same thing 🤔 

If we agree on `findBy` being simpler and packed with a better error message, then I'd say this is the way docs should suggest 👍 

`waitFor` is useful for other stuff (such as waiting for XHR requests to happen, or other). It might appear somewhere else in docs.

kimroen:
Some relevant comments by @kentcdodds over here: https://github.com/testing-library/react-testing-library/issues/673#issuecomment-643706909

docs/guide-disappearance.md
```diff
@@ -3,93 +3,114 @@ id: guide-disappearance
...
+`waitForElementToBeRemoved`.
+
+The `waitForElementToBeRemoved` [async wait utility][async-api] takes an element
+as an argument and waits until that element has been removed from the document.
+
+```javascript
+test('movie title no longer present in document', async () => {
+  // element is present
+  const movie = screen.queryByText('the mummy')
+  await waitForElementToBeRemoved(movie)
```
kimroen, line 56: I see @kentcdodds commented over here that this should receive a callback, not the element directly: https://github.com/testing-library/react-testing-library/issues/673#issuecomment-643706909

I feel like I've seen that doing it this way works too though, so I'm not sure.

afontcu: It can either receive a callback, an element, or an array of elements:

https://github.com/testing-library/dom-testing-library/blob/master/types/__tests__/type-tests.ts#L158-L164

docs/guide-disappearance.md
afontcu: I think this is the last bit that needs some attention, then we can merge!

docs/guide-disappearance.md
alexkrolick: I think there's value in keeping the first examples minimal, without having to decipher fake application code. Expanding on the comment and making clear this is a snippet of a larger test should be good.

docs/guide-disappearance.md
alexkrolick, line 40:
```suggestion
These do more or less the same thing (`findBy` uses `waitFor` under the hood),
but the `findBy` version produces errors more like a `getBy` query,
instead of a general-purpose timeout error.
```

docs/guide-disappearance.md
alexkrolick: might be worth expanding on this section later with a note about working with React, because sometimes the node gets destroyed and replaced, making these types of tests either pass or fail when they shouldn't

docs/guide-disappearance.md
```diff
@@ -3,93 +3,114 @@ id: guide-disappearance
...
-test('movie title no longer present in DOM', async () => {
-  // element is removed
-  await waitForElementToBeRemoved(() => queryByText('the mummy'))
+  // element has been removed
+  expect(movie).not.toBeInTheDocument()
 })
 ```
 
-Using
+`waitForElementToBeRemoved` works by using a
```
alexkrolick, line 63: I think the `waitFor` style should also be documented here, particularly for working with React, when the node reference itself is unstable but a retryable query inside a callback would work.

docs/guide-disappearance.md
```diff
@@ -3,93 +3,114 @@ id: guide-disappearance
...
+You can also pass an array or a callback to `waitForElementToBeRemoved` -
+[see the documentation for more options](dom-testing-library/api-async.md#waitforelementtoberemoved).
 
 ## Asserting elements are not present
 
-The standard `getBy` methods throw an error when they can't find an element, so
-if you want to make an assertion that an element is _not_ present in the DOM,
-you can use `queryBy` APIs instead:
+As opposed to waiting for removal, this is for when you are at a point in your
+test where you know an element shouldn't be present.
```
alexkrolick, line 74: This section can be tightened up. It's unclear what "this" refers to here. Also, we shouldn't assume jest-dom is installed and should mention the `null` check.

```suggestion
If you want to assert that an element is not present _right now_, 
rather than testing that it is present and will disappear, use
`queryBy` instead of `getBy` for the query, and then assert that
nothing was found. 
```

alexkrolick:
There are a few open items on this, but it's almost there. Adding some labels for visibility. Hoping someone can finish it off!

PR #502, opened 2020-06-16
Elevate fire event page in prep for merge

JacobParis:
As per conversation on #490 (and #485), the event firing utilities are largely universal and independent of the testing library family products they're imported into

This PR moves [the existing Fire Event page](https://testing-library.com/docs/dom-testing-library/api-events) from its nested folder in DOM Testing Library to the top level in preparation for it to be merged with the [docs from User Event](https://github.com/testing-library/user-event)

@MatanBobi had begun working on #485 so I followed his model for the page structure to keep things consistent

This PR includes a 301 Redirect which, assuming I implemented it correctly, will cover all traffic going to the original Fire Event page and make this a non-breaking change

Screen Shot 2020-06-15 at 11 57 28 PM

I'm interested in any feedback or pushback with the goal of being able to migrate the [docs from User Event](https://github.com/testing-library/user-event) to this repository in short order

### DIFF
--- a/docs/angular-testing-library/api.md
+++ b/docs/angular-testing-library/api.md
@@ -288,9 +288,7 @@ are automatically bound to the Angular Component. This to ensure that the
 Angular change detection has been run by invoking `detectChanges` after an event
 has been fired.
 
-See
-[Firing Events](https://testing-library.com/docs/dom-testing-library/api-events)
-for a complete list.
+See [Firing Events](universal-api-events.md) for a complete list.
 
 **example**:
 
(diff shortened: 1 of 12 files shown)


MatanBobi:
Hi @JacobParis !
The work I did in this issue is still a WIP, I will issue a PR once done, but as you can see in the comments there, the structure will probably change and the Queries page won't be under "Getting Started"..

JacobParis:
Closing since the merge isn't happening now and when it does this will be horribly out of date

nickmccurdy:
This still seems like a useful change to me, even without merging user event. Perhaps we could reopen it and use similar formatting for #519.

nickmccurdy:
Seems like this was completed elsewhere.

PR #507, opened 2020-06-20
docs: Link to new Discord

nickmccurdy:


docs/contributing.md
```diff
@@ -32,7 +32,7 @@ For questions related to using the library, please visit a support community
 instead of filing an issue on GitHub.
 
 - [Spectrum][spectrum]
```
kentcdodds, line 34: Let's delete this

website/core/Footer.js
```diff
@@ -55,8 +55,8 @@ class Footer extends React.Component {
             <ExternalLink href="https://stackoverflow.com/questions/tagged/react-testing-library">
               Stack Overflow
             </ExternalLink>
-            <ExternalLink href="https://www.reactiflux.com/">
-              Reactiflux on Discord
+            <ExternalLink href="https://discord.gg/c6JN9fM">
+              Discord
             </ExternalLink>
             <ExternalLink href="https://spectrum.chat/testing-library">
               Spectrum
```
kentcdodds, line 62: Let's delete spectrum everywhere

website/pages/en/support.js
```diff
@@ -28,7 +28,7 @@ function Support(props) {
       title: 'Spectrum',
```
kentcdodds, line 28: Here too

nickmccurdy:
I've removed most of Spectrum links, but I'm running into a layout issue involving the support links in Docusaurus, which I'm not very experienced with. When removing a link, the two rows no longer have three items each, and I get this unnatural layout where the second row is different.
image


One thing I've noticed is that we're giving more than three items to `<GridBlock layout="threeColumn" align="left" />`, which is only supposed to take three. However, even if I split this into two `GridBlock`s, I can't seem to get the second row to be left aligned.

kentcdodds:
I'm not sure 😬

nickmccurdy:
@alexkrolick Would you mind taking a look at this issue if you have time?

kentcdodds:
Meh, let's just merge it. We can fix it later.

alexkrolick:
@nickmccurdy sorry I missed this, yeah, I think we got lucky with the layout aligning in the right number of columns. I recall trying a few different combos (like center-alignining different things, and different column counts) but the layout never really looked that great. Feel free to change it up.

PR #508, opened 2020-06-20
Merge support page into help page

nickmccurdy:
The content of the [help](https://testing-library.com/help) and [support](https://testing-library.com/support) pages is very similar, and it's not very clear what the distinction is from the navbar as support usually refers to help for users. I think it would be easier to browse and maintain this information if we put it all in one place in the help page.


nickmccurdy:
I'd like to give @kentcdodds a chance to review before merging, since I'm changing the copy about Testing JavaScript and his other courses/workshop

PR #513, opened 2020-06-27
Correct Assert section in Example page of React Testing Library

gyuwon:
Assert section in Example page of React Testing Library does not
contains assertion code but SUT. So rename the section to
'System Under Test' and insert a new section that describes assertion
code correctly.


kentcdodds:
@all-contributors please add @gyuwon for docs

PR #516, opened 2020-07-01
Add React Hooks broke my tests, now what? EN and PT talk

danieljcafonso:
- Adds both Portuguese and English version of my React Hooks broke my tests, now what? talk
	- it approaches a migration from enzyme to react testing library and does a dive into the fundamentals to use the React Testing Library


nickmccurdy:
Thanks, I think this is a very helpful and informative talk

PR #518, opened 2020-07-03
Add documentation for queryFallback

nickmccurdy:
It seems like #386 hasn't been fixed yet, so I added some information about `queryFallbacks` based on the `dom-testing-library` type and documentation text suggested in the issue. Feel free to suggest better wording/organization.

**Preview:** Scroll to the bottom of this section https://deploy-preview-518--testing-library.netlify.app/docs/dom-testing-library/api-queries#byrole

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -605,6 +605,7 @@ getByRole(
     name?: TextMatch,
     normalizer?: NormalizerFn,
     selected?: boolean,
+    queryFallbacks?: boolean,
   }): HTMLElement
 ```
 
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-queries.md
```diff
@@ -710,6 +711,10 @@ cy.findByRole('dialog').should('exist')
 
 <!--END_DOCUSAURUS_CODE_TABS-->
 
+By default, it's assumed that the first role of each element is supported, so only the first role can be queried. If you need to query an element by any role in its fallbacks instead, you can use `queryFallbacks: true`.
```
eps1lon, line 714: Does 

```suggestion
By default, it's assumed that the first role of each element is supported, so only the first role can be queried. If you need to query an element by any of its fallback roles instead, you can use `queryFallbacks: true`.
```

PR #526, opened 2020-07-18
Add a recommendation on how to import `act`

### LINKED ISSUE #524: Clarify preference between `act` from react-testing-library and react-dom
**Is your feature request related to a problem? Please describe.**

Currently there are two almost identical ways of importing `act` helper:

```js
import {act} from 'react-dom/test-utils';
import {act} from '@testing-library/react';
```

The only difference is that built-in `act` is not available in react < 16.8. If I am using newer react, which one would I choose?

There are two possible answers:

1. In future, react-testing-library will drop support for older react versions and deprecate `act` re-export. In this case, it is better to stick with the import from react-dom.
2. Alternatively, react-testing-library could take a path of extending `act` with extra functionality (for example, extra checks for `waitFor` handlers, flushing any other timers). In this case, it is better to stick with the import from the testing-library.

Could you clarify, which path is more likely to take place?

just-boris:
closes #524

### DIFF
--- a/docs/react-testing-library/api.md
+++ b/docs/react-testing-library/api.md
@@ -290,4 +290,4 @@ errors in your tests).
 This is a light wrapper around the
 [`react-dom/test-utils` `act` function](https://reactjs.org/docs/test-utils.html#act).
 All it does is forward all arguments to the act function if your version of
-react supports `act`.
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks!

@all-contributors please add @just-boris for docs

PR #528, opened 2020-07-21
docs(fakeTimers): Explain how to use fake timers in testing-library

### LINKED ISSUE #523: 10.4.4 causing timeout issue with useEffect+jest.useFakeTimers
- `@testing-library/react` version: 10.4.4
- Testing Framework and version: Jest 24.9.0 (Create React App)
- DOM Environment: jsdom 14.1.0 (Create React App)

### Relevant code or config:
```js
// In a component
useEffect(() => {
  // any effect
}, [])

// In the tests
beforeEach(jest.useFakeTimers)
afterEach(jest.useRealTimers)
```

### What you did:

I have a component that has a `useEffect` in it and its tests use `jest.useFakeTimers`. My tests for the component passed until I updated to `@testing-library/react` 10.4.4.

### What happened:

The test now fails with this message:
```
Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

      at mapper (node_modules/jest-jasmine2/build/queueRunner.js:25:45)
```

### Reproduction:

See https://github.com/awesomeunleashed/rtl-10-4-4-repro

### Problem description:

I believe these tests should still be passing.

### Suggested solution:

I imagine this has something to do with the scheduler changes in testing-library/react-testing-library#732, as `useEffect` and `jest.useFakeTimers` sound like scheduling-related things, but I don't see anything obviously related without knowing the code.

MatanBobi:
Following the conversation in https://github.com/testing-library/react-testing-library/pull/743#discussion_r454001823 with @kentcdodds and @eps1lon, 
I've created this page to explain the work with fake timers in testing-library.
I wasn't 100% sure where to put this section so for now I put it under the `Getting Started` section since this is relevant knowledge for all testing library frameworks (I really think we should build a `General` section).
This PR closes #523

### DIFF
--- a/docs/using-fake-timers.md
+++ b/docs/using-fake-timers.md
@@ -5,8 +5,8 @@ sidebar_label: Using Fake Timers
 ---
 
 Using real timers in your tests is less common since they depend on real time
-lapse. For that, some testing frameworks offer the option to use fake timers in
-your tests so you won't need to depend on real times.
+lapse. For that, most testing frameworks offer the option to use fake timers in
+your tests so you won't need to rely on real times.
 
 When using fake timers in your tests, all of the code inside your test uses fake
 timers. The common pattern to setup fake timers is usually within the
@@ -20,14 +20,15 @@ beforeEach(() => {
 
 When doing so, you'll probably want to restore the timers after your test runs.
 For that you usually call `useRealTimers` in `afterEach`. It's important to
-remember that before calling `useRealTimers` you have to `clearAllTimers`. This
-will ensure you clear all the timers even if they weren't executed. That way,
-your fake timers are encapsulated to your tests only and when we try to cleanup,
-we will work with real timers. So you'll need to do something like this:
+remember that before calling `useRealTimers` you should call
+`runOnlyPendingTimers`. This will ensure you progress all pending timers so non
+of them are left hanging and don't get executed.  
+That way,your fake timers are encapsulated to your tests only and all of them
+behave as you would expect. Here's an example of doing that in jest:
 
 ```js
 afterEach(() => {
-  jest.clearAllTimers()
+  jest.runOnlyPendingTimers()
   jest.useRealTimers()
 })
 ```
--- a/docs/using-fake-timers.md
+++ b/docs/using-fake-timers.md
@@ -4,14 +4,14 @@ title: Using Fake Timers
 sidebar_label: Using Fake Timers
 ---
 
-Using real timers in your tests is problematic since they depend on real time
-lapse.  
-When you depend on real time, your tests can be unpredictable, slow, flaky. This
-will will also prevent you from making any assumptions about timestamps within
-your tests.
+In some cases, when your code uses timers (`setTimeout`, `setInterval`,
+`clearTimeout`, `clearInterval`), your tests may become unpredictable, slow and
+flaky.
 
-To solve these problems, most testing frameworks offer the option to use fake
-timers in your tests so you won't need to rely on real times.
+To solve these problems, or if you need to rely on specific timestamps in your
+tests, most testing frameworks offer the option to replace the real timers in
+your tests with fake ones. This should be used sporadically and not on a regular
+basis since using it contains some overhead.
 
 When using fake timers in your tests, all of the code inside your test uses fake
 timers.  
--- a/docs/using-fake-timers.md
+++ b/docs/using-fake-timers.md
@@ -1,0 +1,33 @@
+---
+id: using-fake-timers
+title: Using Fake Timers
+sidebar_label: Using Fake Timers
+---
+
+Using real timers in your tests is less common since they depend on real time
+lapse. For that, some testing frameworks offer the option to use fake timers in
+your tests so you won't need to depend on real times.
+
+When using fake timers in your tests, all of the code inside your test uses fake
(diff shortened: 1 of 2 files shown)


docs/using-fake-timers.md
```diff
@@ -0,0 +1,34 @@
...
+beforeEach(() => {
+  jest.useFakeTimers()
+})
+```
+
+When doing so, you'll probably want to restore the timers after your test runs.
+For that you usually call `useRealTimers` in `afterEach`. It's important to
+remember that before calling `useRealTimers` you should call
+`runOnlyPendingTimers`. This will ensure you progress all pending timers so none
+of them are left hanging and don't get executed.
```
eps1lon, line 25: This does ensure they get execute. The point was to flush scheduled timers. If we just switch to real timers then the scheduled tasks wouldn't get executed.

docs/using-fake-timers.md
```diff
@@ -0,0 +1,34 @@
...
+timers. The common pattern to setup fake timers is usually within the
+`beforeEach`, here's an example of how to do that using jest:
+
+```js
+beforeEach(() => {
+  jest.useFakeTimers()
+})
+```
+
+When doing so, you'll probably want to restore the timers after your test runs.
```
eps1lon, line 21: We probably want to explain why e.g. cleanup tasks from 3rd party libraries usually rely on real timers.

docs/using-fake-timers.md
```diff
@@ -0,0 +1,34 @@
+---
+id: using-fake-timers
+title: Using Fake Timers
+sidebar_label: Using Fake Timers
+---
+
+Using real timers in your tests is less common since they depend on real time
+lapse. To solve this, most testing frameworks offer the option to use fake
+timers in your tests so you won't need to rely on real times.
```
eps1lon, line 9: I think it's better to explain why real timers are bad for tests:
- flaky tests
- poor predictability
- can't make assumptions about specific timestamps.

MatanBobi: I wasn't sure we want to explain the rational to using fake timers since this page is for people already using it, 
that's why my explanation wasn't thorough.. I'll expand it :)

docs/using-fake-timers.md
MatanBobi: Sure, I'll add this

docs/using-fake-timers.md
MatanBobi: I'll explain it better, thanks! :)

docs/using-fake-timers.md
eps1lon: Hmm that works as well. I wouldn't make any assumptions about how common one or the other is though.

kentcdodds: I agree. If I were to guess, real timers are far more common (I personally use them only very rarely) mostly because real timers are the default.

docs/using-fake-timers.md
MatanBobi: Updated, thanks.

docs/using-fake-timers.md
MatanBobi: Also fixed this :) thanks

docs/using-fake-timers.md
MatanBobi: Updated and expanded the explanation on why to use fake timers :)
Thanks!

MatanBobi:
Well, I'm still not 100% happy with the phrasing there but I've updated it based on the comments,
Thanks! :)

docs/using-fake-timers.md
```diff
@@ -4,29 +4,44 @@ title: Using Fake Timers
 sidebar_label: Using Fake Timers
 ---
 
-Using real timers in your tests is less common since they depend on real time
-lapse. To solve this, most testing frameworks offer the option to use fake
+Using real timers in your tests is problematic since they depend on real time
+lapse.  
+When you depend on real time, your tests can be unpredictable, slow, flaky. This
+will will also prevent you from making any assumptions about timestamps within
```
eps1lon, line 10:
```suggestion
will also prevent you from making any assumptions about timestamps within
```

docs/using-fake-timers.md
```diff
@@ -0,0 +1,49 @@
+---
+id: using-fake-timers
+title: Using Fake Timers
+sidebar_label: Using Fake Timers
+---
+
+Using real timers in your tests is problematic since they depend on real time
```
kentcdodds, line 7: I feel like the wording here strongly advises the use of fake timers 100% of the time or as the default. I don't know about anyone else, but fake timers as complexity to test code and only sometimes provide any benefit and I therefore only use them when I really want to fake timers (like a debounce). Either way, I don't think this mentions the drawbacks fairly and I think we should do so. I don't think people should come away from reading this with the ideas that fake timers should be their default way for testing. Maybe we could add a section for when fake timers are worth the trouble?

MatanBobi: Yeah you're right, it's pretty misleading, I'll work on that.

MatanBobi: Sorry it took me some time, I rephrased it. 
I think it's still not 100% explanatory though, wasn't really sure how I can expand it..

PR #534, opened 2020-07-27
[findByLabelText] Adds a bad example using for w/ a non-form element

sharmilajesupaul:
This came up at my work today.

An engineer thought that a `for` attr on a `<label>` along with a related `<section>` `id` can be queried by `findByLabelText`. 

But really, this is not semantically correct and we end up with this error: 
```
Timed out retrying: Found a label with the text of: Photos, however no form control was found associated to that label. Make sure you're using the "for" attribute or "aria-labelledby" attribute correctly.
```

I think making that explicit in the docs would avoid future confusion.


kentcdodds:
@all-contributors please add @sharmilajesupaul for docs

PR #542, opened 2020-08-02
Add typescript section to svelte docs to clarify usage with Typescript

BatuhanW:
Typescript isn't supported by current documentation. With a simple step, we can enable Typescript support.

https://testing-library.com/docs/svelte-testing-library/setup


kentcdodds:
@all-contributors please add @BatuhanW for docs

PR #543, opened 2020-08-02
docs(svelte): add reference to typescript

camjackson:
I currently have [this PR open](https://github.com/mihar-22/svelte-jester/pull/10) in `svelte-jester` to make it a bit clearer how to get svelte-testing-library to play nice with TypeScript.

Assuming that that gets merged (or something like it), I thought it would be good to add a reference to TypeScript here in the testing-library docs, just so the breadcrumbs are a little easier to follow for a newbie setting up svelte-testing-library who is also using TypeScript. That's what this PR does 🙂

### DIFF
--- a/docs/svelte-testing-library/setup.md
+++ b/docs/svelte-testing-library/setup.md
@@ -74,20 +74,10 @@ with any testing framework and runner you're comfortable with.
 
 ## Typescript
 
-In order to support Typescript, you need to enable `preprocess`.
-
-1. Update your jest configuration
-
-  ```json
-  {
-    "jest": {
-      "transform": {
-        "^.+\\.svelte$": ["svelte-jester", { "preprocess": true }]
-      },
-      "moduleFileExtensions": ["js", "ts", "svelte"]
-    }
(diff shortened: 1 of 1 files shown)


kentcdodds:
Ah, could you resolve the conflicts here please?

camjackson:
Oh, um, the merge conflict was actually with someone else who had the same idea. #542 added added inline instructions for how to get typescript to work, but they're a bit incomplete based on the discussion we had over in the svelte-jester issues.

So to resolve the conflict what I've done is remove the specific instructions that were added here in #542, and replaced them with just a reference to go and read the svelte-jester docs, seeing as those are more complete and will hopefully stay up to date with any changes to svelte jester.

I feel a little bit bad for deleting work that someone else just did, but I feel like this is a better solution. Happy to discuss 🙂

mihar-22:
Looks good and as discussed I think this is much better. Thanks for all your help again :)

mihar-22:
@all-contributors please add @camjackson for docs

PR #545, opened 2020-08-02
docs(example-reach-router): add example of route with parameters

deadkff01:
I just add an example of how to test route components with parameters with `@reach/router`.

Previously I had problems testing this type of component, there are very few examples demonstrating this with `@reach/router`, so I decided to publish this example. 🙂


kentcdodds:
@all-contributors please add @deadkff01 for docs

PR #550, opened 2020-08-06
Fix use of wait in docs

yakirn:
Example is importing `wait` but calling `waitFor`

### DIFF
--- a/docs/pptr-testing-library/intro.md
+++ b/docs/pptr-testing-library/intro.md
@@ -32,7 +32,7 @@ const $email = await getByLabelText($form, 'Email')
 // interact with puppeteer like usual
 await $email.type('pptr@example.com')
 // waiting works too!
-await waitFor(() => getByText('Loading...'))
+await wait(() => getByText('Loading...'))
 ```
 
 A little too un-puppeteer for you? You can attach all the `DOM Testing Library`


kentcdodds:
@yakirn, if you'd like to contribute to pptr-testing-library, I'm sure they'd be more than happy to accept your help :)

kentcdodds:
@all-contributors please add @yakirn for docs

patrickhulce:
will do 👍

patrickhulce:
Actually @kentcdodds we won't be matching this in pptr-testing-library. All the queries for puppeteer *must* be asynchronous because they require hops to the browser and back through evaluateAsync, the new API for wait/waitFor does not support this and we will continue to use `wait-for-expect`

kentcdodds:
That makes sense. Could we at least rename it to `waitFor` so it's consistent with what the rest of the ecosystem is doing?

patrickhulce:
Sure we can export both 👍

kentcdodds:
Great! Once that's done, let's revert this change to keep things consistent.

patrickhulce:
published v0.6.3 and updated our repo examples with `waitFor` 🎉

kentcdodds:
Sweet! https://github.com/testing-library/testing-library-docs/pull/557

yakirn:
> @yakirn, if you'd like to contribute to pptr-testing-library, I'm sure they'd be more than happy to accept your help :)

Hey @kentcdodds sorry for not replying sooner, always happy to help if I can :) Were you talking about this particular issue or in general?

kentcdodds:
I think we're all set at this point, thanks anyway!

PR #551, opened 2020-08-06
docs(ByRole): Add `{ pressed: boolean }` option for querying by aria-pressed state

jluxenberg:
See https://github.com/testing-library/dom-testing-library/pull/729

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -617,6 +617,7 @@ getByRole(
     normalizer?: NormalizerFn,
     selected?: boolean,
     checked?: boolean,
+    pressed?: boolean,
     queryFallbacks?: boolean,
   }): HTMLElement
 ```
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-queries.md
```diff
@@ -677,8 +678,12 @@ assertions about the `Open dialog`-button you would need to use
 The default value for `hidden` can
 [be configured](api-configuration#configuration).
 
-Certain roles can have a selected state. You can filter the returned elements by
-their selected state by setting `selected: true` or `selected: false`.
+#### Querying by ARIA states
+
+##### `aria-selected`
```
eps1lon, line 683: Since not all `selected` elements are `aria-selected` I would omit it from the heading and just name these after their name in the options bag. Applies to checked and pressed as well.

kentcdodds:
@all-contributors please add @jluxenberg for docs

PR #554, opened 2020-08-07
docs(intro): note for  create-react-app OOTB support

rahulbhooteshwar:


kentcdodds:
Whoops, pressed a few weird buttons and somehow both closed and merged this PR 😆

Thanks!

@all-contributors please add @rahulbhooteshwar for docs

nickmccurdy:
This is useful information, but we only know for sure that the official (`cra-template` and `cra-template-typescript`) and Redux (`cra-template-redux` and `cra-template-redux-typescript`) would be using testing library. Would it be worth it to mention that other templates and older versions of CRA may still require manually React Testing Library, or are there too few users in this situation for it to be worth mentioning?

PR #556, opened 2020-08-07
Add `timeout` to configuration docs

majapw:
According to the release notes for `dom-testing-library@7` (https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0), this is configurable at the global level.

I also wanted to fix the docusaurus code tabs because they don't seem to register new lines (which is confusing), but... that seems to also be a problem in the docusaurus documentation itself: https://docusaurus.io/docs/en/doc-markdown#language-specific-code-tabs Don't know if anyone has any ideas on that.


kentcdodds:
@all-contributors please add @majapw for docs

PR #562, opened 2020-08-09
Update the documentation to resolve confusion around AREA semantic an…

### LINKED ISSUE #561: Better documentation of explicit and implicit aria semantics
The current explainer for ByRole regarding default `role` could be expanded a bit:

- warn that `role` should not be added for a test (but for a user)
- link to https://www.w3.org/TR/html-aria/#docconformance for implicit roles and allowed roles instead of https://html.spec.whatwg.org/index.html#contents (navigating a spec is hard)
- and more? See https://github.com/A11yance/aria-query/issues/73

m98:
Developers are wrongly adding `role` attribute to their HTML elements in order to access the element easier, some are not aware of the W3C specifications, some IDE will warn them about the wrong `role` names, but still, we need to clarify by providing more detail about its usage and a link to a better reference which cover more elements and has more examples.

Thanks to @eps1lon for the  #561 

Closes #561

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -625,12 +625,10 @@ getByRole(
 Queries for elements with the given role (and it also accepts a
 [`TextMatch`](#textmatch)). Default roles are taken into consideration e.g.
 `<button />` has the `button` role without explicitly setting the `role`
-attribute. The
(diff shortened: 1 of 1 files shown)


kentcdodds:
@all-contributors please add @m98 for docs

PR #567, opened 2020-08-18
docs: fix `api-queries.md` minor grammar typos and "Note" block styles

shui91:
I noticed there was a small typo (imo) when I was reading  "...will not work in the case where a `for` attr on a label matches and `id` field on a non-form element." and made some changes. 

I've highlighted the text changes in comments as unfortunately the diffs from the changes I made to the "Note" blocks make them hard to immediately spot.

I also took this opportunity to make the "Note" blocks within the page to be consistent with each other. 
The format was based off of this commit I found previously: https://github.com/testing-library/testing-library-docs/commit/5bab9829e25b8c962ffa2ca006b4893c4421762a

i.e. 
Screen Shot 2020-08-18 at 08 58 42

Please let me know if I've missed anything! I also ran `format-docs` just to make sure my changes stayed formatted properly - so there were also a few extra `md` files that were also updated.

### DIFF
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -226,8 +226,8 @@ const inputNode = screen.getByLabelText('Username', { selector: 'input' })
 
 > **Note** 
 >
-> `getByLabelText` will not work in the case where a `for` attr on a `<label>` 
-> element matches and `id` field on a non-form element.
+> `getByLabelText` will not work in the case where a `for` attribute on a
+> `<label>` element matches an `id` attribute on a non-form element.
 
 
 ```js
--- a/docs/dom-testing-library/api-queries.md
+++ b/docs/dom-testing-library/api-queries.md
@@ -38,7 +38,7 @@ more than one element is found after a default timeout of `1000`ms. If you need
 to find more than one element, then use `findAllBy`.
 
 > **Note**
-> 
+>
 > this is a simple combination of `getBy*` queries and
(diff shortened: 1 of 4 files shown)


docs/dom-testing-library/api-queries.md
```diff
@@ -220,8 +224,10 @@ If it is important that you query an actual `<label>` element you can provide a
 const inputNode = screen.getByLabelText('Username', { selector: 'input' })
 ```
 
-Note that it will not work in the case where a `for` attr on a `label` matches
-and `id` field on a non-form element.
+> **Note**
+>
+> `getByLabelText` will not work in the case where a `for` attribute on a
+> `<label>` element matches an `id` attribute on a non-form element.
```
shui91, line 230: I've changed it to read: "a label matches **an** `id` field" instead of **and**. I also removed "it" in favour of explicitly referring to `getByLabelText` and updated references to `attr` and `field` to be `attribute`.

docs/dom-testing-library/api-queries.md
```diff
@@ -106,7 +108,9 @@ screen.debug(screen.getAllByText('multi-test'))
 
 ## Queries
 
-> NOTE: These queries are the base queries and require you pass a `container` as
+> **Note**
+>
+> These queries are the base queries and require you to pass a `container` as
```
shui91, line 113: I just added "to" on this line :)

alexkrolick:
Thanks for these improvements!

PR #568, opened 2020-08-19
Add Migrate from Enzyme page

### LINKED ISSUE #563: Request to add more practical example to the document query page
I know the whole point of this testing library is to do testing "how a user would" and "how user see".

But imagine the time that we have multiple headlines on our page (ex, h2, h3, h4), h2 holds the product name, and h3 holds the product manufacturer. Just by `getAllByRole('heading')` we can't exactly address the specific tag we need to test (ex, product name), there is a need to exactly address which "heading" do we mean. I'm sure there are several good ways to React Testing Library to solve this problem easily, but the documentation is a problem. I saw many StackOverflow questions about the `getByRole`, and many user are even writing "role" even when it's not necessary. On the other hand, accessing by "getByRole" is not that clear for developers who used to work with Enzyme (and JQuery API)

When you look at the documentation, most of the examples are about images, input, label, and button. For edge cases like the example I provided (multiple headlines, how to address each exactly) does not have a good practical example there. Or what if there are several div, and user needs to access one (I mean without using `data-testid`, which is this library's main focus as long as possible)

[The documentation page ](https://testing-library.com/docs/dom-testing-library/api-queries) need some polishing and re-writing, and I would be glad to help

m98:
Hello everyone!

As discussed earlier on #563, we realized a new page for those developers who have experience with Enzyme is needed, for at least the following purposes:
- Why use React Testing Library instead of Enzyme?
- How to use it? (installation, basic usage)
- some helpful link, and explanation to help them get started with the RTL's queries

I did not go into detail on the queries, since it has a complete documentation page itself, and it would be hard to keep two pages updated (I mean if going into detail on this page) instead I just explained the difference and some helpful link to help them get started with the RTL.

Please note that I'm not a native English speaker, so, grammar problem is expected even though I tried to check it with some grammar checker software.

Thanks a lot
Mention: @kentcdodds 

closes #563

### DIFF
--- a/docs/react-testing-library/migrate-from-enzyme.md
+++ b/docs/react-testing-library/migrate-from-enzyme.md
@@ -4,10 +4,10 @@ title: Migrate from Enzyme
 sidebar_label: Migrate from Enzyme
 ---
 This page would not go into detail, but it’s for those who have experience in working with Enzyme and are trying to
- understand how to move to React Testing Library. It also has some helpful information for those who are comparing 
- Enzyme with React Testing Library.
+understand how to move to React Testing Library. It also has some helpful information for those who are comparing 
+Enzyme with React Testing Library.
 
-### What is React Testing Library?
+## What is React Testing Library?
 React Testing Library is part of an open-source project named [Testing Library](https://github.com/testing-library). 
 There are several other helpful tools and libraries in the Testing Library project which you can use them to write more 
 efficient tests. Beside the React Testing Library, here are some other Testing Library’s libraries that can help you 
@@ -20,7 +20,7 @@ you can use to extend jest. These will make your tests more declarative, clear t
 would happen in the browser as the user interacts with it. For example, userEvent.click(checkbox) would change the 
 state of the checkbox.
 
-### Why should I use the React Testing Library?
+## Why should I use the React Testing Library?
 Enzyme is a good test library. The library and its contributors did so much for the community. We thank them all, 
 but it does not mean we should not talk about differences and how we are doing better.
 
@@ -40,13 +40,13 @@ the component’s output.
 Re-writing your tests in React Testing library worth it because you're trading tests that slow you down for tests 
 that give you more confidence and increase your productivity in the long run.
 
-### How to migrate from Enzyme to React Testing Library?
+## How to migrate from Enzyme to React Testing Library?
 One of the keys to a successful migrate is to do it incrementally, by running the two test libraries side by side in 
 the same application and porting Enzyme’s tests to React Testing Library one by one. It makes it possible to migrate 
 even large and complex applications without disrupting other businesses because the work can be done collaboratively 
 and spread over some time. 
 
-### Install React Testing Library
+## Install React Testing Library
 You can check [this page](https://testing-library.com/docs/react-testing-library/setup) for the complete installation 
 and setup guide.
 
@@ -60,7 +60,7 @@ With Yarn
 
 `yarn add --dev @testing-library/react @testing-library/jest-dom`
 
-### Import React Testing Library to your test 
+## Import React Testing Library to your test 
 Let’s say we’re using Jest (you can use other test frameworks as well), then you just have to import the following 
 modules into your test file:
 
@@ -86,7 +86,7 @@ describe('Describe the test', () => {
 
 ```
 
-### Basic Enzyme to React Testing Library migration
+## Basic Enzyme to React Testing Library migration
 One thing to keep in mind that there's not a one-to-one mapping of Enzyme features to React Testing Library features. 
 Many Enzyme features result in inefficient tests, and we are here to learn the React Testing Library’s philosophy and 
 testing approach.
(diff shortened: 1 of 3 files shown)


alexkrolick (commented):
I think this page can trim down on the installation and usage sections (we have other doc pages for that), and focus on migrating specific testing patterns.

- [x] Events - what you have here about `user-event` is good; may want to cover `fireEvent` and discuss differences.
- [x] Finding elements
  - Here's a migration strategy for `wrapper.find(ComponentName)`:
    - Add a testid to the component => Change .find() to use the testid attribute (might now have issues with hostNodes/ReactElementWrappers both showing up) => migrate to RTL.
  - Discuss how tests for component structure can be overconstraining.
  - Show how you can use getByRole, label, etc. instead of testids if your components are ARIA-compliant.
- [ ] Checking prop values in components
  - Address why this is not a pattern commonly used in RTL (validate behavior instead). Show how you can use mocks or spies to do it anyway.
- [ ] Checking component state
  - Discuss how you can validate the render result instead
- [ ] Shallow render
  - Show how you can mock a troublesome component (like a 3rd party lib) that you want to treat as a passthrough/noop componentt in the test env
- [ ] Triggering class methods in tests
  - Discuss how this is not recommended because it overspecifies the component (can you rewrite it with Hooks and still have the tests pass)?
- [ ] Use of act()
  - Not usually necessary in RTL due to wrapping behind the scenes
- [ ] Use of wrapper.update()
  - Not necessary since there is no wrapper
  - Show `await findBy*` and `await waitFor(callback)` for async interactions
 - [ ] Naming conventions
  - Don't call the value returned by `render` wrapper; mention use of `screen` to avoid using that value at all, except for `{rerender}`
- [ ] Updating props
  - Show use of `rerender` method
- [ ] Updating state
  - Don't do it
- [ ] Working with Redux/Context
  - Migrating wrapped/container components
  - Testing "connected" and "pure" components in an integrated and non-integrated fashion
- Differences in queries
  - Enzyme .find() is like jQuery - it can return an array/NodeList. Show how this creates extra handling in tests (like nodes[0]) and allows errors to get farther in the test while throwing non-descriptive error messages. Show `getBy` and `findBy` + `*AllBy` as an alternative - they throw immediately with a detailed DOM printout and you can specify if you want 1 or multiple values.

docs/react-testing-library/migrate-from-enzyme.md
```diff
@@ -0,0 +1,193 @@
...
+
+* **[Jest-dom](https://github.com/testing-library/jest-dom)**: The library provides a set of custom jest matchers that 
+you can use to extend jest. These will make your tests more declarative, clear to read, and to maintain. 
+
+* **[user-event](https://github.com/testing-library/user-event):** user-event tries to simulate the real events that 
+would happen in the browser as the user interacts with it. For example, userEvent.click(checkbox) would change the 
+state of the checkbox.
+
+### Why should I use the React Testing Library?
+The Enzyme is a good test library. The library and its contributors did so much for the community. We thank them all,
```
nickmccurdy, line 24:
```suggestion
Enzyme is a good test library. The library and its contributors did so much for the community. We thank them all, 
```

docs/react-testing-library/migrate-from-enzyme.md
```diff
@@ -0,0 +1,193 @@
...
+You can see the [list of available queries](https://testing-library.com/docs/dom-testing-library/api-queries) in detail. 
+Something else that people have a problem with is that they’re not sure which query should they use, luckily we have 
+a great page which explains [which query to use](https://testing-library.com/docs/guide-which-query), so don’t forget 
+to check it out!
+
+If you still have a problem with the React Testing Library’s queries, and you’re not sure which query to use, just 
+don’t worry! we’ve created a useful Chrome extension named 
+**[which-query](https://chrome.google.com/webstore/detail/testing-library-which-que/olmmagdolfehlpjmbkmondggbebeimoh)**  
+that aims to enable developers to find a better query when writing tests. Right-click on an element on any web page 
+and it will copy the best possible query for that element into your clipboard. Then simply paste into your test code.
```
nickmccurdy, line 137: I think we're recommending Testing Playground instead.

m98:
Thanks for your comments

I was waiting for others to write about their ideas, so I can start working on this page again.

@kentcdodds  do you agree with all of what @alexkrolick said? Thoughts?

kentcdodds:
Yes, I agree with Alex

docs/react-testing-library/migrate-from-enzyme.md
m98: You're right. I'll change that too

m98:
@alexkrolick Thanks for all your good suggestions. 
 I'm working on it, I just have a couple of questions. 

> Add a testid to the component => Change .find() to use the testid attribute (might now have issues with hostNodes/ReactElementWrappers both showing up) => migrate to RTL.

I don't understand about the example you want me to provide, can you please clarify?

This is what I understand:
1. a component is needed
2. A test that users `wrapper.find` should be used
3. Now the test has to be changed to use `testid` (why?)
4. I did not understand hostNodes/ReactElementWrapper
5. What's the point of this migration?  (I mean is there any golden point that you think is good to be answered by providing this specific example?)

In the following day, I might ask some more questions to hopefully complete this page ASAP

Thanks a lot

alexkrolick:
### Question 1
> Add a testid to the component => Change .find() to use the testid attribute (might now have issues with hostNodes/ReactElementWrappers both showing up) => migrate to RTL.

> > I don't understand about the example you want me to provide, can you please clarify?

> > This is what I understand:
> > 1. a component is needed
> > 2. A test that users `wrapper.find` should be used
> > 3. Now the test has to be changed to use `testid` (why?)
> > 4. I did not understand hostNodes/ReactElementWrapper
> > 5. What's the point of this migration?  (I mean is there any golden point that you think is good to be answered by providing this specific example?)

If you have a pretty large test, you may want to rewrite part of the Enzyme test first. Otherwise you are changing both tests and implementation at the same time and that is risky.

The challenge with using test-ids in Enzyme is this:

```jsx
let wrapper = mount(<Foo><Bar data-testid="abc" /></Foo>)
wrapper.find('[data-testid="abc"]').debug()
// ->
// Bar data-testid="abc"
//   div data-testid="abc"
```

Both the div (created by Bar) and the React component (Bar) have the testid attribute. It's not clear which one to use in the test and you might get more than one element when you only want one.

You can skip the Enzyme refactor if you trust that you can manage rewriting the test in RTL from scratch without missing anything. We could start documenting from this assumption I suppose. 


### Question 2
>  Shallow render
> Show how you can mock a troublesome component (like a 3rd party lib) that you want to treat as a passthrough/noop component in the test env

> > As far as I [know](https://github.com/testing-library/react-testing-library/issues/314#issuecomment-470688171), React Testing Library does not support shallow rendering and even it's not something that we should recommend people to use. But I think I'm misunderstood. Can you please clarify how would you do that? 

One reason to shallow render is it doesn't execute the render logic, which is good if a subcomponent has weird dependencies, side-effects, etc.

You can do something like this to create a fake component for a library:

```js
jest.mock('google-maps-react', () => (props) => <google-map data-testid="google-map" {...props} />)
```

### Question 3
> Triggering class methods in tests

>> Does this need any discussion? Like, do people use to write this kind of test in Enzyme? (I guess you mean how they use `wrapper.instance()`)
This might be a general behavior, not just linked to Enzyme, and I don't know how widely it's used, and does it worth to talk about it? Since in the latest Enzyme version, they return `null` for stateless components when you can `wrappper.instance()` (I don't know what they used to return before!) , and more people these days are using stateless components. So this might not be a behavior. But you can correct me because I don't have much experience with Enzyme

People access instance methods on class components and call them all the time. I don't agree that most people are using hooks in legacy codebases, which is where the most Enzyme migration is coming from.


### Question 4
> Working with Redux/Context

>> Should we care about Redux tests? I mean Redux is an implementation detail, and what we care about is that we do an action, and based on run the test against the component output. Should we care about the Redux store? 

I'd cover the `wrapper` (?) options to `mount` and the equivalent option in RTL. Don't have to focus on Redux so much as Context in general (react-intl, styled-components, etc. all require a Provider).

m98:
Dear @kentcdodds and @alexkrolick, I tried my best to add the requested changes. 

In the current context, if there is anything that you think can be changed or improved, please let me know.

This PR is getting a little larger than expected!

Changes are welcome if you feel something is wrong both grammatically or technically.

m98:
@kentcdodds thanks for reviewing and fixing some issues.

I wonder that there were not too many grammar errors in the text! Thank God :)

m98:
@kentcdodds  seems no one else is going to review this!

kentcdodds:
Thanks for your help @m98

kentcdodds:
@all-contributors please add @m98 for docs

m98:
Thanks for accepting the PR @kentcdodds 

I just browsed the Testing library website, but I could not see the "Migrate from Enzyme" page. It might be something that I've done wrongly! 

Can you please check it out?

kentcdodds:
The page is on the site: https://testing-library.com/docs/react-testing-library/migrate

I'm not sure why it's not showing up in the sidebar. I don't know how that works actually 😅 Anyone wanna take a look at this?

MatanBobi:
I think that the name in the sidebar is wrong, it should be as the I'd of the page which is `migrate` and not `migrate-from-enzyme`

PR #573, opened 2020-08-24
Rename `timeout` to `asyncUtilTimeout`

malykhinvi:
Use the actual name for timeout configuration parameter.


kentcdodds:
@all-contributors please add @malykhinvi for docs

PR #578, opened 2020-08-25
Fix failing user-event example

nickmccurdy:
Explanation: https://github.com/testing-library/user-event/issues/429#issuecomment-679856884

### DIFF
--- a/docs/ecosystem-user-event.md
+++ b/docs/ecosystem-user-event.md
@@ -20,7 +20,7 @@ test('click', async () => {
   const { getByRole } = render(<textarea />)
 
   await userEvent.type(getByRole('textbox'), 'Hello, World!')
-  expect(getByRole('textbox')).toHaveAttribute('value', 'Hello, World!')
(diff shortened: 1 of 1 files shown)


docs/ecosystem-user-event.md
```diff
@@ -20,7 +20,7 @@ test('click', async () => {
   const { getByRole } = render(<textarea />)
 
   await userEvent.type(getByRole('textbox'), 'Hello, World!')
-  expect(getByRole('textbox')).toHaveAttribute('value', 'Hello, World!')
+  expect(getByRole('textbox')).toHaveProperty('value', 'Hello, World!')
```
kentcdodds, line 23: While we're here, should we use: https://github.com/testing-library/jest-dom#tohavevalue

```suggestion
  expect(getByRole('textbox')).toHaveValue('Hello, World!')
```

nickmccurdy: Didn't realize that was a thing 😅  thanks

PR #579, opened 2020-08-25
docs: Ecosystem library agnostic

timdeschryver:
This PR makes the docs for user-event and jest-dom library agnostic because these can be used within multiple Testing Libraries.

### DIFF
--- a/docs/ecosystem-jest-dom.md
+++ b/docs/ecosystem-jest-dom.md
@@ -3,7 +3,7 @@ id: ecosystem-jest-dom
 title: jest-dom
 ---
 
-[`jest-dom`][gh] is a companion library for `React Testing Library` that
+[`jest-dom`][gh] is a companion library for Testing Library that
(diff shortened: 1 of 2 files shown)


timdeschryver:
I was thinking the same thing, but I think most of the examples are written for React, so I left it as is.
It doesn't bother me that the examples are written for React because that's the most popular.
Because our API's are aligned, there isn't a big difference anyway.

nickmccurdy:
True, but I'm more concerned about learning curve. While more users are probably familiar with React Testing Library over DOM Testing Library, I think DOM Testing Library examples would still be easier for users that aren't familiar with React yet. And given that the majority of testing library users are using the DOM or some DOM based framework, I think it's safe to assume that more users would have basic DOM knowledge than basic React knowledge.

timdeschryver:
👍 
I will update the examples accordingly, thanks for the input.

timdeschryver:
The examples are updated.

PR #581, opened 2020-08-26
Chore/update react context example

ljosberinn:


ljosberinn:
Sure, curious how you'd setup the changing props for the provider here then. Just extend `options` with `providerProps` or something similar?

```js
const customRender = (ui, { providerProps, ...options }) => {
  return render(
    <NameContext.Provider {...providerProps}>
      {ui}
    </NameContext.Provider>,
    options
  );
}

```
because that'd feel odd to me.

kentcdodds:
That's exactly how I recommend doing it actually: https://github.com/kentcdodds/bookshelf/blob/b783234776aeaab54750068c4ceaf12e864803e6/src/test/app-test-utils.js#L12

ljosberinn:
Yeah I was just confused about adding a `customRender` just for those two tests. But I kinda missed the point given that it's an example anyways, so there you go, I hope :)

kentcdodds:
@all-contributors please add @ljosberinn for docs

PR #587, opened 2020-08-29
docs: fix cheatsheet example

prxg22:
this PR makes `render`'s example in https://testing-library.com/docs/react-testing-library/cheatsheet correct by using `findAllByText` to wait for multiple elements.


eps1lon (approved):
Thanks!

eps1lon (changes requested):
Forgot: We also need to update the destructured method on line 20

nickmccurdy:
Maybe we should also remove the explanation of what's returned by `render` (generally not needed with `screen` and the API docs detail this anyway) and use `userEvent.click`

PR #588, opened 2020-09-01
Replace React Redux recipe with Redux's recipe

nickmccurdy:
@MatanBobi has updated Redux's examples for Testing Library based on our recipe, and I think it would be easier to maintain the recipe by keeping it in one place in Redux (which has additional testing information) and redirecting from our existing page.

**History:** https://github.com/testing-library/testing-library-docs/issues/570#issuecomment-683340570


nickmccurdy:
I had to keep the Markdown file for this page but replace it with a stub so Docusaurus would still keep it in the sidebar. This should work with Netlify redirects, but please let me know if there's a better way.

docs/example-react-redux.md
```diff
@@ -3,127 +3,5 @@ id: example-react-redux
...
-    store,
-  })
-  fireEvent.click(screen.getByText('+'))
-  expect(screen.getByTestId('count-value')).toHaveTextContent('1000')
-  fireEvent.click(screen.getByText('-'))
-  expect(screen.getByTestId('count-value')).toHaveTextContent('1000')
-})
-```
+Moved to
+[Writing Tests | Redux](https://redux.js.org/recipes/writing-tests#connected-components)
```
eps1lon, line 7: Is this even visible with a 301 redirect in place?

eps1lon (approved):

MatanBobi (approved):
This one is tricky, not that I think of a different way to do it but if the anchor `connected-components` will be removed we'll just redirect to the page (not that bad) but if the name of the markdown file will change we will have a broken link here.
I can't think of a different way to do it though..

docs/example-react-redux.md
nickmccurdy: In production no, assuming my Netlify config change works properly. Unfortunately though if I delete this file or empty its contents, Docusaurus doesn't show it in the sidebar properly, and I still want that visible so users can find the redirect.

nickmccurdy:
Yea I don't think there's a way to fully avoid that issue, though it seems like Redux should be keeping the recipe URLs in place even with its doc reworks. I'll keep try to track of it just in case, and I'll update the heading anchor if https://github.com/reduxjs/redux/pull/3876 is merged.

nickmccurdy:
The redirect is broken and I'm hoping to fix it in #598. I'll keep this merged since the link still works and I still consider it to be an improvement.

PR #594, opened 2020-09-03
Documentation: Update async section of cheatsheet

jpenna:
Remove  replace `wait` for `waitFor` and deprecated methods

### DIFF
--- a/docs/react-testing-library/cheatsheet.md
+++ b/docs/react-testing-library/cheatsheet.md
@@ -108,13 +108,8 @@ See [Which query should I use?](guide-which-query.md)
 
 See [dom-testing-library Async API](dom-testing-library/api-async.md)
 
-- **wait** (Promise) retry the function within until it stops throwing or times
+- **waitFor** (Promise) retry the function within until it stops throwing or times
   out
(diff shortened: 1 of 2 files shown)


docs/react-testing-library/cheatsheet.md
```diff
@@ -108,13 +108,8 @@ See [Which query should I use?](guide-which-query.md)
 
 See [dom-testing-library Async API](dom-testing-library/api-async.md)
 
-- **wait** (Promise) retry the function within until it stops throwing or times
+- **waitFor** (Promise) retry the function within until it stops throwing or times
```
alexkrolick, line 111: How about adding a "deprecated" warning for the old methods and a comment about when they were removed? Some people may be referring to this page from an old version of RTL.

jpenna: I can change it to tell it was deprecated. I can't find when it happened, do you know it? Maybe just tell it is deprecated is enough?

alexkrolick: https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0

alexkrolick (approved):

PR #596, opened 2020-09-05
docs(svelte): Adds setup information for testing-library-svelte

timvancleef:
Adds a missing jest transform config setting when using es6 modules.

### DIFF
--- a/docs/svelte-testing-library/setup.md
+++ b/docs/svelte-testing-library/setup.md
@@ -51,6 +51,22 @@ with any testing framework and runner you're comfortable with.
     transform setting (it is set by default, but since we are overriding the
     transform config, we have to add it explicitly)
 
+    5.1 Install `babel-jest`
+
+    ```bash
+    npm install --save-dev babel-jest
+    ```
+
+    5.2. Add a basic `.babelrc` configuration
+
+    ```json
+    {
+      "presets": [["@babel/preset-env", { "targets": { "node": "current" } }]]
+    }
+    ```
+
+    5.3. Update the Jest transform configuration
+
     ```json
     "transform": {
       "^.+\\.js$": "babel-jest",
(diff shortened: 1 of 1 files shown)


docs/svelte-testing-library/setup.md
```diff
@@ -47,25 +47,36 @@ with any testing framework and runner you're comfortable with.
     }
     ```
 
-5.  This is optional but it is recommended, you can install
+5.  If you are using es6 modules in your project you have to add jest's babel
```
m98, line 50: would be nicer to say "ES6" and "Jest" instead of "es6" and "jest"

m98 (changes requested):
I don't have any comment on if this change is necessary or not. But I only found two minor issues which I liked to mention.

docs/svelte-testing-library/setup.md
timvancleef: Agreed and done 🙂

m98 (commented):
I have not any other feedback. 

@mihar-22 and @benmonro do you have any feedback?

mihar-22:
Mmm I'm not really sure because there is a section right now for people who need Babel: https://testing-library.com/docs/svelte-testing-library/setup#babel--preprocessor. I guess the question is whether we should add it by default... and I think we should since I can't imagine anyone testing Svelte with Jest is not using ES modules. A few recommendations:

- Add an install instruction to step 5: `npm install babel-jest -D`
- Add a default `.babelrc` setup to step 5:

```json
{
  "presets": [["@babel/preset-env", { "targets": { "node": "current" } }]]
}
```
- Remove Babel from the bottom of the setup docs and just leave all the information about preprocessors.

timvancleef:
Hi @mihar-22, @m98 do you have any further opinions on this PR?

mihar-22:
Sorry for the delayed response @timvancleef, I have nothing to add looks good to me.

PR #598, opened 2020-09-06
Netlify config fixes

nickmccurdy:
- [x] Reorganize and simplify configuration for safety and readability
- [x] ~~Use Yarn in production~~
- [x] Enable redirects outside of production (includes local development and Netlify preview builds)
- [x] Fix React Redux redirect (#588)


nickmccurdy:
For some reason the recipe redirects work properly on Netlify but not Netlify Dev. You can work around this by temporarily deleting the `/docs/*` redirect. I may reach out to Netlify.

netlify.toml
```diff
@@ -1,147 +1,123 @@
...
-
   # Directory (relative to root of your repo) that contains the deploy-ready
-  # HTML files and assets generated by the build. If a base directory has
-  # been specified, include it in the publish directory path.
+  # HTML files and assets generated by the build.
   publish = "website/build/testing-library-docs"
 
   # Default build command.
-  command = "cd website; npm install; npm run format-docs; npm run build"
+  command = "npm i -g yarn; cd website; yarn; yarn format-docs; yarn build"
```
eps1lon, line 7: `yarn` should already be installed in netlify. If you need a specific version you can check it into the repo with [`yarn set policies`](https://classic.yarnpkg.com/en/docs/cli/policies/#toc-policies-set-version)

MatanBobi: I think that's true only if the root folder contains a `yarn.lock` file, since I see here that we `cd website`, it looks like the root folder isn't configured properly in netlify so maybe that's the reason why we needed the installation.

eps1lon: That's an odd design decision though I can see why you'd want to do this anyway. For apps you should always have a lockfile. I think we just caried over the no-lockfile policy from the lib repos.

MatanBobi: I actually didn't even remember we don't have a lock file.. But that explains the need for global `yarn` installation.

nickmccurdy: I can confirm. I tested this out without manually installing Yarn and got errors that it wasn't installed (since we haven't committed `yarn.lock`). Alternatively it may make more sense to just switch to npm, but I thought it would be best to keep things simple for this PR.

nickmccurdy:
It looks like you're supposed to have separate `docs` and `website` directories for Docusaurus 1. However 2 will lift the `package.json` to the root so I'll skip that in this PR and maybe help out with the migration.

netlify.toml
nickmccurdy: Out of curiosity let's have a vote on keeping vs removing a lockfile for this repo (since it's a site not a package).

👍  Commit `yarn.lock` or switch to npm for consistently and commit `package-lock.json`
👎  Keep ignored and possibly switch to npm since we're not sharing `yarn.lock`

netlify.toml
```diff
@@ -1,147 +1,85 @@
...
   # Directory (relative to root of your repo) that contains the deploy-ready
-  # HTML files and assets generated by the build. If a base directory has
-  # been specified, include it in the publish directory path.
+  # HTML files and assets generated by the build
   publish = "website/build/testing-library-docs"
 
-  # Default build command.
-  command = "cd website; npm install; npm run format-docs; npm run build"
+  # Default build command
+  command = "npm i -g yarn; cd website; yarn; yarn format-docs; yarn build"
```
MatanBobi, line 7: Guess we'll need to change this once https://github.com/testing-library/testing-library-docs/pull/607 is merged..

alexkrolick: Done, we can revert this part

alexkrolick:
```suggestion
  command = "cd website; npm install; npm run format-docs; npm run build"
```

alexkrolick (approved):

kentcdodds (approved):
Solid. Thank you for working on this.

PR #606, opened 2020-09-09
Updated React Router title

TOPOFGR:
Details:

-Changed `Reach` for `React`

### DIFF
--- a/docs/example-reach-router.md
+++ b/docs/example-reach-router.md
@@ -1,6 +1,6 @@
 ---
 id: example-reach-router
-title: Reach Router
+title: React Router
 ---
 
 ```jsx


nickmccurdy:
We should probably update the URL too, right?

TOPOFGR:
> We should probably update the URL too, right?

Yeah and the Sidebar too

nickmccurdy:
Do you want to do that or would you like help?

nickmccurdy:
I just realized this may be a mistake. We already have a page for React Router, this is for Reach Router which is a different library.

TOPOFGR:
> I just realized this may be a mistake. We already have a page for React Router, this is for Reach Router which is a different library.

Ok, my bad. Sorry

nickmccurdy:
No problem, thanks for contributing anyway

PR #607, opened 2020-09-09
Use npm instead of Yarn

nickmccurdy:


MatanBobi:
I'm actually a big fan of `yarn` so I'm trying to understand why we want to switch it here :)

nickmccurdy:
I prefer Yarn too, but the organization prefers npm and I think it's better to keep things consistent so contributors to libraries only need to learn npm to contribute to the docs.

MatanBobi:
> I prefer Yarn too, but the organization prefers npm and I think it's better to keep things consistent so contributors to libraries only need to learn npm to contribute to the docs.

Got it, didn't know that the organization prefers npm :)
Thanks for clarifying this!

.gitignore
```diff
@@ -7,8 +7,6 @@ lib/core/MetadataBlog.js
 
 website/translated_docs
 website/build/
-website/yarn.lock
```
MatanBobi, line 10: Do we want to add `package-lock` to gitignore in order to keep the "no lock file approach" or are we planning to fix it with this PR?

MatanBobi (approved):

.gitignore
alexkrolick: I think we want to allow the lockfile for the site to alleviate build indeterminacy.

.gitignore
```diff
@@ -7,8 +7,6 @@ lib/core/MetadataBlog.js
...
-website/node_modules
```
alexkrolick, line 11: I am guessing the `node_modules` above catches this, making it redundant?

nickmccurdy: Yes

PR #611, opened 2020-09-10
fix(redirects): redirect error message url to correct page

MatanBobi:
As @eps1lon suggested in https://github.com/testing-library/dom-testing-library/pull/764.
At the moment when there's no document defined and we're trying to use `screen` there's an error message with a broken link.
This change is to add a redirect to the correct place if a user tries to enter the broken link.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -145,3 +145,10 @@
   to = "https://redux.js.org/recipes/writing-tests#connected-components"
   status = 301
   force = true
+
+# Redirect screen error messe to correct page
+[[redirects]]
+  from = "https://testing-library.com/s/screen-global-error"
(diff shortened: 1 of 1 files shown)


eps1lon (approved):
Probably just forgotten when adding docs for https://github.com/testing-library/dom-testing-library/pull/412

It's the first link under `/s/`. @kentcdodds Was `/s/` intended to group permalinks?

kentcdodds:
I don't know what the `/s/` this is all about . If I did that, I can't remember why 😅

MatanBobi:
@eps1lon so do you think we can merge this one? :)

MatanBobi:
Tested in production and working. Thanks @eps1lon!

PR #614, opened 2020-09-12
Fix baseElement's default value in react docs

psalaets:
I changed the default value for `baseElement` in render options to be consistent with

- The default value of render result's `baseElement` (it's documented a little further down the page)
- What I'm seeing in my usage of RTL 11.0.2


eps1lon (approved):
Yep: https://github.com/testing-library/react-testing-library/blob/v11.0.2/src/pure.js#L43

eps1lon:
@all-contributors add @psalaets for docs

PR #616, opened 2020-09-15
docs: add findByText example

### LINKED ISSUE #381: More examples for FindBy* queries
Currently, there aren't really any code samples showing the `findBy` variants of the queries. Would be nice to make the async queries more prominent, whether in the Recipes section or under the usage examples of the query docs.

shermanhui:
I'm opening a PR for issue #381 to add a simple JS login
form as an example of how to use the `findByText` query 
to find DOM elements that may not be visible immediately. 

I'm open to any feedback as I haven't written tests for 
vanilla JS in a long time (too spoiled with React). ~~Specifically, I
think I need some clarification on an issue I ran into with my example tests
not passing on Codesandbox. Here is the [link](https://codesandbox.io/s/suspicious-jepsen-qlqhi?file=/src/findBy.test.js) to the sandbox,
it's interesting that the first set of tests pass when 
I change the display value from `display: none` to `display: inline`,
but the same doesn't happen with the mock error message `span` 
elements 🤔.~~

~~Here's a screenshot of a type error I ran into with regards to using 
`userEvent.type` with the input element..perhaps that's why I'm not 
seeing a value in the `input` element when I debug and therefore
the element style is not being updated.~~

EDIT: Turns out it was failing because I was using `document.getElementById` and `document`
is actually empty as we're using the `renderContent` function to render the form.

The live tests can be seen [here](https://codesandbox.io/s/suspicious-jepsen-qlqhi?file=/src/findBy.test.js)

To fix it, I just used `el.querySelector` instead and this got the tests to pass! 🎉 

Screen shot of `hasAttribute` type error

Screen shot of error on Codesandbox
 
Thanks in advance!

Also let me know if I should add React specific examples for `findByText`
or any other `findBy` queries!

Closes #381

### DIFF
--- a/docs/example-findByText.md
+++ b/docs/example-findByText.md
@@ -1,0 +1,147 @@
+---
+id: example-findByText
+title: example `findByText`
+sidebar_label: findByText
+---
+
+```javascript
+// src/__tests__/example.test.js
+// This is an example of how to use findByText to query for text that
+// is not visible right away
+
+import {
+  getByRole,
+  findByText,
+  getByPlaceholderText,
+} from '@testing-library/dom'
+import userEvent from '@testing-library/user-event'
+// provides a set of custom jest matchers that you can use to extend jest
+// i.e. `.toBeVisible`
+import '@testing-library/jest-dom'
+
+const renderContent = el => {
+  el.innerHTML = `
+    <form id='login_form' method='post' name='login'>
+      <label for='username'>User Name:</label>
+
+      <input
+        type='text'
+        name='username'
+        id='username_input'
+        placeholder='Enter user name'
+      />
+
+      <span id='username_required_error' style='color: red; display: none;'>
+        User Name Required
+      </span>
+
+      <span id='invalid_username_error' style='color: red; display: none;'>
+        Invalid User Name
+      </span>
+
+      <label for='password'>Password:</label>
+
+      <input
+        type='password'
+        name='password'
+        id='password_input'
+        placeholder='Enter password'
+      />
+
+      <span id='invalid_password_error' style='color: red; display: none;'>
+        Invalid Password
+      </span>
+
+      <span id='password_required_error' style='color: red; display: none;'>
+        Password Required
+      </span>
+
+      <button id='submit' type='submit'>
+        Login
+      </button>
+    </form>
+  `
+
+  const submitButton = el.querySelector('#submit')
+  const formEl = el.querySelector('#login_form')
+
+  submitButton.addEventListener('click', () => {
+    const userInput = el.querySelector('#username_input')
+    const passwordInput = el.querySelector('#password_input')
+
+    var userName = userInput.value
+    var password = passwordInput.value
+    if (!userName) {
+      el.querySelector('#username_required_error').style.display = 'inline'
+    }
+
+    if (!password) {
+      el.querySelector('#password_required_error').style.display = 'inline'
+    }
+
+    if (userName && userName !== 'Bob') {
+      el.querySelector('#invalid_username_error').style.display = 'inline'
+    }
+
+    if (password && password !== 'theBuilder') {
+      el.querySelector('#invalid_password_error').style.display = 'inline'
+    }
+  })
+
+  formEl.addEventListener('submit', function(evt) {
+    evt.preventDefault()
+    window.history.back()
+  })
+
+  return el
+}
+
+describe('findByText Examples', () => {
+  let div
+  let container
+
+  beforeEach(() => {
+    div = document.createElement('div')
+    container = renderContent(div)
+  })
+
+  it('should show a required field warning for each empty input field', async () => {
+    userEvent.click(
+      getByRole(container, 'button', {
+        name: 'Login',
+      })
+    )
+
+    expect(
+      await findByText(container, "User Name Required")
+    ).toBeVisible();
+
+    expect(
+      await findByText(container, "Password Required")
+    ).toBeVisible();
+  })
+
+  it('should show invalid field errors for each invalid input field', async () => {
+    const userNameField = getByPlaceholderText(container, 'Enter user name')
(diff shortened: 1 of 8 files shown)


docs/eample-findBy*.md
```diff
@@ -0,0 +1,142 @@
...
+    ).toBeVisible()
+  })
+
+  it('should show an async invalid password error', async () => {
+    const userNameField = getByPlaceholderText(container, 'Enter user name')
+    const passwordField = getByPlaceholderText(container, 'Enter password')
+
+    userEvent.type(userNameField, 'wrongName')
+    userEvent.type(passwordField, 'wrongPassword')
+    // fireEvent.focus(passwordField);
```
shermanhui, line 117: I'll remove the commented code - but just wanted to highlight I tried with both `useEvent` and `fireEvent` when making my examples on Codesandbox!

docs/eample-findBy*.md
```diff
@@ -0,0 +1,142 @@
...
+  getByRole,
+  findByText,
+  getByPlaceholderText,
+} from '@testing-library/dom'
+import userEvent from '@testing-library/user-event'
+// provides a set of custom jest matchers that you can use to extend jest
+// i.e. `.toBeVisible`
+import '@testing-library/jest-dom'
+
+const renderContent = el => {
```
shermanhui, line 23: This `renderContent` method is inspired by the example from the [example intro](https://testing-library.com/docs/dom-testing-library/example-intro)

alexkrolick:
Any other maintainers have thoughts on if we would want a recipe like this? We have docs under the Appearance/Disappearance section that are similar. https://testing-library.com/docs/guide-disappearance

docs/eample-findBy*.md
```diff
@@ -0,0 +1,142 @@
+---
+id: example-findBy*
```
alexkrolick, line 2: Note: I am not sure if we want to add this file. But if we do, can we avoid using non-url safe characters like `*`, and fix the filename?

docs/eample-findBy*.md
alexkrolick: Let's remove the comments

docs/eample-findBy*.md
```diff
@@ -0,0 +1,142 @@
...
+    userEvent.click(
+      getByRole(container, 'button', {
+        name: 'Login',
+      })
+    )
+    // await waitFor(() => {
+    //   expect(getByText(container, "Invalid Password")).toBeVisible()
+    // })
+    // screen.debug(await findByText(container, "Invalid Password"));
+    expect(await findByText(container, 'Invalid User Name')).toBeVisible()
```
alexkrolick, line 138: I don't think `findByText` excludes hidden elements by default, does it? That's why this assertion fails.

docs/eample-findBy*.md
shermanhui: Done @alexkrolick! I used `*` as I was trying to follow the convention in the docs but since this example file only contains a `findByText` example, I've renamed it to `example-findByText`.

I've also added this new example to the site navigation.

docs/eample-findBy*.md
shermanhui: Ah it turns out it was my own mistake. The test wasn't finding the input element at all, I should have been using `el.querySelector` instead of `document.getElementById` 😅 . The tests pass as expected now!

docs/example-findByText.md
```diff
@@ -0,0 +1,178 @@
...
+      getByRole(container, "button", {
+        name: "Login"
+      })
+    );
+
+    expect(
+      // https://stackoverflow.com/questions/55509875/how-to-query-by-text-string-which-contains-html-tags-using-react-testing-library
+      await findByText(container, (_, element) => {
+        const hasText = (element) =>
+          element.textContent.replace(/(\r\n|\n|\r)/gm, "").trim() ===
```
shermanhui, line 130: I formatted the `html` in `renderContent` and apparently `textContent` will preserve line breaks and new lines, I've used this regex to remove it. Let me know if there's a better way to handle this!

I also used the custom text matcher function from the above StackOverflow answer.

shermanhui:
@alexkrolick I've fixed the tests on Codesandbox and pushed the requested changes, let me know if there's anything else I can add to help out with closing #381! :)

I committed some unrelated formatted files as well since I ran `format-docs` just to make sure my changeset was compliant with the current documentation format.

website/README.md
```diff
@@ -110,6 +110,9 @@ title: This Doc Needs To Be Edited
 My new content here..
 ```
 
+Note: Ensure the file name and the id value do not include non-url safe
+characters i.e. '\*'.
```
shermanhui, line 114: Updated the website README to make sure future contributors don't use non-url safe characters.

website/sidebars.json
```diff
@@ -138,7 +139,7 @@
       "example-reach-router",
       "example-react-transition-group",
       "example-react-modal",
-      "example-external"
+      "example-update-props"
```
shermanhui, line 142: Reorganized these examples to be in alphabetical order

docs/example-findByText.md
alexkrolick: I would try to keep the incidental complexity in the example to a minimum; focus on showing the key info about the findBy queries. Is there a way you can trim it down?

shermanhui: Thanks, @alexkrolick, that's a good point - I've changed the example so the warning shows the same way that the "invalid user name/password" text does to avoid inline HTML elements to keep the complexity down! Let me know if there's anything else that needs fixing up.

shermanhui:
@alexkrolick just a friendly ping to re-review the changes made - hope we can merge this in! :)

shermanhui:
Thanks @alexkrolick :) Would it be cool if I got added to the Contributors list as well? Thanks again for the approval + merge 🎉

alexkrolick:
@all-contributors please add @shermanhui  for docs

PR #617, opened 2020-09-16
docs: add docs page for query-extensions companion lib

tjefferson08:
Here's a proposal for some documentation that links out to [query-extensions](https://github.com/tjefferson08/query-extensions)

I kept it short and sweet, but I brought over the motivating example from the repo README.md for maximum hype-building 🎉

### DIFF
--- a/docs/ecosystem-query-extensions.md
+++ b/docs/ecosystem-query-extensions.md
@@ -1,0 +1,44 @@
+---
+id: ecosystem-query-extensions
+title: query-extensions
+---
+
+[`query-extensions`][gh] is a companion library for Testing Library that mixes a
+higher-level API into the standard @testing-library core queries
+
+```
+npm install --save-dev query-extensions
+```
+
+```jsx
+test('query-extensions can help us write more maintainable tests', async () => {
+  const ui = {
+    successIcon: { filter: 'role', params: ['img', { name: 'celebration' }] },
+    signUpBtn: { filter: 'role', params: ['button', { name: /sign up/i }] },
+    emailInput: { filter: 'labelText', params: ['Your email'] },
+    loading: { filter: 'text', params: ['Loading...'] },
+  }
+
+  render(<YourComponent />)
+
+  // component starts in loading state
+  expect(screen.get(ui.loading)).toBeInTheDocument()
+
+  // loads up an email input, loading disappears
+  const emailInput = await screen.find(ui.emailInput)
+  expect(screen.query(ui.loading)).toBeNull()
+
+  // fill out email and click to sign up
+  fireEvent.change(emailInput, { target: { value: 'email@example.com' } })
+  fireEvent.click(screen.get(ui.signUpBtn))
+
+  // success modal pops up and takes over component (hiding other content)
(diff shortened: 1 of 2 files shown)


docs/ecosystem-query-extensions.md
```diff
@@ -0,0 +1,44 @@
...
+
+[`query-extensions`][gh] is a companion library for Testing Library that mixes a
+higher-level API into the standard @testing-library core queries
+
+```
+npm install --save-dev query-extensions
+```
+
+```jsx
+test('query-extensions can help us write more maintainable tests', async () => {
```
eps1lon, line 14: Somewhat similar concern like Kent: Using this library doesn't magically make your tests more maintainable. Maybe it improves some tests but for one-off tests using this approach will likely result in a less maintainable test suite. Maybe this is just me but I don't have a problem with my tests being not-DRY and this library looks like it's optimizing for DRY code.

Also as a general recommendation from my experience is writing tests with as little indirection as possible. If your test fails at line 30 and you have to look at line 5 to understand what you expected then you not only have to go back and forth in your code you also have to hold more stuff in your head.

But don't take this as general advise. All my testing opinions are mainly concerned with low-level component tests and not app tests. I expect that there are some differences in the testing approach. Just want to caution against "use X to make your code more Y" which implies that code is some universal construct.

tjefferson08: Thanks for the thoughtful comment ❤️ 

>Maybe this is just me but I don't have a problem with my tests being not-DRY and this library looks like it's optimizing for DRY code.

To be perfectly clear: I wouldn't write a 40 line test like this: I'd repeat a few things and be happy with trading a bit of brittleness for simplicity.

And that'll get you a long way, as I'm sure you've seen. FWIW, I regularly provide feedback at work that encourages un-dry-ing things (especially tests).

My hope is that this library will help folks out if/when they decide that they've reached a tipping point and are willing to put up with some indirection in order to better manage large suites of tests.

Again, thanks for your feedback! Please LMK if you'd like me to change anything specific or close this out entirely (or ship, of course 😂 ) I understand that additions to the ecosystem section still incur maintenance costs for core folks, so no worries if you don't want this.

tjefferson08:
Hey @kentcdodds! Curious to hear your take on the tradeoffs of linking out to `query-extensions`. Again, no worries if y'all decide you'd rather close this for now.

Happy to tweak specific wording/phrasing as well, if that's preferable to closing outright.

kentcdodds:
Thanks for putting this together Travis!

Why don't we start with something a little smaller (no example in the docs) with a note that this is currently in an "Experimental" phase and not officially recommended by the Testing Library team (yet). Then we'll see how this plays out for people. Does that sound good?

tjefferson08:
Seems reasonable to me 👍  I'll prep that change sometime this week! Thanks for the quick reply and decisive action 🥇

tjefferson08:
Hey @kentcdodds! What do you think of this dialed-back documentation?

cc @eps1lon 

Appreciate y'all's time and energy ❤️

kentcdodds (approved):
LGTM 👍 Thanks

PR #620, opened 2020-09-18
api-async: warn flaky with expect dot assertions

alavkx:
https://github.com/testing-library/testing-library-docs/issues/619

### DIFF
--- a/docs/dom-testing-library/api-async.md
+++ b/docs/dom-testing-library/api-async.md
@@ -35,6 +35,8 @@ await waitFor(() => expect(mockAPI).toHaveBeenCalledTimes(1))
 // ...
 ```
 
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-async.md
```diff
@@ -35,6 +35,8 @@ await waitFor(() => expect(mockAPI).toHaveBeenCalledTimes(1))
 // ...
 ```
 
+⚠️  Using this utility with `expect.assertions` will result in flaky tests. Beware!
```
alexkrolick, line 38: If we were to add an explanation here, I would suggest explaining the exact problem, rather than saying it makes your test "flaky".

```suggestion
`waitFor` may run the callback a variable number of times
```

alavkx: Excellent suggestion

PR #622, opened 2020-09-24

MichaelDeBoey:
Bring in line with https://github.com/kentcdodds/react-testing-library-examples/blob/master/src/__tests__/react-router.js


kentcdodds (approved):
Much improved.

PR #623, opened 2020-09-24
Update example-update-props.md

MichaelDeBoey:
Bring in line with https://github.com/kentcdodds/react-testing-library-examples/blob/master/src/__tests__/update-props.js


kentcdodds:
Thanks!

PR #634, opened 2020-10-04
Netlify config fixes... fixes

nickmccurdy:
I fixed some issues with the Netlify config in #598 and related merges.

### DIFF
--- a/docs/example-react-redux.md
+++ b/docs/example-react-redux.md
@@ -1,7 +1,0 @@
----
-id: example-react-redux
-title: React Redux
----
-
-Moved to
-[Writing Tests | Redux](https://redux.js.org/recipes/writing-tests#connected-components)
--- a/netlify.toml
+++ b/netlify.toml
@@ -41,6 +41,11 @@
   from = "/docs/example-react-redux"
   to = "https://redux.js.org/recipes/writing-tests#connected-components"
 
+# Screen error message to correct page
+[[redirects]]
+  from = "/s/screen-global-error"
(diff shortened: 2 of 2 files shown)


nickmccurdy:
By removing the stub file, Docusaurus will no longer put the React Redux recipe link in the sidebar. However, the redirect will still work from any old links, and the recipe link still exists in the Redux docs.

Are we okay with this? If we want to keep the link, I could revert the stub and force the redirect. I think having an extra file lying around is ugly, but it seems to be required for Docsaurus' sidebar, at least in version 1.

kentcdodds:
I'm fine with this personally. Thanks @nickmccurdy!

MatanBobi:
> By removing the stub file for the React Redux recipe, Docusaurus will no longer put the link in the sidebar. However, the redirect will still work from any old links, and the recipe link still exists in the Redux docs.
> 
> Are we okay with this? If we want to keep the link, I could revert the stub and force the redirect. However, I think having an extra file lying around just to get Docusaurus to show it in the sidebar is kinda messy. That may hopefully get easier with Docusaurus 2 though.

@nickmccurdy 
When you talk about Docusaurus 2 do you mean to render a simple [`Redirect`](https://v2.docusaurus.io/docs/docusaurus-core#redirect) component for that route?

nickmccurdy:
I just meant in general, wasn't sure specifically how Docusaurus 2 would let us solve the problem. The redirect component could work but I'd rather have a way to redirect without the extra file (like our existing Netlify config).

PR #640, opened 2020-10-10
Added Formik Recipe

luistak:
**What** 

Added a basic async Formik example test recipe based in [this example](https://formik.org/docs/examples/async-submission)  as requested in #536


luistak:
Is this PR good to go?

website/sidebars.json
```diff
@@ -133,6 +133,7 @@
       "example-findByText",
       "example-input-event",
       "example-react-context",
+      "example-react-formik",
```
nickmccurdy, line 136: If we move this after`useReducer`, the React core examples will be grouped before the third party library examples.

docs/example-formik.md
```diff
@@ -0,0 +1,81 @@
...
+```jsx
+// myForm.test.js
+import React from 'react'
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+
+import { MyForm } from './myForm.js'
+
+test('rendering and submiting a basic Formik form', async () => {
+  const mockedOnSubmit = jest.fn()
```
nickmccurdy, line 64: Let's name this `handleSubmit` to follow React's naming convention for matched `on*` and `handle*` props.

nickmccurdy (changes requested):

luistak:
Hey thank you @nickmccurdy I was kinda busy these days I'm glad you fixed those comments ❤️

PR #641, opened 2020-10-11
ecosystem page for react-hooks-testing-library

rawat07102:
#20

### DIFF
--- a/docs/ecosystem-react-hooks-testing-library.md
+++ b/docs/ecosystem-react-hooks-testing-library.md
@@ -1,0 +1,45 @@
+---
+id: ecosystem-react-hooks-testing-library
+title: react-hooks-testing-library
+---
+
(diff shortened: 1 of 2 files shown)


MatanBobi:
Hi @rawat07102!
Thanks for this PR..
If you want in to automatically link the issue, you'll need to write something like: "Resolves #20", then the issue will be automatically closed once this PR will be merged..

rawat07102:
thanks for the suggestion @MatanBobi and it does make sense for it to be under React section but it does not depend on react-testing-library. So I guess it depends on if we want to separate based on dependency or use case.

MatanBobi:
> thanks for the suggestion @MatanBobi and it does make sense for it to be under React section but it does not depend on react-testing-library. So I guess it depends on if we want to separate based on dependency or use case.

I agree, it's a little tricky but logically I'd go with that, I guess more opinions would be good..
Also, I saw that under ecosystem we already have `rtl-simple-queries` which is clearly related only to React..

alexkrolick:
This page should go under the ecosystem section. It's where all of the libraries that aren't framework adapters go.

MatanBobi:
@rawat07102 can you please resolve the conflicts so we'll be able to merge this one? :)

PR #648, opened 2020-10-15
Update intro.md

corehtml5canvas:
"You may want to avoid following" has a different meaning than "You may want to avoid the following". I believe the intent is the latter.


alexkrolick (approved):

nickmccurdy:
@all-contributors add @corehtml5canvas for docs

PR #656, opened 2020-10-25
docs: add testing-library-selector to ecosystem

domasx2:
Hi! Thanks for a great testing lib. Wrote a small lib for reusable selectors https://github.com/domasx2/testing-library-selector

### DIFF
--- a/docs/ecosystem-testing-library-selector.md
+++ b/docs/ecosystem-testing-library-selector.md
@@ -1,0 +1,44 @@
+---
+id: ecosystem-testing-library-selector
+title: testing-library-selector
+---
+
+[`testing-library-selector`][gh] is a library for `@testing-library` that
+provides reusable selectors. Written in typescript.
+
+```
+npm install --save-dev testing-library-selector
+```
+
+```typescript
+import { byLabelText, byRole, byTestId } from './selector'
+
+// define reusable selectors
+const ui = {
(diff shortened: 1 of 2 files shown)


nickmccurdy (approved):
This example and your library look good thanks.

**Pipe dream:** Maybe it would be simpler if we just curried all of the query functions. If anyone's interested I can try to make an alternate API.

domasx2:
thanks! rebased on latest master

domasx2:
> Pipe dream: Maybe it would be simpler if we just curried all of the query functions. If anyone's interested I can try to make an alternate API.

domasx2:
> Pipe dream: Maybe it would be simpler if we just curried all of the query functions. If anyone's interested I can try to make an alternate API.

I've been thinking about something similar... There's a common use case where I want to scope queries to a particular container, typically identified with a `data-testid`. 
Currently we typically get the container dom node, then query against it. However, sometimes React will recreate the container dom node or elements, making the old dom node reference stale. So I'm toying with the idea to extend this library with smth like this:
```
const inputInContainer = byTestId('foo-container').byLabelText('my input')
```
then `inputInContainer.get()` would get the container first, if that succeeds, get the input inside it, and so on

nickmccurdy:
@all-contributors add @domasx2 for docs

PR #659, opened 2020-10-29
chore: update React imports

MichaelDeBoey:
https://twitter.com/kentcdodds/status/1321885232878923776

### DIFF
--- a/docs/example-input-event.md
+++ b/docs/example-input-event.md
@@ -11,16 +11,16 @@ sidebar_label: Input Event
 > [`user-event`](ecosystem-user-event.md)
 
 ```jsx
-import React, {useState} from 'react'
+import React, { useState } from 'react'
 import { render, fireEvent } from '@testing-library/react'
 
 function CostInput() {
   const [value, setValue] = useState('')
 
-  removeDollarSign = value => (value[0] === '$' ? value.slice(1) : value)
-  getReturnValue = value => (value === '' ? '' : `$${value}`)
-  
-  handleChange = ev => {
+  removeDollarSign = (value) => (value[0] === '$' ? value.slice(1) : value)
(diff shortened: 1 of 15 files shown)


docs/example-input-event.md
```diff
@@ -11,30 +11,25 @@ sidebar_label: Input Event
 > [`user-event`](ecosystem-user-event.md)
 
 ```jsx
-import React, {useState} from 'react'
+import * as React from 'react'
+import { useState } from 'react'
```
eps1lon, line 15: What is the point here? React.useState seems sufficient since `react` isn't tree-shakeable anyway. Note that react isn't published as an ES6 module and anything but `const React = require('react')` is just makeup.

MichaelDeBoey: This is a first step into removing React import statements, which won't be necessary anymore with React 17

eps1lon: Ok. Why do we need to import twice?

MichaelDeBoey: I've done this, so we can easily remove all `import * as React from 'react'` lines whenever we update everything to React 17

eps1lon: But it's confusing now for readers so that we have to do less work now. Seems like we're optimizing at the wrong end. Especially since this is markdown only where it doesn't matter. 

`import * as React from 'react'` works in both 16 and 17 so I'm trying to understand who is benefiting from the style you're proposing.

MichaelDeBoey: I'll revert all double imports then 🙂

MatanBobi:
Hi @MichaelDeBoey!
I'm closing this one since There's no need to import `React` anymore (as you probably know) :)
Maybe we should revisit this to remove the React imports we have.

Thanks for your time and sorry for not merging this in time!

PR #669, opened 2020-11-06
fix: edit button url contains duplicate docs string

### LINKED ISSUE #667: Edit this page button results in 404
**Describe the bug**
The *Edit this page* button goes to the wrong place

**To Reproduce**
1. Go to https://testing-library.com/docs/angular-testing-library/intro
2. Scroll down to the bottom
3. Click on *Edit this page*
4. See you are taken to https://github.com/testing-library/testing-library-docs/blob/master/docs/docs/angular-testing-library/intro.mdx

**Expected behavior**
Should be taken to https://github.com/testing-library/testing-library-docs/blob/master/docs/angular-testing-library/intro.mdx

It has a duplicate `/docs/docs/` in the URL.

**Screenshots**
Screen Shot 2020-11-06 at 6 17 19 pm

 - OS: MacOS
 - Browser: Chrome

MatanBobi:
Resolves #667


docusaurus.config.js
```diff
@@ -217,7 +217,7 @@ module.exports = {
           showLastUpdateAuthor: true,
           showLastUpdateTime: true,
           editUrl:
-            'https://github.com/testing-library/testing-library-docs/blob/master/docs/',
+            'https://github.com/testing-library/testing-library-docs/blob/master/',
```
nickmccurdy, line 220: This does fix the issue, but while we're at it we should probably link directly to the edit page so users don't have to press the edit button.

```suggestion
            'https://github.com/testing-library/testing-library-docs/edit/master/',
```

nickmccurdy (changes requested):

docusaurus.config.js
MatanBobi: You're right, didn't think about it :)
Thanks!

PR #676, opened 2020-11-11
fix: broken user-event link on the page

RichardBray:
Tried using [gh] but I wasn't able to get the desired effects in the Github previewer so I fell-back to this option. Please let me know what the best option is to fix this.

### DIFF
--- a/docs/guide-events.mdx
+++ b/docs/guide-events.mdx
@@ -38,7 +38,7 @@ simply using `fireEvent.click` is worth it.
 We will describe a couple of simple adjustments to your tests that will increase
 your confidence in the interactive behavior of your components. For other
 interactions you may want to either consider using
-[`user-event`](https://testing-library.com/docs/ecosystem-user-event/) or testing your components in a real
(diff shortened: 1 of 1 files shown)


docs/guide-events.mdx
```diff
@@ -38,7 +38,7 @@ simply using `fireEvent.click` is worth it.
 We will describe a couple of simple adjustments to your tests that will increase
 your confidence in the interactive behavior of your components. For other
 interactions you may want to either consider using
-[`user-event`](ecosystem-user-event) or testing your components in a real
```
MatanBobi, line 41: I think that the only thing missing here is the `.mdx` suffix, can you please give it a try and add it instead of using the full site url? That won't help us on local env :)
Thanks!

MatanBobi (changes requested):

docs/guide-events.mdx
RichardBray: I've updated the code but I feel like it'll be better if I test this locally, just to make sure it all works.

PR #680, opened 2020-11-12
docs: fix broken links

MatanBobi:
This is now ready for review.
I added mdx suffix to missing files, removed the relative paths since they aren't needed and also added some titles to a few snippets I saw :)


alexkrolick (commented):
👍 

I wish adding the `.mdx` extension wasn't necessary. The file extension of the source markup  doesn't affect the link for the compiled site :/

But it's important to get these broken links fixed, so I am OK with merging

MatanBobi:
> 👍 
> 
> I wish adding the `.mdx` extension wasn't necessary. The file extension of the source markup  doesn't affect the link for the compiled site :/
> 
> But it's important to get these broken links fixed, so I am OK with merging

I definitely agree.. maybe we can take it up with docusaurus people..

alexkrolick:
Yeah, let's open an issue about the extension in links on the docusaurus side

MatanBobi:
@alexkrolick looks like in the docs they never claim that omitting the file extension will work, I probably found it redundant because it worked on localhost:
https://v2.docusaurus.io/docs/markdown-features#referencing-other-documents

Do you think we should maybe open a feature request for that? Since they never said it's supported I guess it's not a bug..

alexkrolick:
It works without the extension because the links pass through the markdown parser as-is, and the folder structure of the docs matches the output directory structure, so relative links remain the same

MatanBobi:
Sorry but I'm not sure I 100% understand what we want from docusaurus then.
Basically what you're saying is that it will work as long as the file is already there (was loaded), so I'm not sure if that's a bug or a new feature..

PR #685, opened 2020-11-23
docs: remove notes about some polyfills

nickmccurdy:
We mention some polyfills that don't seem to be necessary any more and could cause confusion when setting up libraries.

# Removed
- `core-js`: modern Node versions should already support most useful ES features, and common setups like CRA already polyfill anything that isn't supported
- `regenerator-runtime`: this hasn't been needed since Node 6 which has lost support years ago
- `babel-polyfill` and `@babel/polyfill`: deprecated

# Not removed
- Babel 6 module fix for Jest: could still be an issue for older projects that haven't upgraded to Babel 7 yet
- pre-Node 13 `full-icu`: some LTS releases of Node are older than 13 so this is still important to support

### DIFF
--- a/docs/dom-testing-library/setup.mdx
+++ b/docs/dom-testing-library/setup.mdx
@@ -32,7 +32,3 @@ With mocha, the test command would look something like this:
 ```
 mocha --require jsdom-global/register
 ```
-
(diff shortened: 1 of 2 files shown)


MatanBobi (approved):
Looks good 👍🏻

PR #686, opened 2020-11-23
docs: add setup information for Vue

nickmccurdy:


docs/vue-testing-library/setup.mdx
```diff
@@ -3,4 +3,47 @@ id: setup
...
+[any testing framework](#using-without-jest) (Vue Testing Library does not
+require that you use Jest).
+
+## Using without Jest
+
+If you're running your tests in the browser bundled with webpack (or similar)
+then `Vue Testing Library` should work out of the box for you. However, most
+people using Vue Testing Library are using it with the Jest testing framework
+with the `testEnvironment` set to `jest-environment-jsdom` (which is the default
+configuration with Jest).
```
nickmccurdy, line 19: Is this accurate? As far as I'm aware pretty much everyone uses Vue CLI, but its plugins to enable tests are disabled by default.

docs/vue-testing-library/setup.mdx
```diff
@@ -3,4 +3,47 @@ id: setup
 title: Setup
 ---
 
-`Vue Testing Library` does not require any configuration to be used.
+`Vue Testing Library` does not require any configuration to be used. However,
+there are some things you can do when configuring your testing framework to
+reduce some boilerplate. In these docs we'll demonstrate configuring Jest, but
+you should be able to do similar things with
+[any testing framework](#using-without-jest) (Vue Testing Library does not
+require that you use Jest).
```
nickmccurdy, line 11: Are we missing some other important setup for Jest such as `.vue` support or is that included by default?

docs/vue-testing-library/setup.mdx
afontcu: Yeah, that's right, you need to manually select Jest as the testing framework. To be honest I've never used jsdom in isolation. Notice that Vue Single File Components requires an extra bundling step, so it might be less straightforward

docs/vue-testing-library/setup.mdx
afontcu: Hm, Single File Components requires an extra bundling step, so it might be less straightforward. For instance, we need to add "vue" to `moduleFileExtensions`, and also install vue-jest to process *.vue files. 

Docs from the official library: https://vue-test-utils.vuejs.org/installation

docs/vue-testing-library/setup.mdx
nickmccurdy: Does Vue CLI give the option to install Jest for new projects or should we link to some docs for the plugin?

afontcu: Yep it does :) https://cli.vuejs.org/core-plugins/unit-jest.html#injected-commands

It is also available on the installation wizard 👍

nickmccurdy: I guess I'll leave this alone then, since most users that would want to test would be enabling it in the wizard

docs/vue-testing-library/setup.mdx
brleinad: I just had to set this up today and I can confirm that for single file components it does require following the steps described on the link shared by @afontcu above.

MatanBobi:
@nickmccurdy @afontcu is this still relevant or we can close it? :)

PR #692, opened 2020-11-27
Update example-formik.md

dandelionadia:
onSubmit has two arguments and we need to ignore the second one. (https://formik.org/docs/api/formik#onsubmit-values-values-formikbag-formikbag--void--promiseany)
If we do not do this then we will get an error: 
```
expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

      {"email": "email@example.com", "password": "password2398"},
    + {"resetForm": [Function anonymous], "setErrors": [Function anonymous], "setFieldError": [Function anonymous], "setFieldTouched": [Function anonymous], "setFieldValue": [Function anonymous], "setFormikState": [Function anonymous], "setStatus": [Function anonymous], "setSubmitting": [Function anonymous], "setTouched": [Function anonymous], "setValues": [Function anonymous], "submitForm": [Function anonymous], "validateField": [Function anonymous], "validateForm": [Function anonymous]},
```

### DIFF
--- a/docs/example-formik.md
+++ b/docs/example-formik.md
@@ -75,7 +75,7 @@ test('rendering and submiting a basic Formik form', async () => {
       email: 'jhon.dee@someemail.com',
       firstName: 'Jhon',
       lastName: 'Dee',
-    })
+    }, expect.anything())
   )
 })
 ```


eps1lon (approved):
Thanks!

I thought jest would default to ignoring excess args. But it does indeed fail if not every arg is asserted on: https://codesandbox.io/s/winter-dream-g87iv?file=/src/index.test.js.

eps1lon:
We're reverting this change in https://github.com/testing-library/testing-library-docs/pull/876

We didn't expose the formik bag to begin with. The assertion is only needed if you do pass the formik bag to `onSubmit` which we're not doing in the demo.

PR #694, opened 2020-11-29
fix a broken link in /users

AntonNiklasson:


nickmccurdy (approved):
Thanks, we must have forgotten to fix this when upgrading to Docusaurus 2

nickmccurdy:
@all-contributors add @AntonNiklasson for docs

PR #701, opened 2020-11-30
Use tsconfig/jsconfig paths instead of NODE_PATH

### LINKED ISSUE #693: Usage of NODE_PATH in react-testing-library with create-react-app v4
Hello 👋 

[create-react-app](https://create-react-app.dev/) no longer supports usage of `NODE_PATH` in `.env` since [v4](https://github.com/facebook/create-react-app/releases/tag/v4.0.0):
> ### Removed `typescript` flag and `NODE_PATH` support
> We've removed the deprecated `typescript` flag when creating a new app. Use `--template typescript` instead. We've also dropped deprecated `NODE_PATH` flag as this has been replaced by setting the base path in `jsconfig.json`.

So I think we should update this part of the docs to reflect that:
https://github.com/testing-library/testing-library-docs/blob/99ef60b9acc6193a70fb9e01d1a584ffd15c8f67/docs/react-testing-library/setup.mdx#L211-L234

My project doesn't run on CRA v4 yet so I can't tell how exactly what the alternative should be, but I will look into it.

nickmccurdy:
Fixes #693

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -175,7 +175,7 @@ relative imports (`../../test-utils`), add the folder containing the file to the
 Jest `moduleDirectories` option.
 
 This will make all the `.js` files in the test-utils directory importable
-without `../`.
+without `../`. This can't be used with Create React App.
 
 ```diff title="my-component.test.js"
 - import { render, fireEvent } from '../test-utils';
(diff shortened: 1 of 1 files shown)


MatanBobi (approved):

docs/react-testing-library/setup.mdx
```diff
@@ -175,7 +175,7 @@ relative imports (`../../test-utils`), add the folder containing the file to the
 Jest `moduleDirectories` option.
 
 This will make all the `.js` files in the test-utils directory importable
-without `../`.
+without `../`. This can't be used with Create React App.
```
MatanBobi, line 178: Maybe we can put this one in bold or something to highlight it?

MatanBobi:
Looks good, just left one small comment :)

nickmccurdy:
I agree that it should be prominent, but I'm not sure I like how the bold text looks split up

MatanBobi:
Yeah I agree..
What about a [docusaurus style callout](https://v2.docusaurus.io/docs/markdown-features#calloutsadmonitions)?

nickmccurdy:
That's a good idea for the future, but I'd rather stay consistent and use the same `> **Note**` format as the rest of the code so far.

MatanBobi (approved):

MatanBobi:
Looks good. Merging 👍🏻

PR #703, opened 2020-12-01
Make date format clearer to understand

Ukhu:
While it was specified that the date value should follow the ISO 8601 format, this change makes it much clearer for the reader to instantly tell what the ISO 8601 format is without opening another tab to check, just like I did while reading the doc. By using "24" instead of "12", it makes it clear that that portion of the format is referring to the day and not month since there's no 24th month in a calendar year


alexkrolick (approved):

alexkrolick:
Good thinking

PR #708, opened 2020-12-08
Update faq.mdx

cseas:
Add information on how to find types.

**Context**:

I was struggling to find the types for defining a custom render function in React Testing Library and the documentation gave absolutely no hint about how to do it in TypeScript.

I was only able to figure my way out after I stumbled onto this comment: 
https://github.com/testing-library/react-testing-library/issues/634#issuecomment-613097614

### DIFF
--- a/docs/react-testing-library/faq.mdx
+++ b/docs/react-testing-library/faq.mdx
@@ -252,6 +252,14 @@ For a more in-depth discussion of this topic see
 
 </details>
 
+<details>
+  
+<summary>How do I use this library with typescript?</summary>
+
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Thanks for the effort, but I don't think this is necessary. Now that the types are included in this repository via the standard `types` field in `package.json`, users can find the types through any editor with TypeScript integration. If users want to contribute to the types, it should be easy for them to find the declaration file from the `types` field.

cseas:
Maybe it's a standard practice I'm unaware of? Can you please elaborate a bit on how can I find types for a package like Testing Library using VS Code?

nickmccurdy:
The de-facto standard is to put a `types` (or `typings`) property in `package.json` with the path to the type declarations. React Testing Library uses `"types": "types/index.d.ts"`. In VSCode you can `ctrl`/`command`-click an import or run `Go To Definition` (also bound to context menu and `F12`) as a shortcut

PR #713, opened 2020-12-08
Refer to Ada Lovelace as a programmer

pamuche:
Maybe Ada isn't just famous for beeing a woman than being the first Programmer :)
Thank you for this nice human-like testing library.

Kudos


nickmccurdy (approved):
Yea, I think it's best to focus on her technical accomplishments for inclusivity, since she has a common feminine name. I've also updated the React Native docs in the same way. Thanks!

PR #714, opened 2020-12-15
Add a "gotchas" section to the testcafe docs

citypaul:
Hi there,

Thanks for this great library! This is the first time I've tried to contribute, so hopefully I've done everything correctly 👍 

I spent some time just now trying to make a testcafe assertion work like this:

```javascript
  await t
    .expect(
      screen.getByRole("heading", {
        name: /manage clients/i,
        level: 1,
      }).exists
    )
    .ok();
```

The docs do explicitly state that this is not possible already, however, there was no code example of this, and me being a typical dev, I scanned through the code examples first. It took me some time to go through the docs and read line by line, and it was only then that I spotted the aside. I felt adding a "gotchas" section might help other devs who may also just be skimming through looking for the code :)

### DIFF
--- a/docs/testcafe-testing-library/intro.mdx
+++ b/docs/testcafe-testing-library/intro.mdx
@@ -189,7 +189,6 @@ test('works with results from "byAll" query with index - regex', async (t) => {
 ```javascript
 import "testcafe";
 import { screen } from "@testing-library/testcafe";
-import { config } from "../config";
 
 fixture`gotchas`
   .page`http://localhost:3000`;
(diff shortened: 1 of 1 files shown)


citypaul:
Hey, any chance of a comment on this? If you don't think it adds any value or it's not appropriate that's all good. I just thought it m ight be useful to some people so it could be worth adding :)

MatanBobi:
Hi @citypaul! Sorry it took some time to get to it.
I'm not too familiar with testcafe but if you believe it's missing we can add something to clarify these scenarios better.
I think we should add blocks, I'm not sure if calling it `gotchas` is the best description, maybe we can write the notes you pointed out as specific notes?
Also, I think it would be better to write the stuff you've added in the comments as plain text in the markdown and not as comments inside the blocks.

Thanks for taking the time to put this up and once again sorry it took us time to respond.

citypaul:
Hey @MatanBobi , sure thanks for that.

I'll try to take a look at this shortly and will change it a bit. I thought the "gotchas" title might be an issue, but I wasn't sure what to go with really. I'll have a go at changing the structure a bit soon and will come back to you with another PR if that helps.

I think it would probably help people using the library, as it definitely caused me some confusion at the time. Thanks! 👍

PR #720, opened 2020-12-18
docs: Explicitly mention what is thrown

aleks-rope:
Just updating a sentence to make more sense.

From:
- _This throws_ if more than one match is found (use `queryAllBy` instead).

To: 
- _This throws an error_ if more than one match is found (use `queryAllBy` instead).


eps1lon (approved):
Thanks!

eps1lon:
@all-contributors add @aleks-rope for docs

PR #727, opened 2020-12-27
docs: fix the cypress-testing-library types link

mvasin:
The link should be updated since the types [moved](https://github.com/testing-library/cypress-testing-library/pull/148) from DefinitelyTyped into the repo.


afontcu (approved):
nice catch, thanks!

PR #735, opened 2021-01-13
docs: reorganize query docs

### LINKED ISSUE #485: Unify "Queries" page
There should be a single source of truth for queries.

Current scenario: each flavor has its own Queries page. It is hard to keep them updated, and also it bloats the sidebar and search results.

Desired scenario:
* A single, properly structured Queries page, detached from DOM Testing Lib.
* Also with content from https://testing-library.com/docs/guide-which-query
* Removal of "flavored-based" queries page.
* Redirects from old pages to the new unified one.

alexkrolick:
- give queries a top-level nav entry instead of
  nesting under DOM Testing Library
- give each query its own page
- reorganize pages like "Which Query" under the new heading
- add some pointers to related APIs that users will need after querying
  - #734, #489, #490, #469
- minor content tweaks and formatting fixes
- update internal links and redirects

Closes #485


docs/queries/about.mdx
```diff
@@ -0,0 +1,418 @@
+---
```
alexkrolick, line 1: This and the other pages in the queries folder are the main changes in this PR

kentcdodds (approved):
I'm a big fan of this improvement 👍

this is awesome

docs/dom-testing-library/api-async.mdx
```diff
@@ -1,13 +1,33 @@
...
 
-All the async utils are built on top of `waitFor`.
+The async methods return Promises, so be sure to use `await` or `.then` when
+calling them.
+
+## `findBy` Queries
+
+`findBy` methods are a combination of `getBy`
+[queries](queries/about.mdx#types-of-queries) and [`waitFor`](#waitfor). They
+accept the waitFor options as the last argument (i.e.
```
eps1lon, line 18:
```suggestion
accept the waitFor options as the last argument (e.g.
```

"e.g." <-> "For example"
"i.e." <-> "In other words"

-- https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference

Though it seems like we used this before (and we probably use it in unchanged pages as well) so we should probably do this in a follow-up.

docs/queries/about.mdx
```diff
@@ -0,0 +1,418 @@
...
+resemble how users interact with your code (component, page, etc.) as much as
+possible. With this in mind, we recommend this order of priority:
+
+1. **Queries Accessible to Everyone** queries that reflect the experience of
+   visual/mouse users as well as those that use assistive technology
+   1. `getByRole`: This can be used to query every element that is exposed in
+      the
+      [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/AOM).
+      With the `name` option you can filter the returned elements by their
+      [accessible name](https://www.w3.org/TR/accname-1.1/). This should be your
```
eps1lon, line 107: Did we link to this previously as well? I seem to remember we link to more high level explainer but that might've been on another page. 

At this point I don't find this particular link helpful.

eps1lon (commented):
Really appreciated that!

> give queries a top-level nav entry instead of
nesting under DOM Testing Library

I expected it to appear in the left sidebar and it took me some time to find the new link. What does everyone think about moving it to the left sidebar?

[`Frameworks > DOM Testing Library > Queries`](https://deploy-preview-735--testing-library.netlify.app/docs/dom-testing-library/api-queries) now just lists a single link to `/docs/queries/about/`. Would it be simpler to redirect from `/docs/dom-testing-library/api-queries` to `/docs/queries/about/` and remove internal usage of `/docs/dom-testing-library/api-queries`?

MatanBobi:
I really like this change.
I agree with @eps1lon that we're used to finding these at the left sidebar but I think it's helpful to have it at a top level.
We also wanted to unify the [async utilities](https://github.com/testing-library/testing-library-docs/issues/489) and the [firing events](https://github.com/testing-library/testing-library-docs/issues/490) sections.
I'm trying to see if this solution is scalable or maybe we need to create a more generic title?

docs/queries/about.mdx
alexkrolick: Didn't change that. Personally, I have used that resource and found it helpful.

docs/dom-testing-library/api-async.mdx
```diff
@@ -122,7 +142,20 @@ waitForElementToBeRemoved(() => getByText(/not here/i)).catch((err) =>
 
 The options object is forwarded to `waitFor`.
 
-## `wait` (DEPRECATED, use waitFor instead)
+## Deprecated Methods
+
+`wait`, `waitForDomChange`, and `waitForElement` have been combined into the
+`waitFor` method
```
afontcu, line 148:
```suggestion
`waitFor` method.
```

docs/dom-testing-library/api-async.mdx
```diff
@@ -144,7 +177,9 @@ Unlike wait, the callback parameter is mandatory in waitFor. Although you can
 migrate an existing `wait()` call to `waitFor( () => {} )`, it is considered bad
 practice to use an empty callback because it will make the tests more fragile.
 
-## `waitForDomChange` (DEPRECATED, use waitFor instead)
+### `waitForDomChange`
+
+> (DEPRECATED, use waitFor instead)
```
afontcu, line 182:
```suggestion
> (DEPRECATED, use `waitFor` instead)
```

docs/dom-testing-library/api-helpers.mdx
```diff
@@ -140,8 +140,8 @@ could do:
 ```js
 import { within } from '@testing-library/dom'
 
-const { getByText } = within(document.getElementById('messages'))
-const helloMessage = getByText('hello')
+const messages = document.getElementById('messages')
+const helloMessage = within(messages).getByText('hello')
```
afontcu, line 144: way more clear 👌

afontcu (approved):
This looks great 🎉

alexkrolick:
> remove internal usage of /docs/dom-testing-library/api-queries

I think I got rid of these, but left the page in the DOM TL tree so that it would at least contain the full API if you are browsing library docs directly. It might make sense to add some more text later that explains the relationship between the core APIs, DOM Testlng Lib, and framework wrappers.

> should this go in the sidebar?

One thing I like about giving it its own section is that we can keep the nav list expanded on page load. I could see "Core API" or just "API" being the title, in order to make room for an "interactions" (events) section, but I think the top nav having both "Docs" and "API" would be confusing

> what if we add async and events sections?

I have been thinking we basically need all the concepts from my talk here https://www.youtube.com/watch?v=b8SOFNc_X_A pulled into a docs page. On top of that, we should give all of the top-level APIs the same treatment as the Queries, so you can browse most of them at a glance from the left nav rather than clicking into pages.

alexkrolick:
How does this look? I took out the top-nav Queries and moved it to a prominent section in Docs, and moved the event-related pages in the same section. 

Screen Shot 2021-01-13 at 12 23 17 AM

eps1lon (approved):
Looks great 👍

alexkrolick:
I'm pretty happy with how this brings up some more sections into the top-level and makes it apparent where there are gaps. As @afontcu pointed out, events could use a cleanup. If anyone wants to jump on that soon, feel free.

MatanBobi:
Looks great, I think that's a really good separation :)

MatanBobi:
@alexkrolick I think that this PR accidentally pushed a `yarn.lock` file and i18n files that were on the old sites..
Am I missing anything or we can remove those? :)

alexkrolick:
@MatanBobi you're right, feel free to clean up

PR #741, opened 2021-01-18
updated `toHaveLength` example

medayz:
updated `toHaveLength` example to avoid confusion:
The existing example asserts that two elements exist, and this can be a little confusing since the section is named "Asserting elements are not present"


afontcu (approved):
nice catch!

PR #745, opened 2021-01-18
fix: broken links in api for react and vue

### LINKED ISSUE #736: Broken Link
**Describe the bug**
I find out a broken link in the documentation

**To Reproduce**
1. Go to 'https://testing-library.com/docs/react-testing-library/api#debug'
2. Click on 'DOM Testing Library' below the code snippet.
3. See link is broken

MatanBobi:
Resolves #736.

### DIFF
--- a/docs/react-testing-library/api.mdx
+++ b/docs/react-testing-library/api.mdx
@@ -115,7 +115,7 @@ const { getByRowColumn, getByText } = render(<MyTable />, {
 })
 ```
 
-See [helpers](dom-testing-library/api-helpers.mdx) for guidance on using utility
+See [helpers](dom-testing-library/api-custom-queries.mdx) for guidance on using utility
 functions to create custom queries.
 
 Custom queries can also be added globally by following the
@@ -128,7 +128,7 @@ The `render` method returns an object that has a few properties:
 ### `...queries`
 
 The most important feature of `render` is that the queries from
-[DOM Testing Library](dom-testing-library/api-queries.mdx) are automatically
+[DOM Testing Library](queries/about.mdx) are automatically
 returned with their first argument bound to the [baseElement](#baseelement),
 which defaults to `document.body`.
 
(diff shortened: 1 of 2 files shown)


eps1lon (commented):
Question: Broken as in "they return 404" or broken as in "they return 301"? 

We need to add redirects as well to fix all existing links that we don't control as well.

MatanBobi:
> Question: Broken as in "they return 404" or broken as in "they return 301"?
> 
> We need to add redirects as well to fix all existing links that we don't control as well.

They return 404. There's no redirect mechanism for those urls since they are just leading to a place that doesn't exist.
The problem is that on build time docusaurus understands that it can't find the markdown file (since dom-testing-library/api-queries.mdx doesn't exist anymore) so it adds that route to the current location(docs/react-testing-library/dom-testing-library/api-queries.mdx), so creating a redirect rule might be problematic because there are many cases we'd need to solve.

eps1lon:
> There's no redirect mechanism for those urls since they are just leading to a place that doesn't exist.

With [netlify you can create redirects](https://docs.netlify.com/routing/redirects/).

I'd be shocked if docosaurus has no API to move a page. That's a fundamental part of the web.

MatanBobi:
They do have, what I'm trying to say is that you need to predict the broken url so you'll be able to redirect it to the right page or redirect to a fallback page.
What I'm trying to say is that whenever we delete a markdown file/remove it, the urls that docusaurus generate will lead to a broken page if not fixed, that's the case we had here.

alexkrolick:
I added the redirects here: 
https://github.com/testing-library/testing-library-docs/blob/master/netlify.toml#L57-L59

But I think @MatanBobi is saying the URL is malformed if you reference the deleted .mdx..md files so the redirect doesn't even kick in

MatanBobi:
> I added the redirects here: 
> https://github.com/testing-library/testing-library-docs/blob/master/netlify.toml#L57-L59
> 
> But I think @MatanBobi is saying the URL is malformed if you reference the deleted .mdx..md files so the redirect doesn't even kick in

Exactly, thanks @alexkrolick :)

PR #747, opened 2021-01-20
chore: cleanup old website data

MatanBobi:


eps1lon (commented):
Could you add the lockfiles to our `.gitignore`? They should already be part of it if they're in the root directory.

MatanBobi:
@eps1lon done, also found another package-lock file at the root so I removed that also in this PR.

eps1lon:
Huh, now I'm confused. I thought we didn't have lockfiles. Let's just remove the yarn.lock and then rethink package managment later. I suspect we want to use npm workspaces.

MatanBobi:
> Huh, now I'm confused. I thought we didn't have lockfiles. Let's just remove the yarn.lock and then rethink package managment later. I suspect we want to use npm workspaces.

I think I pushed the package-lock file by accident a couple months ago, based on what I know we don't want to use lock files in the org.

I can revert that change and just keep the package-lock if you think it will be best for now :)

eps1lon:
The "no lockfile policy" is applied to libraries. The docs are an app so a lockfile makes more sense. 

The main goal here is to avoid contributor confusion (yarn vs npm). If the npm lockfile causes problems or the contribution workflow needs work we can discuss these problems separately.

MatanBobi:
> The "no lockfile policy" is applied to libraries. The docs are an app so a lockfile makes more sense.
> 
> The main goal here is to avoid contributor confusion (yarn vs npm). If the npm lockfile causes problems or the contribution workflow needs work we can discuss these problems separately.

Got it, thanks for the explanation :)
Reverted the `package-lock` removal and the `.gitignore` now only contains `yarn.lock`.

PR #760, opened 2021-02-14
Replace jsdom-global with global-jsdom

just-boris:
When trying to use `jsdom-global` with React tests, you will see these warnings in the console:

```
This browser doesn't support requestAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills
This browser doesn't support cancelAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills
```

This package does not expose these functions. However, there is a modern fork, `global-jsdom` which offers the same API, but also exports all global DOM APIs properly.

I propose changing the mention to the new package, as it is a more future-proof option.

cc @modosc who is maintaining global-jsdom

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -294,19 +294,19 @@ configuration with Jest).
 `jsdom` is a pure JavaScript implementation of the DOM and browser APIs that
 runs in Node. If you're not using Jest and you would like to run your tests in
 Node, then you must install jsdom yourself. There's also a package called
-`jsdom-global` which can be used to setup the global environment to simulate the
+`global-jsdom` which can be used to setup the global environment to simulate the
 browser APIs.
 
-First, install `jsdom` and `jsdom-global`.
+First, install `jsdom` and `global-jsdom`.
 
 ```
-npm install --save-dev jsdom jsdom-global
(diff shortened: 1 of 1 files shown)


eps1lon (approved):
jsdom-global doesn't seem to be maintained anymore. The issue regarding rAF hasn't received any response in 2 years: https://github.com/rstacruz/jsdom-global/issues/45

MatanBobi:
> jsdom-global doesn't seem to be maintained anymore. The issue regarding rAF hasn't received any response in 2 years: [rstacruz/jsdom-global#45](https://github.com/rstacruz/jsdom-global/issues/45)

If that's the case, I'm +1 on merging this :)

MatanBobi (approved):

nickmccurdy (approved):
They also explicitly use the MIT license, nice suggestion.

modosc:
sounds good - i just explicitly enabled [pretendToBeVisual](https://github.com/modosc/global-jsdom/issues/97) by default in the 8.0.0.beta release which provides those apis.  i'll probably bump it up to an official release this week unless i hear back from anyone about it (the other changes are related to `.esm` support).

PR #770, opened 2021-02-25
🔧 Fix broken links across several doc pages

jegtnes:
Hi there! Thanks for overall a very comprehensive set of docs. I just got started on migrating our team from Enzyme to react-testing-library today, having never used it before, and was impressed with how good the docs are.

I did come across a few broken links on the way when looking at the Migrating from Enzyme page, today, though, so thought I'd try to fix what I found! I pulled the docs repo down and ran it locally and it did warn about 20 unresolvable internal links, only some of which were to the old queries page under dom-testing-library, so I tried to fix the remaining ones, too.

Keep in mind I'm not super familiar with the project yet, so some of the references to new pages may not be entirely accurate. I'm particularly not sure about the reorganisation of helper/debugging methods, so please shout if some of that could be improved, but I'm fairly certain that all the links to the new top-level queries documentation are going to be roughly accurate.

I did notice there's a redirect set up for these, but they don't always appear to work, particularly for relative links and links with an internal document anchor. So some of these links I'm updating may have technically worked before thanks to the redirect, but figured there'd be no harm in updating the source to point at the correct docs to begin with :)

### DIFF
--- a/docs/preact-testing-library/learn.mdx
+++ b/docs/preact-testing-library/learn.mdx
@@ -15,10 +15,10 @@ If you're still hungry for more at this point than checkout:
 
 - The dom-testing-library:
   - [Introduction](intro.mdx)
-  - [Queries](dom-testing-library/api-queries.mdx)
+  - [Queries](queries/about.mdx)
   - [Firing Events](dom-testing-library/api-events.mdx)
   - [Async Utilities](dom-testing-library/api-async.mdx)
-  - [Helpers](dom-testing-library/api-helpers.mdx)
+  - [Helpers](dom-testing-library/api-custom-queries.mdx)
   - [FAQ](dom-testing-library/faq.mdx)
 - The react-testing-library:
   - [API](react-testing-library/api.mdx)
--- a/docs/react-testing-library/cheatsheet.mdx
+++ b/docs/react-testing-library/cheatsheet.mdx
@@ -130,16 +130,18 @@ See [Events API](dom-testing-library/api-events.mdx)
 
 ## Other
 
-See [Helpers API](dom-testing-library/api-helpers.mdx),
-[Config API](dom-testing-library/api-configuration.mdx)
+See [Querying Within Elements](dom-testing-library/api-within),
+[Config API](dom-testing-library/api-configuration.mdx),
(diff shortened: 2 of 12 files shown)


docs/react-testing-library/cheatsheet.mdx
```diff
@@ -130,16 +130,18 @@ See [Events API](dom-testing-library/api-events.mdx)
 
 ## Other
 
-See [Helpers API](dom-testing-library/api-helpers.mdx),
-[Config API](dom-testing-library/api-configuration.mdx)
+See [Querying Within Elements](dom-testing-library/api-within),
```
m98, line 133: I guess you forgot the `.mdx` at the end of the link

m98 (commented):
Hi! Good job, and thank you for your time and contribution 👍

docs/preact-testing-library/learn.mdx
```diff
@@ -15,10 +15,10 @@ If you're still hungry for more at this point than checkout:
 
 - The dom-testing-library:
   - [Introduction](intro.mdx)
-  - [Queries](dom-testing-library/api-queries.mdx)
+  - [Queries](queries/about.mdx)
   - [Firing Events](dom-testing-library/api-events.mdx)
   - [Async Utilities](dom-testing-library/api-async.mdx)
-  - [Helpers](dom-testing-library/api-helpers.mdx)
+  - [Helpers](dom-testing-library/api-custom-queries.mdx)
   - [FAQ](dom-testing-library/faq.mdx)
```
alexkrolick, line 22: We can probably remove this list entirely instead of trying to maintain the TOC in two places - just link to one page

alexkrolick (commented):
Thanks - didn't realize so many of these were still pointing to the wrong place even with redirects

eps1lon:
In addition to updating them, we should also add redirects. Otherwise links that we don't control remain broken.

MatanBobi:
> In addition to updating them, we should also add redirects. Otherwise links that we don't control remain broken.

Not sure about this one, these were broken links, it's not that they existed before.
We've already added redirects in all places that were changed because of the structure change.

@jegtnes Any chance you can please resolve the conflicts so I'll push this one?
Thanks! :)

PR #779, opened 2021-03-07
docs(react-testing-library): warn about afterEach auto cleanup footgun

cdaringe:
# Problem

Our tests implement an `afterEach(async () => { /* ... */ })` handler. We registered it, thinking all was well! However, jest processes afterEach handlers in reverse order of registration. That means that our integration test execute, effects could be outstanding due to interactions in the test, the test block exits, our `afterEach` _perhaps_ take a little bit of time to complete, the react effects settle from the integration test, and an `act()` error is emitted. Only after our `afterEach` promise settles does RTL `cleanup()` get invoked, prompting effect teardown/cancellation, etc.

We wrote code that makes it s.t. `act()` errors fail our tests. With the above scenario, we found our tests to have some degree of flakiness in noisy/inconsistent environments (like a busy CI agent).

The implicit behavior of RTL, I found, was actually undesirable. If I had known `cleanup` was a RTL provision, I would have more rapidly identified it as a potential culprit in our failures. Generally, side-effects as imports can be risky, with general exception when you explicitly import a verb, like `babel/register`, etc.

I suspect other community members have periodic act() errors that they consider ghosts in the system, when perhaps they really need to look at their own `afterEach()` handlers!

Let's warn those users! :)

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -19,6 +19,16 @@ require that you use Jest).
 Adding options to your global test config can simplify the setup and teardown of
 tests in individual files.
 
+## Safe cleanup
+
+`React Testing Library` will auto `cleanup` rendered components in all 
+frameworks that expose an `afterEach` API. Be warned, however, that any `async` 
+`afterEach` handlers you register that _defers_ the RTL `cleanup` call will 
+leave opporitunity for your rendered components to continue process react state
+updates and emit `act()` errors. If you implement custom async `afterEach` 
+handlers, it is recommended to call `cleanup` explicity before any async 
+behavior.
+
 ## Custom Render
 
 It's often useful to define a custom render method that includes things like


eps1lon (commented):
I'm not following how this translates to actual code.

Could you give a minimal example code of the problem and how you fixed it?

cdaringe:
Sure, I'll send one later.I have an easy repro. The tldr is,

Test a component with >1 asynchronous use effect/setState calls (very
common).

Run test, test passes.

Add an afterEach with async behavior.

Observe act errors. afterEach async calls are not protected by RTLs
internal act wrapping. RTLs auto cleanup is not invoked until after my
afterEach

cdaringe:
Consider this basic `<App />` you are integration testing. This is a very simple, MVP example app:

```tsx
// fake api call
const getData = (numRecords: number) =>
  sleep(numRecords).then(() => numRecords);

const A = () => {
  const [content, setContent] = React.useState("A is loading");
  React.useEffect(() => {
    getData(10).then(() => setContent("A is ready"));
  }, []);
  return <p>{content}</p>;
};

const B = () => {
  const [content, setContent] = React.useState("B is loading");
  React.useEffect(() => {
    getData(50).then(() => setContent("B is ready"));
  }, []);
  return <p>{content}</p>;
};

export function App() {
  return (
    <div>
      <A />
      <B />
    </div>
  );
}
```

Here's an easy test, to assert that `<A />` eventually renders per expectation.

```tsx
import * as React from "react";
import { render, screen } from "@testing-library/react";
import { App } from "../App";

test("cleanup & act error demo", async () => {
  render(<App />);
  expect(await screen.findByText("A is ready")).toBeInTheDocument();
});
```

> PASS  src/__tests__/app.spec.tsx

Everything passes. Everything is fine and serene. 🟢 👼 

Now, what if that test, in reality, needs some setup and async teardown? Here's a very real looking patch you might apply:

```patch
diff --git a/src/__tests__/app.spec.tsx b/src/__tests__/app.spec.tsx
index 4c61540..81bd140 100644
--- a/src/__tests__/app.spec.tsx
+++ b/src/__tests__/app.spec.tsx
@@ -1,6 +1,13 @@
 import * as React from "react";
 import { render, screen } from "@testing-library/react";
 import { App } from "../App";
+import { sleep } from "../timing";
+
+const simulateImportantTeardownWork = () => sleep(100);
+
+afterEach(async () => {
+  await simulateImportantTeardownWork();
+});
 
test("cleanup & act error demo", async () => {
   render(<App />);
```

What happens in the test now?

```txt
 PASS  src/__tests__/app.spec.tsx
  ● Console

    console.error
      Warning: An update to B inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at B (/Users/c0d01a5/src/react-rtl-integration-testing/src/App.tsx:17:39)
          at div
          at App

      17 |   const [content, setContent] = React.useState("B is loading");
      18 |   React.useEffect(() => {
    > 19 |     getData(50).then(() => setContent("B is ready"));
         |                            ^
      20 |   }, []);
      21 |   return <p>{content}</p>;
      22 | };

      at printWarning (node_modules/.pnpm/react-dom@17.0.1_react@17.0.1/node_modules/react-dom/cjs/react-dom.development.js:67:30)
```

What's happened? Why did we get this error now? All we did was add an afterEach!  And that in fact is problematic. 

- in the first demo, RTL `cleanup`  ran **immediately**. Jest froze the console and tore down the worker before additional work could be processed for react errors to surface
- in the second demo, the async nature of our `afterEach` allowed react to continue to process state changes, specifically _outside_ of any RTL or explicit `act()` functions. our `simulateImportantTeardownWork` afterEach occurs _before_ RTLs auto-registering afterEach `cleanup`

Thus, what seemed like a harmless `afterEach` addition, ended up being quite problematic. Even worse--this example was designed to surface the error, by ensuring that `afterEach` was sufficiently slow to surface the act() error. In many cases, async behaviors are mocked and fast. That means that this race condition can be subtle to surface for many peoples' code. `<A />` and `<B />` effects could have certainly settled in the inner `waitFor(...)`, and no one would have been the wiser.


Order matters. What happens if we register our handler, _then_ import RTL and implicitly setup the auto cleanup?

```patch
diff --git a/src/__tests__/app.spec.tsx b/src/__tests__/app.spec.tsx
index 0864a78..033a93f 100644
--- a/src/__tests__/app.spec.tsx
+++ b/src/__tests__/app.spec.tsx
@@ -1,14 +1,14 @@
-import * as React from "react";
-import { render, screen } from "@testing-library/react";
-import { App } from "../App";
-import { sleep } from "../timing";
-
 const simulateImportantTeardownWork = () => sleep(100);
 
 afterEach(async () => {
   await simulateImportantTeardownWork();
 });
 
+import * as React from "react";
+import { render, screen } from "@testing-library/react";
+import { App } from "../App";
+import { sleep } from "../timing";
+
 test("cleanup & act error demo", async () => {
   render(<App />);
   expect(await screen.findByText("A is ready")).toBeInTheDocument();
```

On test, we get:

```
 PASS  src/__tests__/app.spec.tsx
  ● Console

    console.error
      Warning: An update to B inside a test was not wrapped in act(...).
      
    ...SNIP SNIP...

    console.error
      Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

   ...SNIP SNIP...
```

This is probably the *best* possible outcome. RTL's magic cleanup happened immediately, just like in the very first example we ran. We also got to execute `simulateImportantTeardownWork`. Unlike in the first example, though, _just means of adding some timing delay_, jest did not teardown our test, and we learned that we did not appropriately clean up our effects! This could be desirable or undesirable, based on your design goals. Importantly, in this final example, at least we _learned about it_.


The objective of this PR and conversation it to help the community become aware of this subtle, but likely-more-common-than-reported, source of bugs because of 

- implicit side-effects and
- async interaction between react, jest, and RTL

eps1lon:
I personally don't like the `import` in the middle of the file. I would recommend importing from `/pure` instead, importing `cleanup` and calling that manually where you think is appropriate.

cdaringe:
Definitely. I didn't even see that pure was an option. I certainly wouldn't shift the imports--it was done only to demonstrate the significance of order and async afterEach impact

cdaringe:
Anyway, integration testing is prone to regular failure act() warnings if this isn't considered.

PR #783, opened 2021-03-10
docs: Example of testing for absence of an element

larowlan:
We were having a debate amongst some colleagues about the best way to test for absence of an element/text

There are a number of approaches so we thought perhaps it was something the docs could recommend an approach for

The approaches for labels are

```js
expect(screen.queryByLabelText("something")).toBeFalsy()
expect(screen.queryAllByLabelText("something")).toHaveLength(0)
expect(container).not.toHaveTextContent("something")
```

For text content, the equivalents would be 

```js
expect(screen.queryByText("something")).toBeFalsy()
expect(screen.queryAllByText("something")).toHaveLength(0)
expect(container).not.toHaveTextContent("something")
```

Thoughts?

### DIFF
--- a/docs/queries/about.mdx
+++ b/docs/queries/about.mdx
@@ -169,6 +169,19 @@ const container = document.querySelector('#app')
 const inputNode2 = getByLabelText(container, 'Username')
 ```
 
+Similarly, you can use a query to test the absence of an element (byLabelText, in this case):
+
+```js
+import { screen, queryByLabelText } from '@testing-library/dom'
+
+// With screen:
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @larowlan, thanks for taking the time to open this one! :)
We actually do have a section about [asserting for elements that are not present](https://testing-library.com/docs/guide-disappearance#asserting-elements-are-not-present), did you see it? I think it covers most of the parts you're aiming to solve.

larowlan:
Perfect, thanks

MatanBobi:
Thank you for the time and efforts!

larowlan:
My pleasure!

PR #791, opened 2021-04-02
Add example using props

travis5491811:
Add example using props in reference to #1 in https://github.com/testing-library/vue-testing-library/issues/216#issuecomment-812368342

### DIFF
--- a/docs/vue-testing-library/examples.mdx
+++ b/docs/vue-testing-library/examples.mdx
@@ -92,6 +92,90 @@ test('properly handles v-model', async () => {
 })
 ```
 
+## Example using `props` and [`stubs`](https://vue-test-utils.vuejs.org/api/options.html#stubs) for thrid party compoents like `bootstrap-vue`:
+
+```html
+<template>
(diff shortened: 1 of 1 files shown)


MatanBobi:
@afontcu Can you please have a look at this one? :)

afontcu:
Hi! Thank you for taking the time 🙌 

Given the fact that the guiding principle of Testing Lib is to test components in a real environement, I'm not entirely sure we should showcase a test using `stubs` as a "first" example. I would keep the test simpler, and focus only on props (which was the missing bit we identified on https://github.com/testing-library/vue-testing-library/issues/216 :)

Again, thanks for this!

travis5491811:
Welcome @afontcu, I've reduced the example down to only focus on props. I also didn't realize my editor didn't pick up my spelling mistakes on the last commit. They showed immediately when I opened the patch. It's resolved now. Is this change okay?

MatanBobi:
@afontcu is this ready to be merged or can we close it? :)

PR #794, opened 2021-04-13
Fix within link

edag94:
Fixing issue [793](https://github.com/testing-library/testing-library-docs/issues/793)


alexkrolick (approved):
Thanks

PR #808, opened 2021-04-21
Update docs to use toBeDisabled()

mattstobbs:
The examples in the docs currently use `.toHaveAttribute('disabled')`. However, jest-dom provides `toBeDisabled()`, which does the same thing but is more expressive and gives a better failure message.

### DIFF
--- a/docs/react-testing-library/example-intro.mdx
+++ b/docs/react-testing-library/example-intro.mdx
@@ -158,10 +158,12 @@ await waitFor(() =>
 ### Assert
 
 ```jsx
-// assert that the alert message is correct.
+// assert that the alert message is correct using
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @mattstobbs! Thanks for taking the time to doing this :)
Maybe we should also add a comment so people will know that this matcher is taken from `jest-dom`?

mattstobbs:
Hi @MatanBobi, thanks for the idea. I think the other function, `toHaveTextContent`, is also from jest-dom so I've added a comment for that as well. Let me know what you think! :)

MatanBobi:
Sorry it took me some time, thanks @mattstobbs!

MatanBobi:
@all-contributors please add @mattstobbs for docs :)

PR #811, opened 2021-04-22
docs: Replace deprecated api

ErfanMirzapour:
oc page: (q6)
https://testing-library.com/docs/react-testing-library/faq

Replace `wait` which is deprecated with `waitFor`.


eps1lon (approved):
Thanks!

PR #812, opened 2021-04-22
docs: Correct meaning and punctuation of sentence

ErfanMirzapour:
Doc: https://testing-library.com/docs/ecosystem-user-event/#api


eps1lon (approved):
Much better, thanks!

PR #818, opened 2021-05-05
docs: replaced url for queries page

shemExelate:
resolves https://github.com/testing-library/testing-library-docs/issues/803

### DIFF
--- a/docs/react-testing-library/migrate-from-enzyme.mdx
+++ b/docs/react-testing-library/migrate-from-enzyme.mdx
@@ -199,10 +199,10 @@ automatically cleans up the output for each test, so you don't need to call
 
 The other thing that you might notice is `getByRole` which has `heading` as its
 value. `heading` is the accessible role of the `h1` element. You can learn more
-about them on [queries](dom-testing-library/api-queries.mdx#byrole)
-documentation page. One of the things people quickly learn to love about Testing
-Library is how it encourages you to write more accessible applications (because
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
Could you also add a redirect to the new page so that links from older versions of the docs and links from external pages work as well?

shemExelate:
> Could you also add a redirect to the new page so that links from older versions of the docs and links from external pages work as well?

Correct me if I'm wrong but I think there is already a redirect for this page

https://github.com/testing-library/testing-library-docs/blob/ba1ae858d1362dc4e923f1b9915c9994dbc57655/netlify.toml#L58

eps1lon (approved):
> Correct me if I'm wrong but I think there is already a redirect for this page

@shemExelate That's confusing to me. The redirects are working. But on the currently deployed docs the link doesn't make sense at all:

In https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/#test-1-render-the-component-and-check-if-the-h1-value-is-correct "queries" links to https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/dom-testing-library/api-queries.mdx#byrole which obviously is 404.

But I guess this wasn't the original link but rather https://testing-library.com/docs/dom-testing-library/api-queries#byrole which does redirect. So at least links from externals should be good.

eps1lon:
@all-contributors add @shemExelate for docs

PR #820, opened 2021-05-05
feat: Awards section

Liadshiran:
As described in issue #753 
I could not find a more suitable image for the state of js award, so I just cropped it as it appears there. If anyone has a better suggestion I'm open to change this.


Liadshiran:
Recreated this PR with the commit from the correct git user:
https://github.com/testing-library/testing-library-docs/pull/821

PR #827, opened 2021-05-05
chore(docs): Update api-events.mdx issue #842

drorheller:
Closes https://github.com/testing-library/dom-testing-library/issues/842

### DIFF
--- a/docs/dom-testing-library/api-events.mdx
+++ b/docs/dom-testing-library/api-events.mdx
@@ -122,7 +122,8 @@ const myEvent = createEvent.click(node, { button: 2 })
 fireEvent(node, myEvent)
 // myEvent.timeStamp can be accessed just like any other properties from myEvent
 // note: The access to the events created by `createEvent` is based on the native event API,
-// Therefore, native properties of HTMLEvent object (e.g. `timeStamp`) should be set using Object.defineProperty
+// Therefore, native properties of HTMLEvent object (e.g. `timeStamp`, `cancelable`, `type`) should be set using Object.defineProperty
+// For more info see: https://developer.mozilla.org/en-US/docs/Web/API/Event
 ```
 
 You can also create generic events:
--- a/docs/dom-testing-library/api-events.mdx
+++ b/docs/dom-testing-library/api-events.mdx
@@ -121,6 +121,8 @@ programmatically (such as `timeStamp`).
 const myEvent = createEvent.click(node, { button: 2 })
 fireEvent(node, myEvent)
 // myEvent.timeStamp can be accessed just like any other properties from myEvent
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-events.mdx
```diff
@@ -121,6 +121,8 @@ programmatically (such as `timeStamp`).
 const myEvent = createEvent.click(node, { button: 2 })
 fireEvent(node, myEvent)
 // myEvent.timeStamp can be accessed just like any other properties from myEvent
+// note: The access to the events created by `createEvent` is based on the native event API,
```
MatanBobi, line 124: Maybe it would be best to add a reference to the native event API (https://developer.mozilla.org/en-US/docs/Web/API/Event).

MatanBobi (changes requested):

docs/dom-testing-library/api-events.mdx
```diff
@@ -121,6 +121,8 @@ programmatically (such as `timeStamp`).
...
+// Therefore, native properties of HTMLEvent object (e.g. `timeStamp`) should be set using Object.defineProperty
```
MatanBobi, line 125: I'd probably also add that only some properties need to be changed using `Object.defineProperty` as there are no setters in the native implementation.

docs/dom-testing-library/api-events.mdx
drorheller: @MatanBobi I think that this citation is more for users of the testing library, I'm not sure that in the native API you would even need to change the timestamp in the first place

docs/dom-testing-library/api-events.mdx
drorheller: @MatanBobi that's why I said "native" properties, but I'm honestly not exactly sure about the exact list

docs/dom-testing-library/api-events.mdx
MatanBobi: Oh yeah, I agree with you.. But since not all testing library users will be 100% familiar with the event API, I suggested to add a link to it.

docs/dom-testing-library/api-events.mdx
MatanBobi: Everything in [this page](https://developer.mozilla.org/en-US/docs/Web/API/Event) which has `Read only` on it can't be changed without using `Object.defineProperty` :)
I don't mean to add a list, maybe just some examples. As all properties are the native properties.

drorheller: @MatanBobi cool. adding

eps1lon:
@drorheller Thanks!

@all-contributors Add @drorheller for docs

PR #700, opened 2020-11-30
Adding first iteration over TSX on new tab

pedroapfilho:
This is the first iteration of the hypothesis of a new tab for `tsx` code on the examples.

https://github.com/testing-library/testing-library-docs/issues/696

PR #828, opened 2021-05-05
feat: more tsx tabs in RTL setup doc

Liadshiran:
As proposed in #696 and started with #700, I added the rest of the code snippets examples in Typescript for the Setup doc of react testing library.


pedroapfilho:
Could you check these conflicts @Liadshiran

docs/react-testing-library/setup.mdx
```diff
@@ -79,6 +79,7 @@ export { customRender as render }
 
 ```tsx title="test-utils.tsx"
 import React, { FC, ReactElement } from 'react'
+import { Omit } from 'react-redux'
```
MatanBobi, line 82: I'm not sure about the `Omit` here.. I think that the options type should be different as it's the options for our `render` function.

MatanBobi (changes requested):

MatanBobi:
Thanks for taking the time @Liadshiran!
I left one comment :)

MatanBobi:
And also, [here](https://github.com/testing-library/testing-library-docs/pull/828/files#diff-9a8513ec98fb23892fc832ebf723cc61ff3844a81e7b37f2a52d997426bb898cR100)
I think we should also omit `wrapper`..

Liadshiran:
I added `wrapper` and removed `queries` which seemed redundant from the Omit type

MatanBobi (approved):

MatanBobi:
Thanks @Liadshiran and @pedroapfilho! :)

PR #834, opened 2021-05-05
docs: add docs for intl components 

itsik-avidan:


itsik-avidan:
adding documentation following discussion from 
https://github.com/testing-library/testing-library-docs/issues/737

nickmccurdy:
This is good information/copy, but are there any specific approaches we'd want to recommend for most cases?

alexkrolick:
Let's start here since it's been a long-standing issue to document some intl approaces. Probably will want to add some example code for the more complex methods later. I don't know that there is an obvious one-size-fits all option except "use the default strings" right now, and even that has trade-offs.

alexkrolick (approved):

PR #839, opened 2021-05-06
docs(user-event): Remove strange argument passing

yialo:
If I correclty understand `screen.getByRole()` method signature, we shouldn't pass here two string arguments.
And, according to [WAI-ARIA spec](https://w3c.github.io/html-aria/), there is no `email` role.


eps1lon (approved):
Thanks :+1:

eps1lon:
@all-contributors add @yialo for docs

PR #848, opened 2021-05-15
Corrects the location pointed to by broken footer "API" link. Fixes #…

### LINKED ISSUE #847: Footer link to API points at wrong uri
**Describe the bug**
The Footer component has a link to the "API" which points to `docs/dom-testing-library/api-queries` (https://testing-library.com/docs/dom-testing-library/api-queries). The page does not exist and clicking the link redirects to 404 page.

**To Reproduce**
1. Navigate to https://testing-library.com/docs/
2. Scroll to the footer at the bottom of the page
3. Locate and click the "API" link under the "Docs" subheader.

**Expected behavior**
Directs user to `/docs/dom-testing-library/api` (https://testing-library.com/docs/dom-testing-library/api)

Should be applicable on all browsers, OSs, and versions, but for good measure...
 - OS: MacOS Big Sur (11.2.3)
 - Browser: Firefox Developer Edition
 - Version: 89.0b11

vxxce:
Corrects broken footer link to "API". 
Resolves #847

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -283,7 +283,7 @@ module.exports = {
             },
             {
               label: 'API',
-              to: 'docs/dom-testing-library/api-queries',
+              to: 'docs/dom-testing-library/api',
             },
             {
               label: 'Help',


eps1lon (approved):
Thanks! Could you also add a redirect so that existing links outside of our control to https://testing-library.com/docs/dom-testing-library/api-queries still work?

MatanBobi:
@eps1lon I believe we already have a redirect for that page:
https://github.com/testing-library/testing-library-docs/blob/main/netlify.toml#L58

If you agree, I think we can merge this! :)
Thanks @vxxce!

eps1lon:
> @eps1lon I believe we already have a redirect for that page:
https://github.com/testing-library/testing-library-docs/blob/main/netlify.toml#L58

Right, I always get confused that redirects do not apply to on-page navigation. Only when landing :confused:

vxxce:
Looks like there is an existing re-direct from `docs/dom-testing-library/api-queries` to `/docs/queries/about`:
https://github.com/testing-library/testing-library-docs/blob/b81803272d4ecbc1d0688d6a603b052d4cf86557/netlify.toml#L57-L59

I do wonder if it might be better to redirect the "api-queries" endpoint to `docs/dom-testing-library/api` (as in e7279f6) rather than `docs/queries/about`. I.e:

```suggestion
[[redirects]]
  from = "/docs/dom-testing-library/api-queries"
  to = "/docs/dom-testing-library/api"
```

Given the possible ambiguity in the url, it makes sense to me to link them to the core-api page, from which there is a direct link to queries if that was the context in which it was linked to externally. Kind of catches all possible intentions. I'm happy to commit that suggestion if you there's agreement about that. 

And if I'm missing something re: that re-direct already existing, let me know! Thanks.

MatanBobi:
> I think we should go through every redirect and check if a link in the docs applies.
@eps1lon I went over the redirects we had there and found one broken, I opened [a PR](https://github.com/testing-library/testing-library-docs/pull/856)

MatanBobi:
Thanks @vxxce!

MatanBobi:
@all-contributors please add @vxxce for docs :)

PR #850, opened 2021-05-18
Add .keyboard method to user-event page

### LINKED ISSUE #845: Add keyboard method to the user-event page
**The problem**
I had to use the user-event [`.keyboard()`](https://github.com/testing-library/user-event#keyboardtext-options) API method to simulate a keyboard event, but I found that it's not part of the official [doc](https://testing-library.com/docs/ecosystem-user-event), even though it was released some time ago.

**The solution**
It would be nice to have the user-event keyboard API method being part of the official documentation website.

**Additional context**
I'm more than happy to fork the repo and add it to the [page](https://github.com/testing-library/testing-library-docs/blob/main/docs/ecosystem-user-event.mdx), but I just wanted to double-check first if this actually makes sense before jumping in.

luizbaldi:
Closes #845.

_The content was simply copied from the [user-event keyboard readme](https://github.com/testing-library/user-event/blob/master/README.md#keyboardtext-options), but let me know if changes are needed._

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -30,8 +30,9 @@ test('types inside textarea', () => {
 ## API
 
 Note: All userEvent methods are synchronous with one exception: when `delay`
-option used with `userEvent.type` as described below. We also discourage using `userEvent`
-inside `before/after` blocks at all, for important reasons described in
+option used with `userEvent.type` as described below. We also discourage using
+`userEvent` inside `before/after` blocks at all, for important reasons described
+in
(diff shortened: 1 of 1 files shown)


nickmccurdy:
I think we should figure out how to deduplicate this content eventually, but for now it's better to have this updated.

nickmccurdy (approved):
I think we should figure out how to deduplicate this content eventually, but for now it's better to have this updated.

eps1lon:
@all-contributors add @luizbaldi for docs

nickmccurdy:
> I think we should eventually figure out how to deduplicate this content, but for now it's better to have this updated (and it could potentially make migration easier).

If you'd like updates on this, see https://github.com/testing-library/testing-library-docs/issues/519#issuecomment-882126733

PR #867, opened 2021-06-20
docs: add note about including dom types in WTL

olivierwilkinson:
WebdriverIO Testing Library assumes that users include the DOM types in
their typescript default types. Add a note to the Typescript section of
the WTL docs to help users who are finding type errors when they haven't.

Related issue: https://github.com/testing-library/webdriverio-testing-library/issues/22


eps1lon:
Thanks! 

@all-contributors add @olivierwilkinson for docs

PR #872, opened 2021-06-26
Updated Formik testing example, added a missing parameter

HouCoder:
[This PR](https://github.com/testing-library/testing-library-docs/pull/692) forgot to add the second onSubmit parameter in the example code, which causes the testing to always fail.

```
  ● rendering and submitting a basic Formik form

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

      {"email": "john.dee@someemail.com", "firstName": "John", "lastName": "Dee"},
    - Anything,

    Number of calls: 1
```

### DIFF
--- a/docs/example-formik.md
+++ b/docs/example-formik.md
@@ -14,9 +14,9 @@ import { Formik, Field, Form } from 'formik'
 const sleep = ms => new Promise(r => setTimeout(r, ms))
 
 export const MyForm = ({ onSubmit }) => {
-  const handleSubmit = async values => {
+  const handleSubmit = async (values, formikBag) => {
     await sleep(500)
-    onSubmit(values)
+    onSubmit(values, formikBag)
   }
 
   return (


docs/example-formik.md
```diff
@@ -14,9 +14,9 @@ import { Formik, Field, Form } from 'formik'
 const sleep = ms => new Promise(r => setTimeout(r, ms))
 
 export const MyForm = ({ onSubmit }) => {
-  const handleSubmit = async values => {
+  const handleSubmit = async (values, formikBag) => {
```
eps1lon, line 17: For future reference, could you point to forkik documentation for this? I couldn't find it.

HouCoder: Do you mean the Formik doc for the onSubmit callback? If so then this is the doc - https://formik.org/docs/api/formik#onsubmit-values-values-formikbag-formikbag--void--promiseany

Thanks.

eps1lon: We just need to revert https://github.com/testing-library/testing-library-docs/pull/692.

You probably don't want to expose the formik bag anyway.

PR #873, opened 2021-06-26
Update api-async.mdx

Notr0M:
typo


MatanBobi:
Hi @Notr0M! Thanks for taking the time to open this one.
I actually don't see this as a typo so I'm ok with keeping it as it is.
Thanks again for the initiative and for taking the time :)

PR #881, opened 2021-06-30
docs(vue): update faq for Vue Router abstract mode

blimmer:
This update is in response to a conversation between @maxarndt and I [here](https://github.com/testing-library/vue-testing-library/issues/210#issuecomment-871208563). The workaround of resetting `window.location` has some pitfalls, so this PR updates the documentation to only recommend using `abstract` mode.

### DIFF
--- a/docs/vue-testing-library/faq.mdx
+++ b/docs/vue-testing-library/faq.mdx
@@ -90,40 +90,62 @@ environment in between test invocations, so route transitions from previous
 tests can leak into subsequent tests, even though a new Vue Router is created
 with each call to `render`.
 
-You can work around this in one of two ways:
-
-1. **Pass an instantiated router using `abstract` mode**. `abstract` mode does
-   not store route information on `window`, so transitions will not leak between
-   tests. For example:
+To work around this issue, pass an instantiated router using `abstract` mode.
+`abstract` mode does not store route information on the JSDOM `window`, so
+routing information will not leak between tests. For example:
 
 ```js
-import { render, fireEvent } from '@testing-library/vue'
+import { render } from '@testing-library/vue'
 import Component from './Component.vue'
 import VueRouter from 'vue-router'
 
-test('uses abstract mode for the router', () => {
-  render(Component, {
-    routes: new VueRouter({
-      mode: 'abstract',
-      routes: [
-        // Your routes here
-      ],
-    }),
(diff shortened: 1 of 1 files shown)


docs/vue-testing-library/faq.mdx
```diff
@@ -90,40 +90,62 @@ environment in between test invocations, so route transitions from previous
...
+    routes: router,
   })
+
+  // Unlike the router in `hash` mode, the initial routing stack is empty. So,
+  // you need to push an initial route to the stack.
+  await router.push('/')
 })
 ```
 
-2. **Reset the window location `afterEach`**. If you don't want to pass an
```
blimmer, line 116: This isn't the best option for people - they should instead use `abstract` mode, so don't recommend it.

docs/vue-testing-library/faq.mdx
```diff
@@ -90,40 +90,62 @@ environment in between test invocations, so route transitions from previous
...
+    ],
+  })
+
+  const renderResult = render(Component, {
+    routes: router,
   })
+
+  // Unlike the router in `hash` mode, the initial routing stack is empty. So,
+  // you need to push an initial route to the stack.
+  await router.push('/')
```
blimmer, line 116: cc @maxarndt - this adds the initial `push` you mentioned in https://github.com/testing-library/vue-testing-library/issues/210#issuecomment-871208563

docs/vue-testing-library/faq.mdx
```diff
@@ -90,40 +90,62 @@ environment in between test invocations, so route transitions from previous
...
 })
 ```
 
-2. **Reset the window location `afterEach`**. If you don't want to pass an
-   instantiated Router, you can instead reset the `window.location` after each
-   test, like this:
+To reduce boilerplate, you can create a custom render function to use throughout
```
blimmer, line 120: @maxarndt - if you've got some time for a contribution, it would be great to pull over the Custom Render setup guide from the react-testing-library docs (https://testing-library.com/docs/react-testing-library/setup#custom-render) to the currently blank setup page of Vue Testing Library (https://testing-library.com/docs/vue-testing-library/setup). Then, we could link to the custom render docs from this FAQ.

afontcu (approved):
LGTM! We can work on the custom render section in a upcoming PR.


thanks and nice catch!

PR #882, opened 2021-06-30
Updates disappearance examples to use screen. queries

fildon:
It is generally recommended to make queries via the screen object. This PR updates some examples on the "Appearance and Disappearance" page to stick to that best practice.

### DIFF
--- a/docs/guide-disappearance.mdx
+++ b/docs/guide-disappearance.mdx
@@ -20,7 +20,7 @@ when calling them._
 test('movie title appears', async () => {
   // element is initially not present...
   // wait for appearance and return the element
-  const movie = await findByText('the lion king')
+  const movie = await screen.findByText('the lion king')
 })
 ```
 
@@ -32,7 +32,7 @@ test('movie title appears', async () => {
 
   // wait for appearance inside an assertion
   await waitFor(() => {
-    expect(getByText('the lion king')).toBeInTheDocument()
+    expect(screen.getByText('the lion king')).toBeInTheDocument()
   })
 })
 ```
@@ -46,7 +46,7 @@ when the element is removed.
 ```jsx
 test('movie title no longer present in DOM', async () => {
   // element is removed
-  await waitForElementToBeRemoved(() => queryByText('the mummy'))
+  await waitForElementToBeRemoved(() => screen.queryByText('the mummy'))
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide-disappearance.mdx
```diff
@@ -32,7 +32,7 @@ test('movie title appears', async () => {
 
   // wait for appearance inside an assertion
   await waitFor(() => {
-    expect(getByText('the lion king')).toBeInTheDocument()
+    expect(screen.getByText('the lion king')).toBeInTheDocument()
```
timdeschryver, line 35: We encourage the usage of `findBy` over this, so should this example be added to the guide?
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

docs/guide-disappearance.mdx
```diff
@@ -64,7 +64,7 @@ test('movie title goes away', async () => {
   // note use of queryBy instead of getBy to return null
   // instead of throwing in the query itself
   await waitFor(() => {
-    expect(queryByText('i, robot')).not.toBeInTheDocument()
+    expect(screen.queryByText('i, robot')).not.toBeInTheDocument()
```
timdeschryver, line 67: Same here.

alexkrolick: findBy would throw an error here because the element is being asserted as _not_ present. The previous query also uses queryBy for symmetry in the example.

timdeschryver: Oh right, good catch. Thanks for pointing this out 👍

docs/guide-disappearance.mdx
timdeschryver: Hey @fildon, do you want to update the guide by removing this section (`2. Using waitFor`) please?

fildon: Can do, but then I don't think this section of the docs flows well any more. 

This example is number 2 in a pair of examples. The first example demonstrates "findByText" and this one demonstrates "waitFor". If we want to change this one to use "findByText" then we are just duplicating the two examples effectively. 

Perhaps the second example should be removed entirely? And then this part of the docs shouldn't be a numbered list (since then it would have only one entry)?

timdeschryver: Good catch, I think that's a good proposition 👍

alexkrolick: Maybe this example is too simple, but there are still reasons to use waitFor in a positive case, such as batching together several element queries in one retry block or adding in additional assertions after the getBy that also need to rerun (ie, checking element properties are in the right state).

timdeschryver: So you would leave it like this, or should it have a comment/more complex example?

PR #887, opened 2021-07-16
Add note about Mocha's watch mode

jankalfus:
The docs currently say that auto cleanup is run automatically if `afterEach` global is present. That is true, however, that doesn't work in Mocha's watch mode, as it executes only the changed files. Globals are not re-run in watch mode. Figuring this out took me an hour or two, so I'm proposing to add info about it to the docs, as I believe it can be helpful for others.

Also, I'm thinking I could add a mocha-specific hook to testing-library/react, similarly to `@testing-library/react/dont-cleanup-after-each`, but that might not be desired, so I think this change to the docs should probably be good enough for now.

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -368,3 +368,29 @@ mocha -r @testing-library/react/dont-cleanup-after-each
 Alternatively, you could import `@testing-library/react/pure` in all your tests
 that you don't want the `cleanup` to run and the `afterEach` won't be setup
 automatically.
+
+### Auto Cleanup in Mocha's watch mode
+
+When using Mocha in watch mode, the globally registered cleanup is run only the
+first time after each test. Therefore, subsequent runs will most likely fail
+with a _TestingLibraryElementError: Found multiple elements_ error.
+
+To enable automatic cleanup in Mocha's watch mode, add a cleanup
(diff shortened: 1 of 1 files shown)


nickmccurdy (approved):

nickmccurdy:
This looks good, but would it possibly be simpler if we exposed a hook like this in Testing Library?

jankalfus:
It might be :) Then on the other hand, this might make folks see what's going on and why it's needed. Anyway, I'll open up a PR with a hook in Testing Library, link it to this one, and we can decide then.

jankalfus:
Created a PR with a hook in Testing Library: https://github.com/testing-library/react-testing-library/pull/945

eps1lon (changes requested):
Much appreciated. We don't want to extend the library with support for more test runners at the moment. A documentation only change is sufficient for now.

jankalfus:
That was my initial thought as well. Reverted.

eps1lon (approved):

eps1lon:
Thanks!

@all-contributors add @jankalfus for docs

tassoevan:
It wasn't immediately obvious for me that this advice was at the "Setup" section because I've read about `cleanup` in the "API" one. As there is a remark about it being automatically called in some testing frameworks, could it be useful to link this note as a small observation?

PR #888, opened 2021-07-17
Add peer dependency to User Event docs

nickmccurdy:
User Event peer depends on DOM Testing Library, so it needs to be manually installed, as stated in the readme.

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -12,7 +12,7 @@ advanced simulation of browser interactions than the built-in
 ## Install
 
 ```
-npm install --save-dev @testing-library/user-event
+npm install --save-dev @testing-library/user-event @testing-library/dom
 ```
 
 ```jsx


MatanBobi (approved):
This looks good for now, but I'm trying to understand why only `user-event` defines a peerDep unlike all the other packages that have `testing-library/dom` as a dependency.

nickmccurdy:
There's rationale here, but I'm not sure why it's better than just putting the `>=` version range in `dependencies`.

@kentcdodds @ph-fritsche Do you have any thoughts on this? I think it's a bit surprising that User Event still uses peer dependencies, while framework libraries like React Testing Library use regular dependencies. 

Anyway I will merge this for now, since it's currently still a peer dependency, and the docs should be up to date with the readme.

ph-fritsche:
I think that another copy of `@testing-library/dom` might cause problems with the config.
When importing e.g. `@testing-library/react` it automatically replaces the `actWrapper` and `eventWrapper`.

If `@testing-library/dom` imports inside `@testing-library/userEvent` resolves to another copy, it would be one with default config.

kentcdodds:
The reason is pretty simple actually.

People using `@testing-library/react` don't need to think about using or installing `@testing-library/dom` because it's a wrapper, but the people using `@testing-library/user-event` need to have `@testing-library/dom` installed (either directly or transitively via `@testing-library/react`) because it's a companion utility, not a wrapper.

nickmccurdy:
Reverting this in #1329.

PR #890, opened 2021-07-18
add: react testing library course by techbase

MitchelSt:
I recently released a free React Testing Library course on YouTube. It might be a good starting point for people that prefer watching video's over reading docs.


eps1lon:
Thanks but we keep the current course list for now. Otherwise this will become a place people want to run advertisement on.

PR #784, opened 2021-03-17
feat: convert npm to yarn commands on docs

semoal:
Hey folks 👋

I've introduced this little plugin to introduce automatic conversion of npm commands to yarn.

### Demo:
Kapture 2021-03-17 at 14 55 52

PR #892, opened 2021-07-20
docs: update docs for user-event from readme

nickmccurdy:


docs/ecosystem-user-event.mdx
```diff
@@ -7,24 +7,28 @@ title: user-event
 advanced simulation of browser interactions than the built-in
 [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) method.
 
-[`user-event` on GitHub][gh]
+## Installation
 
-## Install
+With NPM:
```
MatanBobi, line 12: Maybe we can use a `Tab` for this one? :)

MatanBobi (approved):

docs/ecosystem-user-event.mdx
nickmccurdy: What do you think about using #784 for that? It was closed, but personally I think it would be an easier solution.

MatanBobi: I agree that #784 has a nice and easy solution but I'm not sure we really need a third party package for something we usually do once in a long time.
Anyways, I don't have a strong opinion to either way so we can push that PR also if you think it's worth it.

nickmccurdy:
I've also thought about splitting up the Installation and API into separate packages like the main packages. Is this something we could do in the ecosystem section, and if so is it worth trying for user-event?

MatanBobi (approved):
Looks good, thanks!

About splitting the page to two pages, I believe it's worth doing, especially if we're planning to port the whole documentation for `user-event` into the docs site (Which I think is the right approach). I'm starting to think that `user-event` should have it's own place since even though it's a part of the eco-system, it's a pretty important package for us.

nickmccurdy:
I'm going to merge this now, since the user-event readme already links here and we don't want the docs to be out of date. However, I'll probably try to split this up in another PR soon.

MatanBobi:
> I'm going to merge this now, since the `user-event` readme already links here and we don't want the docs to be out of date. However, I'll probably try to split this up in another PR soon.
> 
> > I'm starting to think that `user-event` should have it's own place since even though it's a part of the eco-system, it's a pretty important package for us.
> 
> How would this be represented in the sidebar?

@nickmccurdy It can be the same as the frameworks section, I'm imagining something like this:
image

PR #894, opened 2021-07-20
docs: use `kcd-scripts` Prettier config

nickmccurdy:
`kcd-scripts` is used by almost all Testing Library packages, and it includes a Prettier config by default. For consistent formatting between source files, readmes, and docs, the docs repository should use the same Prettier config. I achieved this by simply installing `kcd-scripts` and re-exporting its Prettier config, so we don't need to change our build config.


alexkrolick:
Are you sure you want proseWrap? That's causing most of the reformatting

eps1lon (approved):
Makes sense 👍🏻

PR #900, opened 2021-08-04
Fixes findByText usage in Cypress example test

sfrieson:
It seems that chained `.findByText()` API was used incorrectly incorrectly. Running the example on the Cypress page with the HTML below results in this error:
image

These changes allow the test to succeed.

I specifically ran into this because I tried copy/pasting this example and altering it for my site and I was confused as to why it wasn't working.

___

Example HTML
```html
<!DOCTYPE html>
<html>
<body>
  <button>Jackie Chan</button>
  <form>
    <label for="input">Label text</label>
    <input type="text" id="input">
    <button type="button">Button Text</button>
  </form>
  <dialog open>
    <p>Do you want to read the docs?</p>
    <button>Confirm</button>
  </dialog>
</body>
</html>
```

### DIFF
--- a/docs/cypress-testing-library/intro.mdx
+++ b/docs/cypress-testing-library/intro.mdx
@@ -58,7 +58,7 @@ cy.findByRole('button', {name: /Button Text/i}).should('exist')
 cy.findByRole('button', {name: /Non-existing Button Text/i}).should('not.exist')
 cy.findByLabelText(/Label text/i, {timeout: 7000}).should('exist')
 
-// findAllByText _inside_ a form element
+// findByText _inside_ a form element
 cy.get('form')
   .findByText(/Button Text/i)
   .should('exist')
--- a/docs/cypress-testing-library/intro.mdx
+++ b/docs/cypress-testing-library/intro.mdx
@@ -60,7 +60,7 @@ cy.findByLabelText(/Label text/i, {timeout: 7000}).should('exist')
 
 // findByText _inside_ a form element
 cy.get('form')
-  .findByText(/Button Text/i)
+  .findByRole('button', {name: /Button Text/i})
   .should('exist')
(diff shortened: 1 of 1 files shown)


docs/cypress-testing-library/intro.mdx
```diff
@@ -58,9 +58,9 @@ cy.findByRole('button', {name: /Button Text/i}).should('exist')
 cy.findByRole('button', {name: /Non-existing Button Text/i}).should('not.exist')
 cy.findByLabelText(/Label text/i, {timeout: 7000}).should('exist')
 
-// findAllByText _inside_ a form element
+// findByText _inside_ a form element
 cy.get('form')
-  .findByText('button', {name: /Button Text/i})
+  .findByText(/Button Text/i)
```
timdeschryver, line 63: Have you tried using `getByRole`?
Reason: `getByRole` is the recommended selector to query elements, https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time

```suggestion
  .findByRole('button', {name: /Button Text/i})
```

sfrieson: Thanks for the suggestion of using `findByRole`. This is what I would generally use in this situation as well, but in this PR I was not trying to make any judgement on the correctness of the selector but rather to use the idea that was already there appropriately.

I was assuming that, since the example already uses `findByRole` multiple times (you can see them if you expand the diff), the example author wanted to show a different API and used `findByText`. I guess it could alternatively be seen as wanting to use `findByRole` again because the arguments match that.

I am fine with either version and just want the example to be valid. Since it seems better for the example to follow bet practices I will switch to `findByRole`.

timdeschryver (approved):

eps1lon:
@all-contributors add @sfrieson for docs

PR #901, opened 2021-08-05
Make summary table details open by default.

dwjohnston:
It's very easy to miss the summary table - which is like the most useful part of the page. Have it open by default.


eps1lon (approved):
Thanks!

eps1lon:
@all-contributors add @dwjohnston for docs

PR #911, opened 2021-08-10
docs: fix broken React Native Testing Library links

nickmccurdy:
The old link `https://www.native-testing-library.com/` seems to have been taken over by a malicious website. I replaced links with the new RNTL docs and Brandon's GitHub (his Twitter is deleted, which means it would also be taken over by a malicious account).


nickmccurdy:
I'm just going to merge this for security/safety purposes, but if anyone wants me to update this, you can leave a review and I'll open another PR.

PR #929, opened 2021-09-09
Update info about multiple elements queries.

BenHurMartins:
There is a max number os elements when queries have matching nodes.

### DIFF
--- a/docs/queries/about.mdx
+++ b/docs/queries/about.mdx
@@ -71,6 +71,11 @@ test('should show login form', () => {
       [`waitFor`](../dom-testing-library/api-async.mdx#waitfor). They accept the
       `waitFor` options as the last argument (i.e.
       `await screen.findByText('text', queryOptions, waitForOptions)`)
+> **Note**
+>
+> All multiple elements queries with matching nodes will return an Array with a 
+> max of 10 elements.
+
 
 <details open>
 


docs/queries/about.mdx
```diff
@@ -71,6 +71,11 @@ test('should show login form', () => {
       [`waitFor`](../dom-testing-library/api-async.mdx#waitfor). They accept the
       `waitFor` options as the last argument (i.e.
       `await screen.findByText('text', queryOptions, waitForOptions)`)
+> **Note**
+>
+> All multiple elements queries with matching nodes will return an Array with a
```
alexkrolick, line 76: Do they? They should not

BenHurMartins: Yea, at least I could realize that after a lot of tests

eps1lon: @benhurmartins Could you share a repro? That sounds like a bug to me.

BenHurMartins: @eps1lon , yes. the repo is:
https://github.com/BenHurMartins/react-native-test-library-showcase
It is on the src/screens/List/List.test.tsx file. I am testing with a real api, firing an event of scrolling, then finding more elements that will load after the Flatlist load more data, each time it should load 20 elements, but the findAll will always return 10. I tried simulating a FLatlist with 10+ elements, but it still returning 10, so I thought that was the default behavior.

eps1lon: > It is on the src/screens/List/List.test.tsx file. I am testing with a real api, firing an event of scrolling, then finding more elements that will load after the Flatlist load more data, each time it should load 20 elements, but the findAll will always return 10. I tried simulating a FLatlist with 10+ elements, but it still returning 10, so I thought that was the default behavior.

I suspect that the list implementation you use, implements some virtualization behavior and therefore only rendering 10 elements.

And the docs sound like it is actually virtualized:

> This is a convenience wrapper around <VirtualizedList>, and thus inherits its props (as well as those of <ScrollView>) that aren't explicitly listed here, along with the following caveats:

-- https://reactnative.dev/docs/flatlist

Try it with a basic list implementation and check how many list items that returns.

BenHurMartins: Alright, I will do a test with Text elements perhaps, outside of a list structure and using findAll, u thought that was a undocumented feature.

MatanBobi:
@BenHurMartins Since this doesn't seem related to `testing-library` I'm closing it.
If you believe this should be re-opened please comment here and I'll do that.
Thanks for taking the time to work on this PR :)

PR #930, opened 2021-09-09
pecies.mdx: add React Testing Library React Dnd Chessboard Example

ryota-murakami:
When I found this blank Examples page with coincidences I thought I could add content from my repos.  
Is this addition welcome?

Let me know if there is something to should fix or improve, thank you!

### DIFF
--- a/docs/recipes.mdx
+++ b/docs/recipes.mdx
@@ -5,3 +5,5 @@ title: About the Examples
 
 [Examples](./example-codesandbox) and [Help Resources](./learning) are
 contributed by the Testing Library community
+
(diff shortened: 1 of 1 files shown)


timdeschryver (commented):
Hi, thanks for adding a resource.
However, I don't think this is the place to add the example.
I think https://testing-library.com/docs/example-external would be better suited for it.

ryota-murakami:
> I think https://testing-library.com/docs/example-external would be better suited for it.

@timdeschryver Thank you for tell me about it! I'll submit PR to that page!

PR #940, opened 2021-09-23
Update incorrect statement about jsdom in Jest

asztal:
[Jest 27 has been released](https://jestjs.io/blog/2021/05/25/jest-27#feature-updates), which no longer defaults to `jest-environment-jsdom`.


alexkrolick:
👍

PR #944, opened 2021-09-29
Fix typo in guide-disappearance.mdx

stanleyugwu:


docs/guide-disappearance.mdx
```diff
@@ -69,7 +69,7 @@ test('movie title goes away', async () => {
 })
 ```
 
-## Asserting elements are not present
+## Asserting elements that are not present
```
timdeschryver, line 72: I think "Asserting elements are not present" is more correct here, but I'm not an English native speaker 😅
Why do we need "that" here?

How I see it:
- Asserting elements are not present: you're asserting an element isn't present, e.g. you test it isn't there
- Asserting elements that are not present: you're asserting on an element that isn't present, e.g. you that that the element that isn't there isn't there (if that makes sense)

alexkrolick: `Asserting elements are not present` or `Asserting that elements are not present` are both OK, but not this change

stanleyugwu: OH! it just made sense when I substituted 'asserting' with 'checking' 🤣😂. 
I was thinking it means "checking elements that are not present (the result)"now I see its "checking **to know** if elements are not present".

stanleyugwu: @alexkrolick thanks, your comment made it much clearer. And I think changing it to `Asserting that elements are not present` will be much more easy to understand.

stanleyugwu: Thanks @timdeschryver for that explanation

PR #945, opened 2021-10-01
docs: remove Open Collective link

nickmccurdy:
We're no longer using Open Collective, so I don't think it makes sense for us to keep the link around.

I'm also planning to focus more time on the docs soon, if so I may PR my own sponsor profile link if others are okay with it (of course we can also add links for other maintainers).

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,0 @@
-open_collective: testing-library


eps1lon (approved):

nickmccurdy:
I still see this badge to set up Sponsors on issues. Maybe it's just because of my repository permissions. I'm curious what others see. 
image

MatanBobi:
@nickmccurdy I see it too but on my other user it's not there and neither on incognito so I guess we're good :)

PR #946, opened 2021-10-06
docs(ByAltText): Add notes about support for custom elements

eps1lon:
Docs for https://github.com/testing-library/dom-testing-library/pull/1049

### DIFF
--- a/docs/queries/byalttext.mdx
+++ b/docs/queries/byalttext.mdx
@@ -23,7 +23,7 @@ getByAltText(
 ```
 
 This will return the element (normally an `<img>`) that has the given `alt`
-text. Note that it only supports elements which accept an `alt` attribute:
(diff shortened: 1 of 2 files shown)


SantoJambit:
about.mdx also needs to be changed.

eps1lon:
@SantoJambit Thanks for the hint. This works now?

docs/queries/byalttext.mdx
```diff
@@ -23,7 +23,7 @@ getByAltText(
 ```
 
 This will return the element (normally an `<img>`) that has the given `alt`
-text. Note that it only supports elements which accept an `alt` attribute:
+text. Note that it only supports elements which accept an `alt` attribute or [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) (since we can't know if a custom elements implements `alt` or not):
```
timdeschryver, line 26:
```suggestion
text. Note that it only supports elements which accept an `alt` attribute or [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) (since we don't know if a custom element implements `alt` or not):
```

timdeschryver (approved):

PR #955, opened 2021-10-21
doc: use getBy* in waitFor

ipovos:
`expect(queryByText('the lion king')).toBeInTheDocument()` throws an error when element with the text 'the lion king' is not found in the document.
`getByText('the lion king')` throws an error in the same case.
We can simplify expression
```
await waitFor(() => {
  expect(getByText('the lion king')).toBeInTheDocument()
})
```
to
```
await waitFor(() => getByText('the lion king'))
```
as it has the same meaning.

### DIFF
--- a/docs/guide-disappearance.mdx
+++ b/docs/guide-disappearance.mdx
@@ -31,9 +31,7 @@ test('movie title appears', async () => {
   // element is initially not present...
 
   // wait for appearance inside an assertion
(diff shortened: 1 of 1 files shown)


MatanBobi:
I agree with @eps1lon on this one.
I'm closing this as I don't think it's worth merging.
Thanks for taking the time to work on this one @ipovos!

PR #966, opened 2021-11-18
chore: update algolia credentials

MatanBobi:
Since there's no migration guide for docusaurus v2 yet, I read the [v1 migration guide](https://v1.docusaurus.io/blog/2021/11/05/algolia-migration) and updated according to it and according to [algolia's recommendations](https://docsearch.algolia.com/docs/migrating-from-legacy/).

I tested this one locally and search seems to work properly.

Resolves https://github.com/testing-library/testing-library-docs/issues/965

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -344,7 +344,8 @@ module.exports = {
       },
     },
     algolia: {
-      apiKey: 'bda29e6557dc5be1ce5c05f2dbff8f33',
+      appId: 'TUPO88CFRP',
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
LGTM 👍

Is this all that needs to be done?

MatanBobi:
As far as I was able to see, it looks like it is the only thing we need to do on our end (once we received the email):
image

(That's from algolia's migration guide)

I also tested the PR's preview and search is working good.

kentcdodds:
Let's do it!

PR #980, opened 2022-01-05
feat: document user-event v14

ph-fritsche:
Add documentation for `user-event@14` alongside `v13` for rest of `beta`.

New projects / projects that someone is actively working on should probably already switch to `v14-0.0-beta`, but I want to wait for more feedback before declaring this `stable`.
I hope the new documentation helps with this.

### DIFF
--- a/docs/user-event/api-clipboard.mdx
+++ b/docs/user-event/api-clipboard.mdx
@@ -1,0 +1,49 @@
+---
+id: clipboard
+title: Clipboard
+---
+
+Note that the
+[Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) is
+usually not available outside of secure context.  
+To enable testing of workflows involving the clipboard,
+[`userEvent.setup()`](setup) replaces `window.navigator.clipboard` with a stub.
+
+## copy()
+
+```ts
+copy(): Promise<DataTransfer|undefined>
+```
+
+Copy the current selection.
+
+If [`writeToClipboard`](options#writetoclipboard) is `true`, this will also
+write the data to the `Clipboard`.
+
+## cut()
+
+```ts
+cut(): Promise<DataTransfer|undefined>
+```
+
+Cut the current selection.
+
+If [`writeToClipboard`](options#writetoclipboard) is `true`, this will also
+write the data to the `Clipboard`.
+
+When performed in editable context, it removes the selected content from the
+document.
+
+## paste()
+
+```ts
+paste(clipboardData?: DataTransfer|string): Promise<void>
+```
+
+Paste data into the document.
+
+When called without `clipboardData`, the content to be pasted is read from the
+`Clipboard`.
+
+When performed in editable context, the pasted content is inserted into to the
+document.
--- a/docs/user-event/api-convenience.mdx
+++ b/docs/user-event/api-convenience.mdx
@@ -1,0 +1,83 @@
+---
+id: convenience
+title: Convenience APIs
+---
+
+The following APIs are shortcuts for equivalent calls to the underlying
+[`pointer()`](pointer) and [`keyboard()`](keyboard) APIs.
+
+## Clicks
+
+### click()
+
+```ts
+click(element: Element): Promise<void>
+```
+
+```js
+pointer([{target: element}, {keys: '[MouseLeft]', target: element}])
+```
+
+The first action might be skipped per [`skipHover`](options#skiphover).
+
+### dblClick()
+
+```ts
+dblClick(element: Element): Promise<void>
+```
+
+```js
+pointer([{target: element}, {keys: '[MouseLeft][MouseLeft]', target: element}])
+```
+
+### tripleClick()
+
+```ts
+tripleClick(element: Element): Promise<void>
+```
+
+```js
+pointer([
+  {target: element},
+  {keys: '[MouseLeft][MouseLeft][MouseLeft]', target: element},
+])
+```
+
+## Mouse movement
+
+### hover()
+
+```ts
+hover(element: Element): Promise<void>
+```
+
+```js
+pointer({target: element})
+```
+
+### unhover()
+
+```ts
+hover(element: Element): Promise<void>
+```
+
+```js
+pointer({target: element.ownerDocument.body})
+```
+
+## Keyboard
+
+### tab()
+
+```ts
+tab(options: {shift?: boolean})
+```
+
+```js
+// without shift
+keyboard('{Tab}')
+// with shift=true
+keyboard('{Shift>}{Tab}{/Shift}')
+// with shift=false
+keyboard('[/ShiftLeft][/ShiftRight]{Tab}')
+```
--- a/docs/user-event/api-keyboard.mdx
+++ b/docs/user-event/api-keyboard.mdx
@@ -79,5 +79,5 @@ keys.
 > Future versions might try to interpolate the modifiers needed to reach a
 > printable key on the keyboard. E.g. Automatically pressing `{Shift}` when
 > CapsLock is not active and `A` is referenced. If you don't wish this behavior,
-> you can deactivate the [`autoModify`](options#automodify) option to opt of out
+> you can deactivate the [`autoModify`](options#automodify) option to opt out
 > of this non-breaking change.
--- a/docs/user-event/api-keyboard.mdx
+++ b/docs/user-event/api-keyboard.mdx
@@ -1,0 +1,83 @@
+---
+id: keyboard
+title: Keyboard
+---
+
+```ts
+keyboard(input: KeyboardInput): Promise<void>
+```
+
+The `keyboard` API allows to simulate interactions with a keyboard. It accepts a
+`string` describing the key actions.
+
+Keystrokes can be described:
+
+- Per printable character
+  ```js
+  keyboard('foo') // translates to: f, o, o
+  ```
+  The brackets `{` and `[` are used as special character and can be referenced
+  by doubling them.
+  ```js
+  keyboard('{{a[[') // translates to: {, a, [
+  ```
+- Per
+  [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)
+
+  ```js
+  keyboard('{Shift}{f}{o}{o}') // translates to: Shift, f, o, o
+  ```
+
+  This does not keep any key pressed. So `Shift` will be lifted before pressing
+  `f`.
+
+  Characters with special meaning inside the key descriptor can be escaped by
+  prefixing them with a backslash `\`.
+
+  ```js
+  keyboard('{}}') // translates to: }
+  ```
+
+- Per
+  [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
+  ```js
+  keyboard('[ShiftLeft][KeyF][KeyO][KeyO]') // translates to: Shift, f, o, o
+  ```
+
+Keys can be kept pressed by adding a `>` to the end of the descriptor.  
+If this should result in repeated `keydown` events, you can add the number of
+repetitions.  
+If the key should also be released after this, add a slash `/` to the end of the
+descriptor.
+
+```js
+keyboard('{a>}') // press a without releasing it
+keyboard('{a>5}') // press a without releasing it and trigger 5 keydown
+keyboard('{a>5/}') // press a for 5 keydown and then release it
+```
+
+A previously pressed key can be lifted by prefixing the descriptor with `/`.
+
+```js
+keyboard('{/a}') // release a previously pressed a
+```
+
+This allows to simulate key combinations.
+
+```js
+keyboard('{Shift>}A{/Shift}') // translates to: Shift(down), A, Shift(up)
+```
+
+The mapping of `key` to `code` is performed by a
+[default key map](https://github.com/testing-library/user-event/blob/beta/src/keyboard/keyMap.ts)
+portraying a "default" US-keyboard. You can provide your own local keyboard
+mapping per [`keyboardMap`](options#keyboardMap) option.
+
+Currently the different `key` meanings of single keys are treated as different
+keys.
+
+> Future versions might try to interpolate the modifiers needed to reach a
+> printable key on the keyboard. E.g. Automatically pressing `{Shift}` when
+> CapsLock is not active and `A` is referenced. If you don't wish this behavior,
+> you can deactivate the [`autoModify`](options#autoModify) option to opt of out
+> of this non-breaking change.
--- a/docs/user-event/api-keyboard.mdx
+++ b/docs/user-event/api-keyboard.mdx
@@ -16,7 +16,7 @@ Keystrokes can be described:
   ```js
   keyboard('foo') // translates to: f, o, o
   ```
-  The brackets `{` and `[` are used as special character and can be referenced
+  The brackets `{` and `[` are used as special characters and can be referenced
   by doubling them.
   ```js
   keyboard('{{a[[') // translates to: {, a, [
--- a/docs/user-event/api-keyboard.mdx
+++ b/docs/user-event/api-keyboard.mdx
@@ -71,7 +71,7 @@ keyboard('{Shift>}A{/Shift}') // translates to: Shift(down), A, Shift(up)
 The mapping of `key` to `code` is performed by a
 [default key map](https://github.com/testing-library/user-event/blob/beta/src/keyboard/keyMap.ts)
 portraying a "default" US-keyboard. You can provide your own local keyboard
-mapping per [`keyboardMap`](options#keyboardMap) option.
+mapping per [`keyboardMap`](options#keyboardmap) option.
 
 Currently the different `key` meanings of single keys are treated as different
 keys.
@@ -79,5 +79,5 @@ keys.
 > Future versions might try to interpolate the modifiers needed to reach a
 > printable key on the keyboard. E.g. Automatically pressing `{Shift}` when
 > CapsLock is not active and `A` is referenced. If you don't wish this behavior,
-> you can deactivate the [`autoModify`](options#autoModify) option to opt of out
+> you can deactivate the [`autoModify`](options#automodify) option to opt of out
 > of this non-breaking change.
--- a/docs/user-event/api-pointer.mdx
+++ b/docs/user-event/api-pointer.mdx
@@ -14,9 +14,7 @@ accepts a single pointer action or an array of them.
 type PointerInput = PointerActionInput | Array<PointerActionInput>
 ```
 
-## No-layout environment
-
-Our primary target audience tests per `jest` in a `jsdom` environment and there
+> Our primary target audience tests per `jest` in a `jsdom` environment and there
 is no layout in `jsdom`. This means that different from your browser the
 elements don't exist in a specific position, layer and size.  
 We don't try to determine if the pointer action you describe is possible at that
--- a/docs/user-event/api-pointer.mdx
+++ b/docs/user-event/api-pointer.mdx
@@ -1,0 +1,122 @@
+---
+id: pointer
+title: Pointer
+---
+
+```ts
+pointer(input: PointerInput): Promise<void>
+```
+
+The `pointer` API allows to simulate interactions with pointer devices. It
+accepts a single pointer action or an array of them.
+
+```ts
+type PointerInput = PointerActionInput | Array<PointerActionInput>
+```
+
+## No-layout environment
+
+Our primary target audience tests per `jest` in a `jsdom` environment and there
+is no layout in `jsdom`. This means that different from your browser the
+elements don't exist in a specific position, layer and size.  
+We don't try to determine if the pointer action you describe is possible at that
+position in your layout.
+
+## Pointer action
+
+There are to types of actions: press and move.
+
+### <a name="press" href="#"/>Pressing a button or touching the screen
+
+A pointer action is a press action if it defines a key to be pressed, to be
+released, or both.
+
+```js
+pointer({keys: '[MouseLeft]'})
+```
+
+You can declare multiple press actions (on the same position) at once which will
+be resolved to multiple actions internally. If you don't need to declare any
+other properties you can also just supply the `keys` string.
+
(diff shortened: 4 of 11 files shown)


MatanBobi:
Hi @ph-fritsche! Thanks for tagging me to review this one :)
I'll have a look at it later on today but just to raise a discussion: 
`user-event` is an important library in the testing-library family, I think it should get a bigger place in the docs and not just in the eco-system section, especially given the amount of docs you've just added.
Any thoughts?

ph-fritsche:
You won't see me arguing against a more prominent advertisement for `user-event`.

I think the change between `v12.8.3` which still has >500k npm downloads per week and `v14` is huge - not only if you measure it in code lines changed.
There is currently no bug issue left and the more fundamental approach that was necessary to implement some features (and some bug fixes) without endless adding of conditionals also made the codebase more maintainable.
I'd say `user-event@14` is a very mature tool that I would recommend anywhere.

docs/user-event/options.mdx
```diff
@@ -0,0 +1,118 @@
+---
+id: options
+title: Options
```
timdeschryver, line 3: What isn't clear to me are what these options are.
Are these the options for `setup`? 
If that's the case, what about adding them to the `setup` page or to make it explicit that these are render options in an intro text?

timdeschryver (commented):
Nice! 🤩

docs/user-event/install.mdx
```diff
@@ -0,0 +1,36 @@
...
+import TabItem from '@theme/TabItem'
+
+<Tabs defaultValue="npm" values={[
+    { label: 'npm', value: 'npm' },
+    { label: 'Yarn', value: 'yarn' },
+]}>
+<TabItem value="npm">
+
+```sh
+npm install --save-dev @testing-library/user-event@^14.0.0-beta
```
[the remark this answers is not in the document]
timdeschryver: Ah, never mind v14 is "still" in beta.
I was under the impression that v14 was already released because of "I'd say user-event@14 is a very mature tool that I would recommend anywhere."

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+be resolved to multiple actions internally. If you don't need to declare any
+other properties you can also just supply the `keys` string.
+
+```js
+pointer({keys: '[MouseLeft][MouseRight]'})
+// or
+pointer('[MouseLeft][MouseRight]')
+```
+
+For pressing a button without releasing it, the button name is suffixed with
```
MatanBobi, line 46:
```suggestion
In order to press a button without releasing it, the button name is suffixed with
```

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+`>`.  
+For just releasing a previously pressed button, the tag is started with `/`.
+
+```js
+pointer('[MouseLeft>]') // press the left mouse button
+pointer('[/MouseLeft]') // release the left mouse button
+```
+
+Which button are available depends on the [`pointerMap`](options#pointermap).
```
MatanBobi, line 55:
```suggestion
Which buttons are available depends on the [`pointerMap`](options#pointermap).
```

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+### <a name="move" href="#"/>Moving a pointer
+
+Every pointer action that is not a press action describes a pointer movement.
+
+You can declare which pointer is moved per `pointerName` property. This defaults
+to `mouse`.
+
+Note that the `mouse` pointer (`pointerId: 1`) is also the only pointer that
+always exists and has a position. A `touch` pointer only exists while the screen
+is touched and receives a new `pointerId` every time. For these we therefore use
```
MatanBobi, line 66:
```suggestion
is touched and receives a new `pointerId` every time. For these use cases, we use
```

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+}
+```
+
+Pointer actions can alter the selection in the document.  
+In the browser every pointer position corresponds with a DOM position. This is a
+DOM node and a DOM offset which usually translates to the character closest to
+the pointer position.  
+As all character in a no-layout environment are in the same layout position we
+assume a pointer position to be closest to the last descendant of the pointer
+`target`.
```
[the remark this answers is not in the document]
MatanBobi: I guess that it's related to the fact that JSDOM has no layout mechanism so all elements have the same position?

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+If you provide `offset`, we assume the pointer position to be closest to the
```
MatanBobi, line 116: It seems to me like an example will be good here to make our users understand it properly.

docs/user-event/api-utility.mdx
```diff
@@ -0,0 +1,179 @@
+---
+id: utility
+title: Utility APIs
+---
+
+The following APIs don't have one-to-one equivalents in a real user
+interaction.  
+Their behavior is therefore an interpretation how the "perceived" user
+interaction might be translated to actual events on the DOM.
```
MatanBobi, line 9: I'm not sure, but maybe this is better?

```suggestion
This means that their behavior is our own interpretation of how the browser behaves and the eventing mechanism should not be taken as a source of truth.
```

docs/user-event/api-utility.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+## clear()
+
+```ts
+clear(element: Element): Promise<void>
+```
+
+This API can be used to easily clear an editable element.
+
+1. Focus element
```
MatanBobi, line 19:
```suggestion
The events we initiate are:
1. Focus element
```

docs/user-event/api-utility.mdx
```diff
@@ -0,0 +1,179 @@
...
+1. Select all contents as per browser menu
+1. Delete contents as per browser menu
+
+```jsx
+test('clear', async () => {
+  render(<textarea value="Hello, World!" />)
+
```
MatanBobi, line 26: I think we can reduce the line breaks here :)

docs/user-event/install.mdx
MatanBobi, line 16: Since we're already tagging this release as beta, don't we want to use that version tag?

```suggestion
npm install --save-dev @testing-library/user-event@beta
```

docs/user-event/install.mdx
```diff
@@ -0,0 +1,36 @@
...
+```
+
+</TabItem>
+<TabItem value="yarn">
+
+```sh
+yarn add --dev @testing-library/user-event@^14.0.0-beta
```
MatanBobi, line 23: Same here

```suggestion
yarn add --dev @testing-library/user-event@beta
```

docs/user-event/install.mdx
```diff
@@ -0,0 +1,36 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+Note that `@testing-library/user-event` requires `@testing-library/dom`.
+
+If you use one of the
+[framework wrappers](../dom-testing-library/install#wrappers), it is important
```
MatanBobi, line 32: I'm not sure if this comment will confuse people or help. Did you see this happening a lot?

docs/user-event/intro.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+We recommend to use [`userEvent.setup()`](setup) when rendering your component
+and inline that rendering and setup in your test or use a setup function.  
+We discourage rendering or using any `userEvent` functions outside of the test
+itself - e.g. in a `before`/`after` hook - for reasons described in
+["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
+
+```js
+test('trigger some awesome feature when clicking the button', async () => {
+    const user = userEvent.setup()
```
MatanBobi, line 39: This is only my opinion, but is there a way to `setup` userEvent in the `setupFilesAfterEnv` in jest for example?
I think causing our users to write this line in every test might cause some pain.

MatanBobi (commented):
This is amazing!
I left a few comments :)

docs/user-event/api-pointer.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+```ts
+type PointerInput = PointerActionInput | Array<PointerActionInput>
+```
+
+> Our primary target audience tests per `jest` in a `jsdom` environment and there
+is no layout in `jsdom`. This means that different from your browser the
+elements don't exist in a specific position, layer and size.  
+We don't try to determine if the pointer action you describe is possible at that
+position in your layout.
```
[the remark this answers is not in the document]
ph-fritsche: No this literally means that there is no specific position, layer, or size. Elements can have IDL properties and/or CSS styles that usually correspond to elements being at a specific position, having specific dimensions, or being above/under each other. But `jsdom` has no layout. The properties and styles exist, but they have no effect.

```js
document.body.innerHTML=`<img style="position: relative; left: 50px;"/>`
$("img").style.left // '50px'
$("img").offsetLeft // 0
```

docs/user-event/api-pointer.mdx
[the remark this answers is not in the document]
ph-fritsche: Why should we add a comma here? :thinking: 

> As all character in a no-layout environment are in the same layout position

There is no layout. So every character has *effective* width zero and is at *effective* offset `0,0`. (While there might be properties applied that suggest otherwise but have no effect.)

docs/user-event/api-utility.mdx
```diff
@@ -0,0 +1,179 @@
...
+The `Promise` is rejected if the element can not be focused or contents can not
+be selected.
+
+## <a name="selectoptions"/><a name="deselectoptions"/> selectOptions(), deselectOptions()
+
+```ts
+selectOptions(
+    element: Element,
+    values: HTMLElement | HTMLElement[] | string[] | string,
```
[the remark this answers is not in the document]
ph-fritsche: If you file a PR for this, it could as well accept a RegExp.
To be blunt: I don't like the "Utility" APIs at all. I think they should be out of scope because there is no "correct" way to implement them. You can not open a browser and record what happens if you perform the action declared by the API call because the API does not unambiguously declare a user interaction.
But I'm content with separating them as APIs that follow a slightly different approach than the rest of the library.

sidebars.js
```diff
@@ -172,6 +172,19 @@ module.exports = {
     {
       Ecosystem: [
         'ecosystem-user-event',
+        {
```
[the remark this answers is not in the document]
ph-fritsche: This might fall into line with the suggestion by @MatanBobi to grant `user-event` a more prominent spot in the navigation.

docs/user-event/options.mdx
ph-fritsche: I think one of the biggest mistakes of the previous docs was to mention the same options at different locations and splatter them across the docs.

If you know that you should use `setup` and how you should use `pointer` and `keyboard`, I think the other APIs are pretty much self-explanatory.
But one can use the library for months without ever needing any option and the options might also be extended in the future. So I think a prominent place in the docs is a better choice than nesting them under `setup`. (Also IMHO the docs design works much better with short pages.)

Intro text to make this more clear :+1:

docs/user-event/api-keyboard.mdx
```diff
@@ -0,0 +1,83 @@
...
+The `keyboard` API allows to simulate interactions with a keyboard. It accepts a
+`string` describing the key actions.
+
+Keystrokes can be described:
+
+- Per printable character
+  ```js
+  keyboard('foo') // translates to: f, o, o
+  ```
+  The brackets `{` and `[` are used as special character and can be referenced
```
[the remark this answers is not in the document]
ph-fritsche: `characters` :+1: 

I'm not sure. I mean technically (and this is the best kind of correct ;) ) you are also *using* the character if you don't double it - it just has a special meaning and probably results in an error. ^^

docs/user-event/api-clipboard.mdx
```diff
@@ -0,0 +1,49 @@
...
+title: Clipboard
+---
+
+Note that the
+[Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) is
+usually not available outside of secure context.  
+To enable testing of workflows involving the clipboard,
+[`userEvent.setup()`](setup) replaces `window.navigator.clipboard` with a stub.
+
+## copy()
```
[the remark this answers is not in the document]
ph-fritsche: I copied that style with backticks from the old docs at first. But especially at `h2` I perceive the non-code variant as far more readable and a more pleasing page layout - especially when excluding the parameters and adding them in the code blocks below. I think it makes for faster navigation when scrolling.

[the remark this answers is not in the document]
MatanBobi: Cool, I'm with you :)

docs/user-event/api-convenience.mdx
```diff
@@ -0,0 +1,83 @@
+---
+id: convenience
+title: Convenience APIs
+---
+
+The following APIs are shortcuts for equivalent calls to the underlying
```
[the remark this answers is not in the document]
ph-fritsche: Yeah, I don't know. On the one hand they exist because they are easy to use and so - yes - they are probably used often, but on the other hand they are self-explanatory and merely documented for sake of completeness. The auto-completion in IDE would probably be enough.

Maybe this could be solved by another blog post walking through some examples with heavy use of these convenience methods and mentioning the post with a short excerpt in the introduction.
A mention in `pointer` and `keyboard` might also be fine.

docs/user-event/api-utility.mdx
ph-fritsche: I think that would be misleading. The problem is not the browser behavior, it's the user interaction.
There is not *one* specific user interaction that is described by those APIs. So we can not reproduce the *one* correct behavior of the API. So we roll with what *some* (or *most*?) users think that user interaction would be and then implement the correct behavior for this.
But in reality there are either different input methods to reach the described interaction or there simply is no actual interaction like described by the API and perceived by our users.

docs/user-event/api-utility.mdx
ph-fritsche: I'm not sure about this. `events` would be misleading I think. This should describe the steps the API performs as the steps performed by these APIs have not been clear to our users in the past.

docs/user-event/api-utility.mdx
ph-fritsche: imho the line breaks between arrange-act-assert make it more readable and maybe make the test patterns more easily digestible for newcomers because it makes it easier to spot the same pattern in short as well as in longer tests.

(White space is cheap. No reason to be stingy with it. ^^)

docs/user-event/install.mdx
ph-fritsche: I saw it quite a few times and I fear it might have happened a lot more.
Also those who ran into this had no clue why this could be the source of their problem.

After all it is not general knowledge that e.g. `@testing-library/react` when imported has the side-effect to change behavior in `@testing-library/dom`. (I think others frameworks might do similar things.)

But previously the async wrapper was used sparely because most people didn't use the `delay` option.
Now all APIs are asynchronous.

docs/user-event/intro.mdx
ph-fritsche: Yes, this could be done in a `beforeEach`. But we especially mention [Avoid nesting when you're testing](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) in the docs...

The solution for now would be:
```js
function setup(jsx) {
  render(jsx)
}

test('something', () => {
  const { user } = setup(<MyComponent/>)
  // ...
})
```

This pattern might even become *that* common that we should consider returning the user-event instance from `render` in the framework packages.

docs/user-event/api-keyboard.mdx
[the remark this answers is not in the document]
ph-fritsche:
```suggestion
  The brackets `{` and `[` are used as special characters and can be referenced
```

docs/user-event/api-pointer.mdx
ph-fritsche:
```suggestion
is touched and receives a new `pointerId` every time. For these pointers, we use
```

docs/user-event/install.mdx
ph-fritsche: :shrug: I don't know. Doesn't make a difference.
IMHO in case of doubt, I'd go with the more explicit one as it once again highlights the major version bump. We probably won't use the `beta` channel for a while after is is released.

docs/user-event/intro.mdx
ph-fritsche: Added an example for a test setup function below.

docs/user-event/api-utility.mdx
```diff
@@ -0,0 +1,179 @@
...
+    options?: {
+        skipClick?: boolean
+        skipAutoClose?: boolean
+        initialSelectionStart?: number
+        initialSelectionEnd?: number
+    }
+): Promise<void>
+```
+
+Type into an input element.
```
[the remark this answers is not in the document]
ph-fritsche: Added an example below.

docs/user-event/install.mdx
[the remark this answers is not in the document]
ph-fritsche: I added a blockquote note in the `Introduction`.

sidebars.js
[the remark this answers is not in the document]
ph-fritsche: Elevated it in the navigation.

docs/user-event/options.mdx
ph-fritsche: Added an intro text.

docs/user-event/api-pointer.mdx
ph-fritsche: Added an example. I hope this makes it clear enough. Might be worth explaining Layout Positions and DOM offsets and how the browser handles selection per pointer in a blog post, but this would break the mold of docs specific to the library.

docs/user-event/api-pointer.mdx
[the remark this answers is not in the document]
timdeschryver: 👍 I see thanks for the elaboration, the term "no-layout" wasn't clear to me.
As for the comma, it's to make it easier to read imho. It's also something that grammarly (a spelling tool) suggests.

docs/user-event/api-pointer.mdx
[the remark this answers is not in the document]
timdeschryver: 👍 Got it, the comment made it click to me - would it be worth it to also mention it in the docs?

> every character has effective width zero and is at effective offset 0,0. (While there might be properties applied that suggest otherwise but have no effect.)

docs/user-event/api-utility.mdx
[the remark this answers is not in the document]
timdeschryver: Seems fair to me.

docs/user-event/api-pointer.mdx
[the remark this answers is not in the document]
ph-fritsche: I don't know. I feel like it might be better to write a blog post about this topic explaining the concept and then adding a link. Really explaining this, and what difference it makes to testing might be worth it, but is out of scope for these docs.

docs/user-event/api-pointer.mdx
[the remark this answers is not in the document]
ph-fritsche: I'm no linguistic expert :shrug: 
If I understood some articles about this correctly, generally an introductory phrase should have a comma. But it can be omitted for prepositional phrases shorter than five words and should be omitted if there is no natural pause for building suspension and/or the following sentence could not stand on its own.
I'd judge `Every pointer position corresponds with a DOM position.` on its own as technically incorrect, as this paragraph just explains that this is not the case in the no-layout environment. Following this, I think it should be there, but again, I'm no expert on this. (Which might just have been highlighted by the fact that muscle-memory typed `export` instead of `expert` twice in this post :laughing: )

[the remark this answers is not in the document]
timdeschryver: Yea, me neither 😅
I just added it because I had a corresponding question that went with it.
Thanks again @ph-fritsche !

docs/user-event/api-convenience.mdx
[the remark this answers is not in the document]
MatanBobi: Not sure we should rely on the IDE auto-completion for this one and IMO the package's functionality should be easily found in the docs and not in a personal blog post (but we should also think of writing a blog post about user-event 14 in the testing-library blog).

I strongly prefer a mention in `pointer` and `keyboard` pages :)

docs/user-event/api-utility.mdx
MatanBobi: No worries, this sounds like a good reason :)

docs/user-event/intro.mdx
MatanBobi: I'm in favor of returning the user-event instance from `render`. 
But this means people will need to install user-event too and the frameworks will instantiate that for them?
That sounds too "voodoo" to me.

docs/user-event/api-utility.mdx
MatanBobi: Yeah I understand what you're saying here. :)

docs/user-event/api-utility.mdx
MatanBobi: So maybe we should write:
```suggestion
This API mimics a "clear" action by running these steps:
1. Focus element
```

MatanBobi (approved):
This looks good enough for me to push forwards and maybe iterate later if you want :)

docs/user-event/intro.mdx
ph-fritsche: Yes, something like that.

The framework wrapper could `try..require` `user-event`. So that if `user-event` is installed the wrapper will automatically return an instance from `render`.

Or we could take a similar approach like how `@testing-library/react` configures `/dom` when it's imported.
`user-event` could register itself on the framework wrapper when being imported.

timdeschryver (approved):
I've browsed through it for a second time, and it's also looking good to me 👍

PR #981, opened 2022-01-06
Update Router props based on react-router-dom v6

mohetti:
Since react-router-dom v6, there's no history object on Router anymore. Instead, the objects navigator and location needs to be used. Reference:
https://stackoverflow.com/questions/69859509/cannot-read-properties-of-undefined-reading-pathname-when-testing-pages-in

### DIFF
--- a/docs/example-react-router.mdx
+++ b/docs/example-react-router.mdx
@@ -159,7 +159,7 @@ test('rendering a component that uses useLocation', () => {
 })
 ```
 
-## React Router v5 and Testing Library
+## Testing Library and React Router v5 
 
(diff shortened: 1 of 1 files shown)


MatanBobi:
Thanks for this one @mohetti! :)

mohetti:
I added some info text aswell as a short example for v5 users. I hope it's ok as is. Feel free to change it though.

MatanBobi:
Thanks for this one @mohetti! :)

MatanBobi:
@all-contributos please add @mohetti for docs :)

MatanBobi:
@all-contributors please add @mohetti for docs :)

PR #982, opened 2022-01-10
feat: add top-level redirect for `user-event`

ph-fritsche:
Add a top-level redirect for sharing an URL that is a) shorter and b) independent from possible future reorganizing of docs pages.


alexkrolick:
FYI - that wasn't the intent of the redirects. In the first version of the docs, each library used to have a top-level homepage with logos, use cases, etc. The redirects handle the deprecation of those pages.

PR #985, opened 2022-01-14
fix: Correct user-event v13/v14 cross-links.

make-github-pseudonymous-again:


timdeschryver (changes requested):
Could you explain why these changes are required?
It seems to me that the deployed version works, but that the preview version (with these changes) result in 404s.

image

make-github-pseudonymous-again:
Mmh. The behavior may indeed depend on whether a final backslash is added to URLs.

MatanBobi:
@make-github-pseudonymous-again can you please try to add the file extension (`.mdx`) and see if this is resolved for you? (instead of the relative url)

timdeschryver:
I think it must be something else then the browser though, the current version is also working on Firefox on my machine.

make-github-pseudonymous-again:
@timdeschryver Maybe a browser add-on/plugin that rewrites URLs? I tried to disable *HTTPS Everywhere* but that did not change anything. Then I tried chromium incognito (without add-ons) which also adds the trailing slash. I am running Arch Linux with the following packages:

    firefox 96.0-1
    chromium 97.0.4692.71-2

If it cannot be a plugin, could it be something outside the browser? Any ideas?

@MatanBobi Replacing the trailing slash in https://testing-library.com/docs/ecosystem-user-event/ by `.mdx` to obtain https://testing-library.com/docs/ecosystem-user-event.mdx redirects me to the (page not found) https://testing-library.com/docs/dom-testing-library/ecosystem-user-event.mdx. Tested both Firefox and Chromium.

If I go to https://testing-library.com/docs/ecosystem-user-event/ the href to v14-beta is

```html
<a href="/docs/ecosystem-user-event/user-event/intro">...</a>
```

**PS**: Ha sorry you meant adding `.mdx` in `href` inside the sources, I will try that later.

timdeschryver (commented):
Before we take action, I would like to see if it's already solved in the latest version of docusaurus.
Can you reproduce this problem in https://deploy-preview-986--testing-library.netlify.app/ (v2.0.0-beta.14)?

make-github-pseudonymous-again:
> Before we take action, I would like to see if it's already solved in the latest version of docusaurus. Can you reproduce this problem in https://deploy-preview-986--testing-library.netlify.app/ (v2.0.0-beta.14)?

Yes. If I search for `user-event` and click on the result for the v13 page for instance then click on the v14 link I get a "Page Not Found" error.

timdeschryver:
> @MatanBobi Indeed adding .mdx to the links in the sources produces links that work even when the current URL has a trailing slash.

> I think it's the search feature that has trailing slashes in result URLs while regular links have no trailing slashes in the URLs.

> Do you want me to edit this PR or will you fix this yourself, maybe in a more generic way?


If you want, please fix the links by appending the link with the `.mdx` extension.
As pointed out in #996, this also allows us to navigate between the docs within the GitHub interface.

make-github-pseudonymous-again:
> If you want, please fix the links by appending the link with the `.mdx` extension.

Done!

PR #987, opened 2022-01-17
Add vue3 quickstart instruction

Tobertet:
If you don't install using @next and you are using Vue3 you get an error regarding vue-template-compiler not being found.


afontcu (approved):
thanks!

PR #993, opened 2022-01-21
docs(user-event): fix code example of `clear` API

nimaebra:
Fix code example of [`clear`](https://testing-library.com/docs/ecosystem-user-event#clearelement) API in **user-events v13** docs.

The current code example does not work and causes confusion because React value prop without an onChange handler causes the field to be read-only. Also `value` attribute is `null`.

So I changed two things:
- `value` to `defaultValue` prop 
-  `toHaveAttribute('value', '')` to `toHaveValue('')`

The example code for `clear` in [Utility APIs ](https://testing-library.com/docs/user-event/utility#clear) is also following this method.

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -368,10 +368,10 @@ import {render, screen} from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
 test('clear', () => {
-  render(<textarea value="Hello, World!" />)
+  render(<textarea defaultValue="Hello, World!" />)
 
   userEvent.clear(screen.getByRole('textbox'))
-  expect(screen.getByRole('textbox')).toHaveAttribute('value', '')
+  expect(screen.getByRole('textbox')).toHaveValue('')
 })
 ```
 


MatanBobi:
Hi @nimaebra, thanks for taking the time to open this one :)
I'm actually not sure the `textarea` element supports `defaultValue` based on the [HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)

nimaebra:
@MatanBobi, Thanks for your response 💯 
The code example is written in react and according to the [react docs](https://reactjs.org/docs/dom-elements.html#value), `textarea` supports `defaultValue`.
Additionally, as I mentioned in my quote, the `clear` code example of [Utility APIs](https://testing-library.com/docs/user-event/utility#clear) is also uses `defaultValue` for `textarea`.

ph-fritsche (approved):
Thanks for taking the time to fix this :heart:

MatanBobi:
Thanks! :)
@all-contributors please add @nimaebra for docs :)

MatanBobi:
@all-contributors please add @nimaebra for docs

Sorry for the noise, didn't want to manually merge the changes :)

MatanBobi:
@all-contributors please add @ph-fritsche for review :)

PR #994, opened 2022-01-25
fix(queries): use role `img` instead of `image`

Dennis273:
Use `img` instead of `image` since `image` is not a valid aria-role. 

> For example `` will be returned for both `getByAltText('fancy image') and getByRole('image', { name: 'fancy image' })`.


MatanBobi (approved):
Good catch, thanks @Dennis273!

PR #1019, opened 2022-02-28
docs: fix fireEvent Example code

pppp606:
When I copied the code example, there was an error, so I fixed it.
https://testing-library.com/docs/preact-testing-library/api/#example-1

- Fix syntax error
- Delete undefined prop
  - https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

### DIFF
--- a/docs/preact-testing-library/api.mdx
+++ b/docs/preact-testing-library/api.mdx
@@ -113,12 +113,13 @@ fireEvent.click(buttonNode);
 
 // Fire event Option 2.
 fireEvent(
-buttonNode,
-new Event('MouseEvent', {
-  bubbles: true,
-  cancelable: true,
-  button: 0,
-});
+  buttonNode,
+  new Event('MouseEvent', {
+    bubbles: true,
+    cancelable: true,
+    button: 0,
+  })
+);
 
 expect(buttonNode).toHaveTextContent('1');
 expect(cb).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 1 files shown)


docs/preact-testing-library/api.mdx
```diff
@@ -113,12 +113,12 @@ fireEvent.click(buttonNode);
...
-buttonNode,
-new Event('MouseEvent', {
-  bubbles: true,
-  cancelable: true,
-  button: 0,
-});
+  buttonNode,
+  new Event('MouseEvent', {
+    bubbles: true,
+    cancelable: true,
```
timdeschryver, line 119: Thanks for the fix, though I would like to re-add the `button` property.

```suggestion
    cancelable: true,
    button: 0
```

`button` is a valid event property - https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button - and is also impolemented like this in fireEvent - https://github.com/testing-library/dom-testing-library/blob/main/src/event-map.js#L82.

ph-fritsche: @pppp606 is correct. There is no `button` property on `Event`. It is implemented on `MouseEvent`.
https://codesandbox.io/s/event-property-button-p6w8xh?file=/src/index.test.js

ph-fritsche, line 119:
```suggestion
  new MouseEvent('click', {
    bubbles: true,
    cancelable: true,
    button: 0,
```

pppp606:
@timdeschryver @ph-fritsche 
Thanks to both of you. I have committed the suggestion, please recheck pr.

timdeschryver (approved):

timdeschryver:
@all-contributors please add @pppp606 for docs

PR #1020, opened 2022-02-28
Clarify compatible versions for Vue 2 and Vue 3

lukahartwig:
The `latest` tag of `@testing-library/vue` points to a version that is incompatible with Vue 2. This is not reflected in the documentation so I added the correct versions to the example command.


afontcu (approved):

afontcu:
great, thanks!

PR #1023, opened 2022-03-10
docs: keyPress warning

Owrlean:
To stress out the aspect of not firing the `keyPress` event not having specified the charCode property.
https://github.com/testing-library/react-testing-library/issues/269#issuecomment-455854112

### DIFF
--- a/docs/dom-testing-library/api-events.mdx
+++ b/docs/dom-testing-library/api-events.mdx
@@ -96,6 +96,9 @@ fireEvent.keyDown(domNode, {key: 'Enter', code: 'Enter', charCode: 13})
 fireEvent.keyDown(domNode, {key: 'A', code: 'KeyA'})
 ```
 
+> :warning: Be careful to provide the `charCode` property in case of firing the
+> `keyPress` event. Otherwise, it won't be fired.
+
 You can find out which key code to use at
 [https://keycode.info/](https://keycode.info/).
 
--- a/docs/dom-testing-library/api-events.mdx
+++ b/docs/dom-testing-library/api-events.mdx
@@ -96,9 +96,8 @@ fireEvent.keyDown(domNode, {key: 'Enter', code: 'Enter', charCode: 13})
 fireEvent.keyDown(domNode, {key: 'A', code: 'KeyA'})
 ```
 
-> :warning: Be careful to provide the `charCode` property in case of firing the
-> `keyPress` event. Otherwise, it won't be fired.
-
+> :warning: `React` doesn't dispatch a `SyntheticEvent` for a native `keypress` event without `event.charCode === 13 || event.charCode >= 32`.
+> I.e. `onKeyPress` handlers will only be called if you include an appropriate `charCode` property.
 You can find out which key code to use at
 [https://keycode.info/](https://keycode.info/).
 


docs/dom-testing-library/api-events.mdx
```diff
@@ -96,6 +96,9 @@ fireEvent.keyDown(domNode, {key: 'Enter', code: 'Enter', charCode: 13})
 fireEvent.keyDown(domNode, {key: 'A', code: 'KeyA'})
 ```
 
+> :warning: Be careful to provide the `charCode` property in case of firing the
+> `keyPress` event. Otherwise, it won't be fired.
+
```
ph-fritsche, line 101: This isn't true. The native `keypress` event will be dispatched. React just won't call any `onKeyPress` handlers.
See https://github.com/testing-library/user-event/issues/766

Owrlean: Thanks for the response. 
Does it make sense to rephrase that namely the `React` does not fire the event callback?

ph-fritsche: Yes, maybe something along...
```suggestion
> :warning: `React` doesn't dispatch a `SyntheticEvent` for a native `keypress` event without `event.charCode === 13 || event.charCode >= 32`.
> I.e. `onKeyPress` handlers will only be called if you include an appropriate `charCode` property.
```

ph-fritsche: Note, that this is framework-specific and almost all use cases for this should use `user-event` instead.

docs/dom-testing-library/api-events.mdx
```diff
@@ -96,6 +96,8 @@ fireEvent.keyDown(domNode, {key: 'Enter', code: 'Enter', charCode: 13})
 fireEvent.keyDown(domNode, {key: 'A', code: 'KeyA'})
 ```
 
+> :warning: `React` doesn't dispatch a `SyntheticEvent` for a native `keypress` event without `event.charCode === 13 || event.charCode >= 32`.
```
timdeschryver, line 99: If I'm not mistaken, this would be the first time that we mention something React specific into the core docs. From my point of view, I think we should keep the core docs framework agnostic, or else we can add more and more cases like these into the docs, which would make them harder to read.
What if we move this to the React FAQ or [cheatsheet](https://testing-library.com/docs/react-testing-library/cheatsheet#events)?
Thoughts @MatanBobi ?

ph-fritsche: From the *resolved* conversation above:
> Note, that this is framework-specific and almost all use cases for this should use `user-event` instead.

timdeschryver: Yea, you're correct @ph-fritsche , but then maybe this comment shouldn't be added to the docs?

ph-fritsche: Yes, this is some framework-specific knowledge that should be irrelevant to almost all developers.
The longer I think about it, the more I'm convinced that not adding such information would be better and that the fact that someone ran into this rather means we should do more to encourage people to use the right tool.
See https://github.com/testing-library/testing-library-docs/issues/963#issuecomment-1023150615

timdeschryver:
Thanks for the PR @Owrlean , but based on the discussion in this PR I'm going to close it.

PR #1030, opened 2022-03-25
docs: update user-event references

ph-fritsche:
Update all references to `user-event` preparing for the upcoming release.

This should be merged once the version is released on `latest`:
- [ ] `user-event@14` has been released.

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -10,10 +10,10 @@ import TabItem from '@theme/TabItem'
 advanced simulation of browser interactions than the built-in
 [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) method.
 
-> This page describes `user-event@13.5.0`.  
-> This version is no longer maintained. Please use
-> [`user-event@14`](user-event/intro.mdx) instead, as it includes important bug
-> fixes and new features.
+:::caution End of life This page describes `user-event@13.5.0`.  
+This version is no longer maintained. Please use
+[`user-event@14`](user-event/intro.mdx) instead, as it includes important bug
+fixes and new features. :::
 
 ## Installation
 
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -10,10 +10,14 @@ import TabItem from '@theme/TabItem'
 advanced simulation of browser interactions than the built-in
 [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) method.
 
-:::caution End of life This page describes `user-event@13.5.0`.  
+:::caution End of life
+
+This page describes `user-event@13.5.0`.  
 This version is no longer maintained. Please use
 [`user-event@14`](user-event/intro.mdx) instead, as it includes important bug
-fixes and new features. :::
+fixes and new features.
+
+:::
 
 ## Installation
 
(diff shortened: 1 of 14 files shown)


ph-fritsche:
Note: The `format` script produces broken block quotes in Markdown
```md
> bug fixes and new features. You can find the docs for `user-event@13.5.0`
> [here](../ecosystem-user-event.mdx).

becomes

> bug fixes and new features. You can find the docs for `user-event@13.5.0` >
> [here](../ecosystem-user-event.mdx).
```

docs/react-testing-library/migrate-from-enzyme.mdx
```diff
@@ -330,10 +330,14 @@ input's "checked" property is properly set. Let's see how we might write a test
 for that case:
 
 ```jsx
-test('handles click correctly', () => {
+test('handles click correctly', async () => {
   render(<Checkbox />)
+  const user = userEvent.setup()
+
+  // You can also call this method directly on userEvent,
```
MatanBobi, line 337: I'm not sure if that's even something we want to "share". If we believe that the recommended approach is to use the methods returned from `setup`, maybe this comment isn't helpful and can only cause bad practices. Wdyt?

MatanBobi (approved):
Looks good :)
Question about the formatting, it looks like a lot of things changed there, did we not run format before or maybe you're by any chance using a different prettier configs?

ph-fritsche:
> did we not run format before or maybe you're by any chance using a different prettier configs?

I think I skipped `format` before.

I'm not sure I like what prettier does to the Markdown files. A lot of lines are much harder to comprehend with the line breaks at fixed width instead of at the end of subclauses.

docs/react-testing-library/migrate-from-enzyme.mdx
ph-fritsche: I think it eases the transition for people who are used to the previous APIs. That's why the [Direct APIs](https://testing-library.com/docs/user-event/setup#direct-apis) still exist.

If there is only a single API call, it makes no difference, as the Direct APIs call `.setup()` internally and then use the method from there.

If it helps or hurts to mention it in the example - I don't know.

eps1lon:
Are the old links properly redirecting? For example, if an external site linked to a user-event page, will that link continue to work or return a 404?

ph-fritsche:
Good point. For now the old page isn't removed. Later I would remove it from the navigation, but keep the page itself as long as there is a significant number of downloads for v13.

eps1lon (approved):

docs/ecosystem-user-event.mdx
```diff
@@ -10,9 +10,10 @@ import TabItem from '@theme/TabItem'
 advanced simulation of browser interactions than the built-in
 [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) method.
 
-> This page describes `user-event@13.5.0`.
-If you are starting or actively working on a project,
-we recommend to use [`user-event@14.0.0-beta`](user-event/intro.mdx) instead, as it includes important bug fixes and new features.  
+> This page describes `user-event@13.5.0`.
```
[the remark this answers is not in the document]
ph-fritsche: I converted those blockquotes on the intro into admonitions. :)

ph-fritsche:
v14 has been [released](https://github.com/testing-library/user-event/releases/tag/v14.0.0) :rocket:

PR #1035, opened 2022-04-04
docs: queryByRole new description option

PaquitoSoft:
Docs for https://github.com/testing-library/dom-testing-library/pull/1120

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -20,6 +20,7 @@ getByRole(
     exact?: boolean = true,
     hidden?: boolean = false,
     name?: TextMatch,
+    description?: TextMatch,
     normalizer?: NormalizerFn,
     selected?: boolean,
     checked?: boolean,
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
Let's just create a new section for `description`. `name` will later have their own. 

This ensures readers don't get overwhelmed with context when they just want to read about either one option

PaquitoSoft:
> Let's just create a new section for `description`. `name` will later have their own.
> 
> This ensures readers don't get overwhelmed with context when they just want to read about either one option

@eps1lon thanks for the guidance.

I created the new section in this commit: 2eb9777e53c6e8ca35e997e5cd022e2e0f747531

docs/queries/byrole.mdx
```diff
@@ -314,3 +315,36 @@ To learn more about the `aria-level` property, see
...
+
+### `description`
+
+You can filter the returned elements by their
+[accessible description](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_description)
+for those cases where you have several elements with the same role and they
+don't have an accessible name but they do have a description.<br /> This would
+be the case for elements with
+[alertdialog](https://www.w3.org/TR/wai-aria-1.1/#alertdialog) role, where the
+`aria-describedby` attribute is used to describe the elements content.
```
timdeschryver, line 327: From a spellchecker:

```suggestion
`aria-describedby` attribute is used to describe the element's content.
```

or 

```suggestion
`aria-describedby` attribute is used to describe the content of the element.
```

docs/queries/byrole.mdx
```diff
@@ -314,3 +315,36 @@ To learn more about the `aria-level` property, see
...
+    </li>
+    <li role="alertdialog" aria-describedby="notification-id-2">
+      <div><button>Close</button></div>
+      <div id="notification-id-2">Your session is about to expire</div>
+    </li>
+  </ul>
+</body>
+```
+
+You can query an specific element like this
```
timdeschryver, line 346:
```suggestion
You can query the specific element like this
```

docs/queries/byrole.mdx
PaquitoSoft: Thanks for the catch 👍 

Updated in this commit: 9da0a1598132cb4c75ffa91d03ab9f16f24a9784

docs/queries/byrole.mdx
PaquitoSoft: In this case, I think the right wording would be
> You can query a specific element like this

since we want the user know he can select any specific element of those which have the same role.

Do you think we should actually use "the" instead?

(updated in this commit: 9da0a1598132cb4c75ffa91d03ab9f16f24a9784)

eps1lon (approved):

timdeschryver:
@all-contributors please add @PaquitoSoft for docs

PR #1039, opened 2022-04-07
docs(user-event): add `advanceTimers` option

ph-fritsche:
Add docs for https://github.com/testing-library/user-event/pull/907

### DIFF
--- a/docs/user-event/options.mdx
+++ b/docs/user-event/options.mdx
@@ -6,6 +6,13 @@ title: Options
 The following options allow to adjust the behavior of `user-event` APIs. They
 can be applied per [`setup()`](setup.mdx).
 
+### advanceTimers
+
+A function to advance your fake timers (if applicable), when we
+[`delay`](#delay) subsequent calls.
+
+`default`: `() => Promise.resolve()`
+
 ### applyAccept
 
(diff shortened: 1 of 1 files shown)


docs/user-event/options.mdx
```diff
@@ -6,6 +6,13 @@ title: Options
 The following options allow to adjust the behavior of `user-event` APIs. They
 can be applied per [`setup()`](setup.mdx).
 
+### advanceTimers
+
+A function to advance your fake timers (if applicable), when we
```
timdeschryver, line 11: This could be me, but I had to read the sentence multiple times to understand it.
What do you think about the following?

```suggestion
A function to advance your fake timers (if applicable). Used when we
```

```suggestion
A function to advance your fake timers (if applicable). Useful when we
```

ph-fritsche: "Useful" doesn't nail it. "Used" is okay I think. But maybe we should rephrase the whole sentence if it is hard to comprehend.

How about?
> If you are using fake timers, you need to advance your timers when we internally [delay](#delay) subsequent code.
> ```ts
> (delay: number) => Promise<void> | void
> ```

timdeschryver: I like that! Let's use that one.

PR #1044, opened 2022-04-19
Update queries/about.mdx

meatnordrink:
`queryOptions` are mentioned, but it's never explained where you actually pass them in. Once you know it's obvious, but it wasn't initially to me ("do I pass it into the render function, to modify behavior for all queries?"), and readers shouldn't have to test until they find it.

It would be great to link to a list of the available options in `queryOptions` in the section I added, but I couldn't find such a list.

I also added a note aboout the parameters that functions passed to queries will receive. It can be derived from examples later in the document, but you have to look for them, and I think it's clearer to make it explicit when the option to pass functions is mentioned.

### DIFF
--- a/docs/queries/about.mdx
+++ b/docs/queries/about.mdx
@@ -143,7 +143,8 @@ want to query `document.body` then you can use the [`screen`](#screen) export as
 demonstrated below (using `screen` is recommended).
 
 The primary argument to a query can be a _string_, _regular expression_, or
-_function_. There are also options to adjust how node text is parsed. See
+_function_. If you pass a function, it will receive `content` and `element` as parameters.
+There are also options to adjust how node text is parsed. See
 [TextMatch](#textmatch) for documentation on what can be passed to a query.
 
 Given the following DOM elements (which can be rendered by React, Vue, Angular,
@@ -169,6 +170,13 @@ const inputNode1 = screen.getByLabelText('Username')
 // Without screen, you need to provide a container:
 const container = document.querySelector('#app')
 const inputNode2 = getByLabelText(container, 'Username')
+```
+### `queryOptions`
+
+There are several options that you can pass to queries, though in most cases you won't need any. 
+
+```js
+screen.getByRole('list', { hidden: true })
 ```
 
 ### `screen`
@@ -238,7 +246,7 @@ cy.findByLabelText('Example').should('exist')
 
 Most of the query APIs take a `TextMatch` as an argument, which means the
 argument can be either a _string_, _regex_, or a _function_ which returns `true`
-for a match and `false` for a mismatch.
+for a match and `false` for a mismatch. If you pass a function, it will receive `content` and `element` as parameters.
 
 ### TextMatch Examples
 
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @meatnordrink, thanks for taking the time to open this one :)
Two things about this one, the first is that `queryOptions` is different for every type of query so I'd suggest maybe linking to an example and stating it's just an example reference (instead of giving an example that might change later on).
The second regarding the `content` and `element`, If you believe that's missing, I think it will be best to add the function signature we're expecting including the types, wdyt about that?

Thanks again!

meatnordrink:
Hi @MatanBobi, thanks so much for the feedback! 

I tried to address your feedback in the following ways: 

* For the `queryOptions` section, I noted that the options are different for each query type. (I'd love to be able to link to a list of the possible options, if it exists.) You suggested linking to an example, but it seemed better to me to include it in the code, both because it's more convenient, and because links can break?
* For the `content` and `element` note, I followed your suggestion and added the function signature in the `textMatch` section. I removed the edit to the initial "Using queries" section, since it references the `textMatch` section.

Let me know if these changes aren't what you were hoping for! Thanks!

MatanBobi:
Thanks @meatnordrink :)
Regarding the example and not links, I get your point but docusarus outputs whenever there's a broken link and we do follow those warnings so it will be easier for us to spot those over a code example in a place where we don't think even exists, I believe a link here will be much more maintainable looking forwards.
The signature looks good, thanks!
If you can just change the example to a link, I'll merge this one, thanks again :)

meatnordrink:
@MatanBobi I tried again, linking to an example of documentation for the `queryOptions` for a specific query; is this what you meant? 

I initially thought you meant to link to a CodePen example or something, but that just seems really weird, and I don't see that being done anywhere else in the docs.

So let me know if I'm still misunderstanding.

Thanks!

docs/queries/about.mdx
```diff
@@ -170,6 +170,9 @@ const inputNode1 = screen.getByLabelText('Username')
 const container = document.querySelector('#app')
 const inputNode2 = getByLabelText(container, 'Username')
 ```
+### `queryOptions`
+
+You can pass a `queryOptions` object with the query type. See the docs for each query type to see available options, e.g. [here](https://testing-library.com/docs/queries/byrole#api).
```
MatanBobi, line 175:
```suggestion
You can pass a `queryOptions` object with the query type. See the docs for each query type to see available options, e.g. [byRole API](queries/byrole.mdx#api).
```

MatanBobi (changes requested):
Sorry it took me time to review this one @meatnordrink.
This is definitely a good step forwards but here's a small feedback:
1. We prefer using relative links so when building, all links will refer to the same "version" and also if we change the domain, those will still work and won't break.
2. As a general rule of thumb, I prefer not having a link named over text like "here" is it gives a bad user experience for users using screen readers :)

meatnordrink:
No worries, @MatanBobi , we all have actual jobs to keep up with too, I appreciate you taking the time to give feedback! 

Great points; I implemented the changes as suggested. 

Let me know if there's anything further!

MatanBobi (approved):

MatanBobi:
That's perfect, thanks @meatnordrink :)

MatanBobi:
@all-contributors please add @meatnordrink for docs :)

PR #1046, opened 2022-04-29
Update into Intro.mdx

amjedidiah:
More explanation on implementation details to avoid testing for

### DIFF
--- a/docs/intro.mdx
+++ b/docs/intro.mdx
@@ -54,7 +54,7 @@ You may want to avoid the following implementation details:
 1. Internal state of a component
 1. Internal methods of a component
 1. Lifecycle methods of a component
-1. Child components
+1. Child components, i.e: each component(parent and children) should have its own test file
 
 [jest]: https://jestjs.io
 [jsdom]: https://github.com/jsdom/jsdom


docs/intro.mdx
```diff
@@ -54,7 +54,7 @@ You may want to avoid the following implementation details:
 1. Internal state of a component
 1. Internal methods of a component
 1. Lifecycle methods of a component
-1. Child components
+1. Child components, i.e: each component(parent and children) should have its own test file
```
ph-fritsche, line 57: This is the opposite of what is encouraged in this paragraph.

Testing Library encourages you to test your components through integration. A child component is an implementation detail.

MatanBobi: Integration tests aims to test a use-case or a user-flow, so in this case I'd suggest testing the `Login` component :)

amjedidiah: Alright then.
And should the form component still be used in a Register component, the test will be written for the Register component and not the form component, yes?

MatanBobi: Yeah, exactly.
If you have a `Form` component which is a generic one and you use it multiple time in different cases, you might want to also unit test that component. I recommend reading this one:
https://matan.io/posts/to-unit-or-integration

MatanBobi:
Closing as this change isn't what we aim to push :)
Thanks for the effort!

docs/intro.mdx
amjedidiah: Alright, thanks man.
Better understood.

PR #1059, opened 2022-05-12
Just add an alternative location for the React Testing Examples

tarjei:
I am in no way affiliated with the project, I just wanted to add a link that was not dead.

### DIFF
--- a/docs/example-external.mdx
+++ b/docs/example-external.mdx
@@ -6,10 +6,7 @@ sidebar_label: External Examples
 
 ## Code Samples
 
-- You can find more React Testing Library examples at
-  [react-testing-examples.com](https://react-testing-examples.com/jest-rtl/).
-  
-  If the link does not work, try the netlify one: [react-testing-examples.netlify.app](https://react-testing-examples.netlify.app/).
+- You can find more React Testing Library examples at [https://react-testing-examples.netlify.app/](https://react-testing-examples.netlify.app/).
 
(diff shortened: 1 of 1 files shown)


docs/example-external.mdx
```diff
@@ -8,6 +8,8 @@ sidebar_label: External Examples
 
 - You can find more React Testing Library examples at
   [react-testing-examples.com](https://react-testing-examples.com/jest-rtl/).
+  
+  If the link does not work, try the netlify one: [react-testing-examples.netlify.app](https://react-testing-examples.netlify.app/).
```
eps1lon, line 12: The original link may just be dead temporarily. Did you check with the project if they're aware of any issue?

The problem with the netlify link is that it's now tied to their deploy service. When they switch (e.g. to Vercel) the link is dead again. 

I'd rather ask the maintainers of https://react-testing-examples.com/ first for clarificiation. For example, our domain was dead a few days ago as well which was just a temporary netlify issue.

timdeschryver, line 12: Hey @tarjei sorry for the late response, I forgot about it.

I asked the author and this can point directly to https://react-testing-examples.netlify.app/ (https://github.com/skidding/react-testing-examples/issues/56)

tarjei: :+1:  I updated the PR

MatanBobi (approved):
Thanks! :)

PR #1061, opened 2022-05-17
Updated jest v27 instructions to v28

abelhbeyene:
Added instruction on jsdom setup as it now isn't shipped with jest as of v28.

https://jestjs.io/docs/upgrading-to-jest28#jsdom
"If you are using JSDOM test environment, jest-environment-jsdom package now must be installed separately"

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -285,17 +285,36 @@ using Create React App without TypeScript, save this to `jsconfig.json` instead.
 
 ### Jest 28
 
-If you're using a recent version of Jest (28 or later), jest-environment-jsdom package now must be installed separately. 
+If you're using Jest 28 or later, jest-environment-jsdom package now must be installed separately. 
 ```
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/setup.mdx
```diff
@@ -283,15 +283,19 @@ using Create React App without TypeScript, save this to `jsconfig.json` instead.
 }
 ```
 
-### Jest 27
+### Jest 28
```
timdeschryver, line 286: What do you think of keeping the Jest 27 docs and adding a new section for Jest 28?
Because Jest 27 is still popular (and other Jest versions are also included).

abelhbeyene: Sure.
I initially thought to do that but the terminology was getting confusing and duplicated. E.g. 27 or later, the docblocks are duplicated.
I have updated it now. Let me know what you think

timdeschryver (approved):

PR #1084, opened 2022-06-08
docs: Vitest doesn't need global-jsdom

silvenon:
Vitest doesn't need `global-jsdom`, setting `environment` to `jsdom` already defines it globally.


eps1lon (approved):

eps1lon:
@all-contributors add @silvenon for docs

PR #1086, opened 2022-06-09
Improving intro clarity

meatnordrink:
Several engineers at our company have had trouble understanding what the v14 upgrade entailed for us, after just looking at the intro. Having now actually done it, it's simple enough, but these changes attempt to clarify things in the intro by 
a) improving grammar
b) clarifying a few key points and examples
c) providing a direct link to the v14 changelog

### DIFF
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -21,8 +21,8 @@ with any framework as long as there is a DOM.
 ## Differences with `fireEvent`
 
 The built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a
-utility for easily dispatching events. It dispatches exactly the events you tell it
-to and just those - even if those exact events would not have been dispatched in a
+utility for easily dispatching events. It dispatches exactly, and only, the events you tell it
+to - even if those exact events would not have been dispatched in a
 real browser interaction.
 
 `user-event` on the other hand dispatches events as they would be dispatched by a user interacting with the document in a browser. That might lead to the same events you previously dispatched per `fireEvent` directly, but it also might catch bugs
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -32,8 +32,8 @@ to test interaction with your components.
 
 ## Writing tests with `userEvent`
 
-We recommend using [`userEvent.setup()`](setup.mdx) when rendering your
-component and inlining that rendering and setup in your test, or using a setup
+We recommend invoking [`userEvent.setup()`](setup.mdx) before the component is rendered.
+This can be done in the test itself, or by using a setup
 function. We discourage rendering or using any `userEvent` functions outside of
 the test itself - e.g. in a `before`/`after` hook - for reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -11,36 +11,35 @@ events that would happen if the interaction took place in a browser.
 
 These docs describe `user-event@14`. We recommend updating your projects to this
 version, as it includes important bug fixes and new features. You can find the
-docs for `user-event@13.5.0` [here](../ecosystem-user-event.mdx).
+docs for `user-event@13.5.0` [here](../ecosystem-user-event.mdx), and the changelog for the release [here](https://github.com/testing-library/user-event/releases/tag/v14.0.0).
 
 :::
 
 While most examples with `user-event` are for `React`, the library can be used
 with any framework as long as there is a DOM.
 
-## Difference to `fireEvent`
+## Differences with `fireEvent`
 
 The built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a
-utility to easily dispatch events. It dispatches exactly the events you tell it
-to and just those - even if those exact events never had been dispatched in a
-real interaction in a browser.
+utility for easily dispatching events. It dispatches exactly the events you tell it
+to and just those - even if those exact events would not have been dispatched in a
+real browser interaction.
 
-`user-event` on the other hand dispatches the events like they would happen if a
-user interacted with the document. That might lead to the same events you
-previously dispatched per `fireEvent` directly, but it also might catch bugs
+`user-event` on the other hand dispatches events as they would be dispatched by a user interacting with the document in a browser. That might lead to the same events you previously dispatched per `fireEvent` directly, but it also might catch bugs
 that make it impossible for a user to trigger said events. This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
 to test interaction with your components.
 
 ## Writing tests with `userEvent`
 
-We recommend to use [`userEvent.setup()`](setup.mdx) when rendering your
-component and inline that rendering and setup in your test or use a setup
+We recommend using [`userEvent.setup()`](setup.mdx) when rendering your
+component and inlining that rendering and setup in your test, or using a setup
 function. We discourage rendering or using any `userEvent` functions outside of
 the test itself - e.g. in a `before`/`after` hook - for reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
 
 ```js
+// inlining
 test('trigger some awesome feature when clicking the button', async () => {
   const user = userEvent.setup()
   render(<MyComponent />)
@@ -52,6 +51,7 @@ test('trigger some awesome feature when clicking the button', async () => {
 ```
 
 ```js
+// setup function
 function setup(jsx) {
   return {
     user: userEvent.setup(),
@@ -64,3 +64,5 @@ test('render with a setup function', async () => {
   // ...
 })
 ```
+
+Note that, while directly invoking APIs such as `userEvent.click()` (which will trigger `setup` internally) is (still supported in v14)[https://testing-library.com/docs/user-event/setup#direct-apis], this option exists to ease the migration from v13 to v14, and for simple tests. We recommend using the methods on the instances returned by `userEvent.setup()`.
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -20,15 +20,9 @@ with any framework as long as there is a DOM.
 
 ## Differences from `fireEvent`
 
-The built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a
-utility for easily dispatching events. It dispatches exactly, and only, the events you tell it
-to - even if those exact events would not have been dispatched in a
-real browser interaction.
+`fireEvent` simulates _browser events_, whereas `user-event` simulates full *interactions*, which may fire multiple events and do additional checks along the way.
 
-`user-event` on the other hand dispatches events as they would be dispatched by a user interacting with the document in a browser. That might lead to the same events you previously dispatched per `fireEvent` directly, but it also might catch bugs
-that make it impossible for a user to trigger said events. This is
-[why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
-to test interaction with your components.
+Testing Library's built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a lightweight wrapper around the browser's low-level `dispatchEvent` API, which allows developers to trigger any event on any element. This is often sufficient for testing simple behavior, e.g., checking that a click event triggers the right callback, but there are cases where the browser does more than just trigger one event for one interaction. For example, when a user types into a text box, the element is first focused, and then input events are fired as they type. `user-event` also adds visibility and interactability checks before triggering the flow, to simulate how a user wouldn't click an element they can't see, and the browser wouldn't let them type in a disabled text box.
 
 ## Writing tests with `userEvent`
 
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -11,7 +11,9 @@ events that would happen if the interaction took place in a browser.
 
 These docs describe `user-event@14`. We recommend updating your projects to this
 version, as it includes important bug fixes and new features. You can find the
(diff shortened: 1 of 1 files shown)


docs/user-event/intro.mdx
```diff
@@ -11,36 +11,35 @@ events that would happen if the interaction took place in a browser.
...
 that make it impossible for a user to trigger said events. This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
 to test interaction with your components.
 
 ## Writing tests with `userEvent`
 
-We recommend to use [`userEvent.setup()`](setup.mdx) when rendering your
-component and inline that rendering and setup in your test or use a setup
+We recommend using [`userEvent.setup()`](setup.mdx) when rendering your
+component and inlining that rendering and setup in your test, or using a setup
```
timdeschryver, line 36: Hey, thanks for the improvements.
For me, "inlining" sounds confusing, and I have to take a look at the examples to know what it means. What do you think of the following? 

```suggestion
We recommend invoking [`userEvent.setup()`](setup.mdx) before the component is rendered.
This can be done in the test itself, or by using a setup
```

docs/user-event/intro.mdx
```diff
@@ -11,36 +11,35 @@ events that would happen if the interaction took place in a browser.
...
 
-## Difference to `fireEvent`
+## Differences with `fireEvent`
 
 The built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a
-utility to easily dispatch events. It dispatches exactly the events you tell it
-to and just those - even if those exact events never had been dispatched in a
-real interaction in a browser.
+utility for easily dispatching events. It dispatches exactly the events you tell it
+to and just those - even if those exact events would not have been dispatched in a
```
ph-fritsche, line 25: This is wrong grammar according to [Cambridge Dictionary](https://dictionary.cambridge.org/de/grammatik/britisch-grammatik/conditionals-if).

docs/user-event/intro.mdx
meatnordrink: Sure - "inlining" was just carried over from the previous wording, I found it a bit confusing as well.

docs/user-event/intro.mdx
meatnordrink: Thanks for the comment @ph-fritsche! Can you be a little more specific about what is wrong? I believe this is correct under "Imagined conditions: The third conditional", in the same entry you referenced. 

I'll admit others may beat me at finding grammar rules, but I can say that, as a native English speaker and former copy editor, this sounds far more correct and native to me than the original wording, which I think is both confusing and incorrect.

I did make an edit to try to bring the whole sentence a little closer to grammatical correctness. (I was going, in the first place, for greater clarity rather than technical grammar, but I agree that both are valuable.)

timdeschryver (approved):

docs/user-event/intro.mdx
ph-fritsche: "Imagined conditions: The third conditional" lists the past perfect (3rd form) as the correct conjugation for an imagined condition that would never happen: "even if those had been"
A "would" would be part of the main clause not the conditional. But as the main clause precedes the conditional and is not imaginary itself, thus not subjunctive, I think it's correct to omit it here.

I'm a native German speaker, the use of "would" in the conditional *feels* right to me as English lacks a Konjunktiv and the irrealis mood is only inferred by the time relation between the conditional and the main clause. But I was told it is only used for courtesy in spoken English, but is considered incorrect in formal writing.

> "He is on his way and will probably forget this important thing."
> Courtesy/Spoken: "And even if you would have written him a note, he would still forget it."
> Written: "And even if someone had written him a note, he would still forget it."

meatnordrink: @ph-fritsche , thanks for clarifying! 

The technical grammar of English is indeed tricky, and is made trickier by the fact that, as native speakers, we tend not to study it formally, so even if we know something is correct/incorrect, it's often difficult for us to point to the grammar rule that says so. 

In reference to the use of "would" in the conditional - in the example you gave

>  "And even if you would have written him a note, he would still forget it."

We'd actually say, or write - 

> "Even if you had written him a note, he would still have forgotten it." 

(Informal, "Even if you'd written him a note, he'd still have forgotten it.")

Or, if it hasn't happend yet: 

> "Even if you write him a note, he will still forget."
 
Or speaking in general: 

> "Even if you write him a note, he still forgets." OR "Even if you write him a note, he will still forget."

For the sentence in question here:

**Previous Form:** 

> It dispatches exactly the events you tell it to and just those - even if those exact events never had been dispatched in a
real interaction in a browser.

This is incorrect. "Never had been" refers to something concrete that actually happened; this is an imagined instance, grammatically. Further, it would be "had never been" (though that's not correct here either).

**Proposed Form:**
 
> It dispatches exactly, and only, the events you tell it to - even if those exact events would not have been dispatched in a real browser interaction.

I believe this is the correct form for describing a hypothetical conditional. The "would" is necessary; if you drop it and just write "had not been," it incorrectly refers to a concrete, non-hypothetical event. If you replace it with "will", it incorrectly refers to a concrete future. Both forms add confusion.

Do you have a specific alternative form you believe would be more correct? I'm not opposed to changing the wording, if we can find something that feels correct to both of us. 

Also, if it would be helpful, I can solicit feedback from other engineers at my workplace on this PR. There are plenty of native English speakers there, and even a few former teachers. We have writers on staff, too, come to think of it. 

Let me know how you'd like to proceed.

ph-fritsche: >This is incorrect. "Never had been" refers to something concrete that actually happened; this is an imagined instance, grammatically.

"even if it never had been" is already *irrealis* as per being a conditional referring to the past.
According to Oxford Dictionary it isn't ambiguous because the main clause is in present tense, so a non-hypothetical conditional would need to be in present or present perfect.
In a *conditional III* the correct phrasing would be:
> Even if the events never had been dispatched in a browser, the utility would dispatch exactly those.

I guess the confusion is caused by the main clause omitting the "would". Which in the case of inversion *feels* correct to me, as we describe that it factually does dispatch those events, but I don't have any dictionary reference to back this up.

Maybe this discussion boils down to differences between US and Oxford English.

> However, some varieties of English regularly use would (contracted to 'd) and would have ('d have) in counterfactual condition clauses, although this is often considered non-standard:
>
>      If you'd leave (standard: you left) now, you'd be on time.
>      If you'd have told (standard: you had told) me, we could've done something about it.
>
>Such use of would is widespread especially in spoken [American English](https://en.wikipedia.org/wiki/American_English) in all sectors of society. It is not usually found in more formal writing; however some sources describe it as acceptable US English, no longer labeling it colloquial.[[8]](https://en.wikipedia.org/wiki/English_conditional_sentences#cite_note-8)[[9]](https://en.wikipedia.org/wiki/English_conditional_sentences#cite_note-9)

As mentioned above, using the would in the conditional here *feels* better to me too. But this is something I've been scolded for too much. ;)
If we can ask some native English speakers who also have a professional background on linguistics, let's do it!
I'm more than happy about a workaround for the lack of Konjunktiv in the English language. \*wavingmyfistatformerteachers\* ;D

docs/user-event/intro.mdx
```diff
@@ -11,36 +11,35 @@ events that would happen if the interaction took place in a browser.
...
+to - even if those exact events would not have been dispatched in a
+real browser interaction.
 
-`user-event` on the other hand dispatches the events like they would happen if a
-user interacted with the document. That might lead to the same events you
-previously dispatched per `fireEvent` directly, but it also might catch bugs
+`user-event` on the other hand dispatches events as they would be dispatched by a user interacting with the document in a browser. That might lead to the same events you previously dispatched per `fireEvent` directly, but it also might catch bugs
 that make it impossible for a user to trigger said events. This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
 to test interaction with your components.
```
alexkrolick, line 31: _exactly, and only,_ doesn't greatly clarify what's going on here, in my opinion. Let's try reworking this paragraph completely, like this:

```suggestion
`fireEvent` simulates _browser events_, whereas `user-event` simulates full `interactions`, which may fire multiple events and do additional checks along the way.

Testing Library's built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a lightweight wrapper around the browser's low-level dispatchEvent API, which allows developers to trigger any event on any element. This is often sufficient for testing simple behavior, e.g., checking that a click event triggers the right callback, but there are cases where the browser does more than just trigger one event for one interaction, for example, when a user types into a text box, the element is first focused, and then input events are fired as they type. `user-event` also adds visibility and interactability checks before triggering the flow, to simulate how a user wouldn't click an element they can't see, and the browser wouldn't let them type in a disabled text box.
```

docs/user-event/intro.mdx
meatnordrink: @ph-fritsche 
> As mentioned above, using the would in the conditional here feels better to me too. But this is something I've been scolded for too much. ;)

Haha, okay! I'll find someone with a linguistic background and ask them to comment!

docs/user-event/intro.mdx
meatnordrink: @alexkrolick , I like it! I'll apply the suggestion (though manually, as attempting to just accept the suggested edit produces a "Applying suggestions on deleted lines is not currently supported" error). 

@ph-fritsche , this removes the line we were discussing entirely, which perhaps resolves our problem. Do you feel good about this solution?

ph-fritsche: Rewriting the discussed sentence or the complete paragraph might just be solution.
But I feel this variant is a step back regarding the recommendation our users might read into this.

We've had some discussions about what we want to recommend for simulating any interactions before.
`userEvent` wasn't mature enough, so we pointed people towards `fireEvent` and mentioned that there is also `userEvent` doing more. v14 tackled all the issues being referenced every time this was discussed.
So when we talked about it again prior to releasing v14, the consensus was:
*People should use `userEvent` for testing any user input nowadays.
`fireEvent` works as an escape hatch for edge cases which are not implemented yet and for library authors who need control over the events and their properties to e.g. test their event system.*

> This is often sufficient for testing simple behavior, e.g., checking that a click event triggers the right callback

We want to communicate the opposite: Usually it is not a good idea to test simple behavior with `fireEvent` as it poses the risk to just repeat false assumptions in component and test.
The linked blog post contains an example for how testing an implementation detail (the exact event) provides a false-negative test although the presumably tested interaction is made impossible by a bug.
It also hints at how testing a perceived event (`change`) although working might create a test that is unrelated to the also working component - which might result in a very confusing scenario if one of them breaks one day.

docs/user-event/intro.mdx
```diff
@@ -11,36 +11,29 @@ events that would happen if the interaction took place in a browser.
...
 with any framework as long as there is a DOM.
 
-## Difference to `fireEvent`
+## Differences from `fireEvent`
 
-The built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a
-utility to easily dispatch events. It dispatches exactly the events you tell it
-to and just those - even if those exact events never had been dispatched in a
-real interaction in a browser.
+`fireEvent` simulates _browser events_, whereas `user-event` simulates full *interactions*, which may fire multiple events and do additional checks along the way.
```
eps1lon, line 23:
```suggestion
`fireEvent` dispatches _browser events_, whereas `user-event` simulates full *interactions*, which may fire multiple events and do additional checks along the way.
```

`fireEvent` doesn't really simulate anything. It really just dispatches the events of the corresponding renderer.

docs/user-event/intro.mdx
```diff
@@ -11,36 +11,29 @@ events that would happen if the interaction took place in a browser.
...
-user interacted with the document. That might lead to the same events you
-previously dispatched per `fireEvent` directly, but it also might catch bugs
-that make it impossible for a user to trigger said events. This is
-[why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
-to test interaction with your components.
+Testing Library's built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a lightweight wrapper around the browser's low-level `dispatchEvent` API, which allows developers to trigger any event on any element. This is often sufficient for testing simple behavior, e.g., checking that a click event triggers the right callback, but there are cases where the browser does more than just trigger one event for one interaction. For example, when a user types into a text box, the element is first focused, and then input events are fired as they type. `user-event` also adds visibility and interactability checks before triggering the flow, to simulate how a user wouldn't click an element they can't see, and the browser wouldn't let them type in a disabled text box.
```
alexkrolick, line 25:
```suggestion
Testing Library's built-in [`fireEvent`](dom-testing-library/api-events.mdx#fireevent) is a lightweight wrapper around the browser's low-level `dispatchEvent` API, which allows developers to trigger any event on any element. The problem is that in many cases the browser does more than just trigger one event for one interaction. For example, when a user types into a text box, the element is first focused, and then input events are fired as they type. `user-event` also adds visibility and interactability checks before triggering the flow, to simulate how a user wouldn't click an element they can't see, and the browser wouldn't let them type in a disabled text box.
```

docs/user-event/intro.mdx
alexkrolick: @ph-fritsche OK, how about this then. Removed that sentence:

https://github.com/testing-library/testing-library-docs/pull/1086/files#r897024155

ph-fritsche: @alexkrolick I think that the idea to make this section less abstract might be the right choice.

I tried to make it even more explicit/"correct" in the description of user-event and re-added the link the post with the example, so that the section ends with an explicit recommendation and the option to read more about this if someone cares.

What do you think?

https://github.com/meatnordrink/testing-library-docs/compare/patch-1...testing-library:pr-1086/phf

alexkrolick: That version sounds good to me

alexkrolick: Do you want to accept that patch and merge this then?

ph-fritsche: @meatnordrink I went ahead and merged the change into [meatnordrink:patch-1](https://github.com/meatnordrink/testing-library-docs/commits/patch-1)

ph-fritsche (approved):

docs/user-event/intro.mdx
meatnordrink: Thanks, @ph-fritsche  - sorry for not replying, I was on vacation (and camping) last week!

alexkrolick (approved):

PR #1102, opened 2022-07-01
Clarifying user-event's capabilities vs fire-event

meatnordrink:
The current intro makes it sounds like `user-event` can test all valid user interactions, and is a replacement for `fire-event` in all user-interaction cases. However, in our engineering dept., we've found that this is not the case. For example, to pick a specific value on a slider, we use:

```js
fireEvent.change(screen.getByRole('slider'), { target: { value: '3000' } })
```

We haven't been able to find any `user-event` equivalent to this. So unless there's something we're not aware of (and we've [posted the question to SO](https://stackoverflow.com/q/72775505/9154668) and received no responses), then noting that `user-event` can not replace `fire-event` in all cases would improve the clarity and usefulness of the `user-event` intro.

### DIFF
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -47,6 +47,9 @@ to test interaction with your components.
 There are, however, some user interactions or aspects of these [that aren't yet implemented and thus can't yet be described with `user-event`](https://github.com/testing-library/user-event/issues?q=is%3Aopen+label%3Aaccuracy%2Cenhancement).
 In these cases you can use `fireEvent` to dispatch the concrete events that your software relies on.
 
+Note that this makes your component and/or test reliant upon your assumptions about the concrete aspects of the interaction being correct.
+Therefore if you already put in the work to specify the correct aspects of such interaction, please consider contributing to this project so that user-event might cover these cases too.
+
 ## Writing tests with `userEvent`
 
 We recommend invoking [`userEvent.setup()`](setup.mdx) before the component is
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -44,7 +44,8 @@ This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
 to test interaction with your components. 
 
-There are, however, some user interactions that can't yet be described with `user-event` - selecting values on a slider, for example - for which `fire-event` is necessary.
+There are, however, some user interactions or aspects of these [that aren't yet implemented and thus can't yet be described with `user-event`](https://github.com/testing-library/user-event/issues?q=is%3Aopen+label%3Aaccuracy%2Cenhancement).
+In these cases you can use `fireEvent` to dispatch the concrete events that your software relies on.
 
 ## Writing tests with `userEvent`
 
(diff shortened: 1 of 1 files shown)


ph-fritsche:
`user-event` is not an equivalent replacement to `fireEvent`.

> `fireEvent` dispatches _DOM events_, whereas `user-event` simulates full
_interactions_, which may fire multiple events and do additional checks along
the way.

You *should* use `user-event` to simulate user interactions.

Like with every other software there are limitations.
You can use `fireEvent` in these cases, but then you are not testing a user interaction but the handling of a concrete event.

Regarding `<input type="range"/>` element, see https://github.com/testing-library/user-event/issues/871

docs/user-event/intro.mdx
```diff
@@ -42,7 +42,9 @@ factors in that the browser e.g. wouldn't let a user click a hidden element or
 type in a disabled text box.  
 This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
-to test interaction with your components.
+to test interaction with your components. 
+
+There are, however, some user interactions that can't yet be described with `user-event` - selecting values on a slider, for example - for which `fire-event` is necessary.
```
ph-fritsche, line 47:
```suggestion
There are, however, some user interactions or aspects of these [that aren't yet implemented and thus can't yet be described with `user-event`](https://github.com/testing-library/user-event/issues?q=is%3Aopen+label%3Aaccuracy%2Cenhancement).
In these cases you can use `fireEvent` to dispatch the concrete events that your software relies on.
```

We should link to the issue board here instead of cementing these limitations.

Maybe this is a good place to ask for contributions.

> Note that this makes your component and/or test reliant upon your assumptions about the concrete aspects of the interaction being correct.
Therefore if you already put in the work to specify the correct aspects of such interaction, please consider contributing to this project so that `user-event` might cover these cases too.

ph-fritsche (changes requested):

meatnordrink:
Great suggestions, @ph-fritsche - I implemented both of them. Feel free to tweak the wording, I just put in exactly what you suggested.

timdeschryver (approved):
👍 LGTM.
Leaving this for @ph-fritsche for a final review.

docs/user-event/intro.mdx
```diff
@@ -42,7 +42,13 @@ factors in that the browser e.g. wouldn't let a user click a hidden element or
...
 This is
 [why you should use `user-event`](https://ph-fritsche.github.io/blog/post/why-userevent)
-to test interaction with your components.
+to test interaction with your components. 
+
+There are, however, some user interactions or aspects of these [that aren't yet implemented and thus can't yet be described with `user-event`](https://github.com/testing-library/user-event/issues?q=is%3Aopen+label%3Aaccuracy%2Cenhancement).
+In these cases you can use `fireEvent` to dispatch the concrete events that your software relies on.
+
+Note that this makes your component and/or test reliant upon your assumptions about the concrete aspects of the interaction being correct.
+Therefore if you already put in the work to specify the correct aspects of such interaction, please consider contributing to this project so that user-event might cover these cases too.
```
ph-fritsche, line 51:
```suggestion
Therefore if you already put in the work to specify the correct aspects of such interaction, please consider contributing to this project so that `user-event` might cover these cases too.
```

ph-fritsche (changes requested):
missing backticks

ph-fritsche (approved):

ph-fritsche:
Thanks for helping us improve the docs @meatnordrink :heart:

PR #1103, opened 2022-07-04
docs: fix the user-event example

polinamochan:
Resolve: https://github.com/testing-library/testing-library-docs/issues/912

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -615,16 +615,21 @@ method.
 Usage example:
 
 ```jsx
-import React from 'react'
+import React, { useState } from 'react'
 import {render, screen} from '@testing-library/react'
 import userEvent, {specialChars} from '@testing-library/user-event'
 
+const InputElement = () => {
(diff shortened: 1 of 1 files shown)


docs/ecosystem-user-event.mdx
```diff
@@ -615,22 +615,27 @@ method.
...
-      <input id="my-input" type="text" value="This is a bad example" />
-    </div>,
+    <InputElement/>,
   )
   const input = screen.getByLabelText(/example/i)
   input.setSelectionRange(10, 13)
   userEvent.type(input, `${specialChars.backspace}good`)
 
-  expect(input).toHaveValue('This is a good example')
+  expect(screen.getByRole('textbox')).toHaveValue('This is a good example')
```
MatanBobi, line 638: I think we can still use the `input` variable we created above.. Using `screen.getByRole('textbox')` takes into account that our reader knows that `input` has an implicit aria role..

polinamochan: done :)

MatanBobi (approved):
Thanks! :)

PR #1111, opened 2022-07-09
docs: use correct react router v5 api in example

thanhsonng:
This PR addresses the issue https://github.com/testing-library/testing-library-docs/issues/1110.

### DIFF
--- a/docs/example-react-router.mdx
+++ b/docs/example-react-router.mdx
@@ -136,8 +136,9 @@ test('rendering a component that uses useLocation', () => {
 Given your App uses the correct React Router v5 API.
 
 ```jsx
+// app.js
 import React from 'react'
-import { Link, Route, Switch, useLocation } from 'react-router-dom'
+import {Link, Route, Switch, useLocation} from 'react-router-dom'
 
 const About = () => <div>You are on the about page</div>
 const Home = () => <div>You are home</div>
@@ -149,25 +150,23 @@ export const LocationDisplay = () => {
   return <div data-testid="location-display">{location.pathname}</div>
 }
 
-export const App = () => {
-  return (
-    <div>
-      <Link to="/">Home</Link>
+export const App = () => (
+  <div>
+    <Link to="/">Home</Link>
 
-      <Link to="/about">About</Link>
+    <Link to="/about">About</Link>
 
-      <Switch>
-        <Route exact path="/" component={Home} />
+    <Switch>
+      <Route exact path="/" component={Home} />
 
-        <Route path="/about" component={About} />
+      <Route path="/about" component={About} />
 
-        <Route component={NoMatch} />
-      </Switch>
+      <Route component={NoMatch} />
+    </Switch>
 
-      <LocationDisplay />
-    </div>
-  )
-}
+    <LocationDisplay />
+  </div>
+)
 ```
 
 In your tests, pass the history object as a whole to the Router component.
--- a/docs/example-react-router.mdx
+++ b/docs/example-react-router.mdx
@@ -133,7 +133,7 @@ test('rendering a component that uses useLocation', () => {
 
 ## Testing Library and React Router v5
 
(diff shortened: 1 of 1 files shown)


docs/example-react-router.mdx
```diff
@@ -133,8 +133,35 @@ test('rendering a component that uses useLocation', () => {
...
 ## Testing Library and React Router v5
 
-In React Router v5, you need to pass the history object as a whole to the Route
-component.
+Given your App uses the correct React Router v5 API:
+
+```jsx
+import { Link, Route, Switch, useLocation } from 'react-router-dom'
+
+// ...
```
MatanBobi, line 141: Any reason to remove the `LocationDisplay` and page components? We want an example that will work in a copy paste scenario :)

MatanBobi (changes requested):
Thanks for this one!
Small fix and I'll merge this.

docs/example-react-router.mdx
thanhsonng: No special reason, I was just thinking that we don't need to repeat the same information, but if we want to let user copy-paste then I'm happy to add the rest of the code.

MatanBobi: I think we're good with having these kind of duplications if it will make people's life a bit easier, thanks!
You can publish this PR btw.

thanhsonng: @MatanBobi I have updated as per you feedback. Sorry for the force push, I realized the author email was wrong so I have to amend the commits

thanhsonng:
I have added a small note to make sure users install the correct `history` package version to use with `react-router` v5.

docs/example-react-router.mdx
```diff
@@ -133,8 +133,45 @@ test('rendering a component that uses useLocation', () => {
 
 ## Testing Library and React Router v5
 
-In React Router v5, you need to pass the history object as a whole to the Route
-component.
+Given your App uses the correct React Router v5 API.
```
MatanBobi, line 136: I think we can remove this line as it doesn't really add anything to the reader :)

docs/example-react-router.mdx
```diff
@@ -133,8 +133,45 @@ test('rendering a component that uses useLocation', () => {
...
+      <Route component={NoMatch} />
+    </Switch>
+
+    <LocationDisplay />
+  </div>
+)
+```
+
+In your tests, pass the history object as a whole to the Router component.
+**Note:** React Router v5 [only works with History v4](https://github.com/remix-run/history#documentation),
```
MatanBobi, line 173: Are we linking exactly what we want to link? because it looks like we're linking to the history package's docs. If that's what we want to link to, I think we should only make the "History" word a hyperlink.

MatanBobi (changes requested):

docs/example-react-router.mdx
thanhsonng: @MatanBobi We are. In the README's `Documentation` section, it specifies that

>Version 4 is used in React Router versions 4 and 5.

This sentence is what I want to reference to. Do you think we can use the Link with highlight feature of Chrome, something like this?

https://github.com/remix-run/history#documentation:~:text=Version%204%20is%20used%20in%20React%20Router%20versions%204%20and%205.

docs/example-react-router.mdx
thanhsonng: I removed the line

docs/example-react-router.mdx
MatanBobi: Got it, thanks :)

docs/example-react-router.mdx
MatanBobi: No worries, thanks :)

MatanBobi (approved):
Thanks! :)

MatanBobi:
@all-contributors please add @thanhsonng for docs :)

PR #1116, opened 2022-07-14
Svelte: Update documentation for current best practices

ObieMunoz:
The original Testing Library/Svelte docs have not beed updated in quite a while and I believe the actual examples were written prior to `screen` being released in [DOM Testing Library v6.11.0](https://github.com/testing-library/dom-testing-library/releases/tag/v6.11.0). 

Best practice is to use the `screen` and not access the `container` directly for queries.

### DIFF
--- a/docs/queries/about.mdx
+++ b/docs/queries/about.mdx
@@ -373,8 +373,8 @@ screen.debug(screen.getAllByText('multi-test'))
 ### `screen.logTestingPlaygroundURL()`
 
 For debugging using [testing-playground](https://testing-playground.com), screen
-exposes this convenient method which logs and returns a URL that can be opened
-in a browser.
+exposes this convenient method which logs and returns a URL that can be opened in
+a browser.
 
 ```javascript
 import {screen} from '@testing-library/dom'
--- a/docs/queries/about.mdx
+++ b/docs/queries/about.mdx
@@ -188,7 +188,8 @@ React Testing Library re-export `screen` so you can use it the same way.
 Here's how you use it:
 
 <Tabs defaultValue="native" values={[ { label: 'Native', value: 'native', }, {
-label: 'React', value: 'react', }, { label: 'Cypress', value: 'cypress', }, ] }>
+label: 'React', value: 'react', }, { label: 'Cypress', value: 'cypress', },
+{ label: 'Svelte', value: 'svelte', }, ] }>
 <TabItem value="native">
 
 ```js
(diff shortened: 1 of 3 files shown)


docs/queries/about.mdx
```diff
@@ -223,6 +224,15 @@ const exampleInput = screen.getByLabelText('Example')
...
 cy.findByLabelText('Example').should('exist')
+```
+
+  </TabItem>
+  <TabItem value="svelte">
+
+```js
+import {render, screen} from '@testing-library/svelte'
+
+const exampleInput = screen.getByLabelText('Example')
```
timdeschryver, line 235: Should this render the component first?
I would also prefer to revert this change to keep this consistent, or all the examples in the core api should include a svelte example. If you want to, feel free to create a follow-up PR with the changes.

docs/svelte-testing-library/api.mdx
```diff
@@ -30,6 +31,32 @@ import {render} from '@testing-library/svelte'
 const {results} = render(YourComponent, {ComponentOptions}, {RenderOptions})
 ```
 
+## `screen`
```
timdeschryver, line 34: To keep the docs consistent we should also revert this change because no other libraries include it in their docs.
`screen` is not Svelte specific and is already documented in the core api, ideally we don't want to introduce duplicates.

docs/svelte-testing-library/api.mdx
```diff
@@ -128,8 +155,8 @@ import {render, fireEvent} from '@testing-library/svelte'
 import Comp from '..'
 
 test('count increments when button is clicked', async () => {
-  const {getByText} = render(Comp)
-  const button = getByText('Count is 0')
+  render(Comp)
+  const button = screen.getByText('Count is 0')
```
timdeschryver, line 159: `screen` needs to be imported

timdeschryver (changes requested):
Hi thanks for the PR.
I added some remarks to keep things consistent.
Let me know what you think

docs/queries/about.mdx
ObieMunoz: I reverted this and will follow-up with a future PR to add svelte to all of the examples in the core api.

docs/svelte-testing-library/api.mdx
ObieMunoz: That makes sense, I reverted this.

docs/svelte-testing-library/api.mdx
ObieMunoz: Thanks, updated!

timdeschryver (approved):
Thanks @ObieMunoz !

timdeschryver:
@all-contributors please add @ObieMunoz  for docs

PR #1122, opened 2022-07-22
docs: remove await on userEvent.upload

jhsu:
upload is not async and should not be awaited


ph-fritsche:
All APIs return a `Promise` which should be awaited.

jhsu:
Ah sorry, i was on v13, which had a typing of returning void for upload

PR #1128, opened 2022-08-04
docs(user-event): add missing return type

ph-fritsche:


ph-fritsche:
@all-contributors add @kksy docs

PR #1143, opened 2022-08-22
docs: reintroduce import/require usage for user-event

nickmccurdy:
It seems like when the docs for user-event 14 were released, the docs on import/require usage were removed. However, this reintroduces issues like https://github.com/testing-library/user-event/issues/485 if you don't already know how to destructure the `default` export from the CJS module. I fixed this by copying my docs from user-event 13 (https://github.com/testing-library/user-event/pull/495).

### DIFF
--- a/docs/user-event/install.mdx
+++ b/docs/user-event/install.mdx
@@ -26,6 +26,16 @@ yarn add --dev @testing-library/user-event
 </TabItem>
 </Tabs>
 
+Now simply import it in your tests:
+
+```js
+import userEvent from '@testing-library/user-event'
+
+// or
+
+const {default: userEvent} = require('@testing-library/user-event')
+```
+
 Note that `@testing-library/user-event` requires `@testing-library/dom`.
 
 If you use one of the


ph-fritsche (approved):
It feels redundant. But if it helps someone, it probably doesn't hurt to add it.

nickmccurdy:
I know, but without this there isn't any mention on how to safely use user-event with CJS, and since this issue doesn't apply to other Testing Library packages, I don't think it would be obvious.

If you have any suggestions on how to simplify this while still safely importing user-event, I'm open to it. Otherwise I think it would be a good idea to merge.

docs/user-event/install.mdx
```diff
@@ -26,6 +26,16 @@ yarn add --dev @testing-library/user-event
...
 
+Now simply import it in your tests:
+
+```js
+import userEvent from '@testing-library/user-event'
+
+// or
+
+const {default: userEvent} = require('@testing-library/user-event')
+```
```
ph-fritsche, line 37: Maybe we should use `Tabs` here too. (Backslash to make the suggestion display.)
```suggestion
<Tabs defaultValue="esm" values={[
    { label: 'In EcmaScript Module', value: 'esm' },
    { label: 'In CommonJS', value: 'cjs' },
]}>
<TabItem value="esm">
  ```js
  import userEvent from '@testing-library/user-event'
  \```
</TabItem>
<TabItem value="cjs">
  ```js
  const {default: userEvent} = require('@testing-library/user-event')
  \```
</TabItem>
</Tabs>
```

docs/user-event/install.mdx
```diff
@@ -26,6 +26,16 @@ yarn add --dev @testing-library/user-event
...
 </Tabs>
 
+Now simply import it in your tests:
+
+```js
+import userEvent from '@testing-library/user-event'
+
+// or
+
+const {default: userEvent} = require('@testing-library/user-event')
```
eps1lon, line 36: Shouldn't this be automatically happen when you write `import userEvent from '@testing-library/user-event'` and let it transpile by Babel?
This seems like a step backwards regarding progress of ES6 modules in the ecosystem so I would encourage `user-event` to get this right. You probably want to ship ES6 modules and CommonJS in `module` and `main` entrypoints respectively.

nickmccurdy: > Shouldn't this be automatically happen when you write `import userEvent from '@testing-library/user-event'` and let it transpile by Babel?

Yes, but not everyone uses Babel. [Testing Library Recorder Extension](https://github.com/testing-library/testing-library-recorder-extension) needs to use CJS modules for now because it can't introduce configuration to enable Babel or Jest's experimental ES modules (which in my experience do not work well with that project's dependencies). I edited the documentation to fix this previously, but it wasn't included in the new docs for version 14.

> You probably want to ship ES6 modules and CommonJS in `module` and `main` entrypoints respectively.

user-event does do this. Are you suggesting this is a build configuration issue with user-event?

docs/user-event/install.mdx
nickmccurdy: I was thinking it would be nice to show both variants by default in case users don't understand the difference between EcmaScript and CommonJS modules without seeing code. I also can't seem to get this to parse (event without the backslashes).

PR #1144, opened 2022-08-25
docs: add reference to `applyAccept` from `upload()`

khitrenovich:
See https://github.com/testing-library/user-event/issues/1046#issuecomment-1227819119 for related discussion.

### DIFF
--- a/docs/user-event/api-utility.mdx
+++ b/docs/user-event/api-utility.mdx
@@ -149,6 +149,8 @@ upload(
 
 Change a file input as if a user clicked it and selected files in the resulting
 file upload dialog.
+Files that don't match an `accept` property will be automatically discarded,
+unless [`applyAccept`](options.mdx#applyaccept) is set to `false`.
 
 ```jsx
 test('upload file', async () => {


docs/user-event/api-utility.mdx
```diff
@@ -149,6 +149,8 @@ upload(
 
 Change a file input as if a user clicked it and selected files in the resulting
 file upload dialog.
+Files that don't match an `accept` property will be automatically discarded,
+unless [`applyAccept`](options.mdx#applyaccept) is set to `false`.
```
timdeschryver, line 153: Suggestion: can we add this in an [Admonitions](https://docusaurus.io/docs/2.0.0-rc.1/markdown-features/admonitions)?

```suggestion
:::info
Files that don't match an `accept` property will be automatically discarded,
unless [`applyAccept`](options.mdx#applyaccept) is set to `false`.
:::
```

khitrenovich: We can, but I don't see it being used anywhere in the User Interactions area. Instead, I see quoting (`>`) is being used to mark notes, likely being the less visually-intrusive alternative. I can convert it to admonition, to a quote or just leave it as is - the decision would happen between you and @ph-fritsche , I guess.

timdeschryver: Actually, now that I've seen it on the preview website it looks good.
The reason why I mentioned it was because I would see myself looking at the docs exactly because files might be discarded and I wanted to know why. Hence, I thought it would be better that it would stand out. But it looks good now already.
@ph-fritsche you got the final say here :)

timdeschryver (approved):

docs/user-event/api-utility.mdx
ph-fritsche: I would make this a new paragraph. The other sections start with a very short description of what the API does. Often just the name of the API used in a proper sentence.
This might not make for the best novel read, but it is easy to follow when skim reading.

The admonitions stick out. I'd reserve them for information that doesn't "fit in the story" but is too important to be just placed somewhere else.

But I'm fine with it either way.

khitrenovich: I agree with what I read above - people will come to see the docs "exactly because files might be discarded", so the note should be "easy to follow when skim reading", but not "stick out". I think quoting makes the best balance, and it is already extensively used in the other paragraphs of the same file. I've made the adjustment, LMK if that looks better. I'm happy to tweak it further if needed.

ph-fritsche (approved):

timdeschryver:
@all-contributors please add @khitrenovich  for docs

PR #1148, opened 2022-09-11
update docs to describe usage of COLORS env variable

MynockSpit:
Update docs for https://github.com/testing-library/dom-testing-library/pull/1166.

NOTE: This is written in the same style as the docs for the other environment variable. I'd be happy to change one or both of those sections to some other format provided I'm given an example of a pattern to follow.

### DIFF
--- a/docs/dom-testing-library/api-debugging.mdx
+++ b/docs/dom-testing-library/api-debugging.mdx
@@ -5,8 +5,8 @@ title: Debugging
 
 ## Automatic Logging
 
-When you use any `get` calls in your test cases, the current state of the
-`container` (DOM) gets printed on the console. For example:
+When any `get` or `find` calls you use in your test cases fail, the current
+state of the `container` (DOM) gets printed on the console. For example:
 
 ```javascript
 // <div>Hello world</div>
@@ -26,7 +26,7 @@ Here is the state of your container:
 </div>
 ```
 
-Note: Since the DOM size can get really large, you can set the limit of DOM
+**Note**: Since the DOM size can get really large, you can set the limit of DOM
 content to be printed via environment variable `DEBUG_PRINT_LIMIT`. The default
 value is `7000`. You will see `...` in the console, when the DOM content is
 stripped off, because of the length you have set or due to default size limit.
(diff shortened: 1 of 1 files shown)


docs/dom-testing-library/api-debugging.mdx
```diff
@@ -40,6 +40,20 @@ This works on macOS/Linux, you'll need to do something else for Windows. If
 you'd like a solution that works for both, see
 [`cross-env`](https://www.npmjs.com/package/cross-env).
 
+**Note**: The output of the DOM are colorized by default if you tests are
+running in a node environment. However, you may sometimes wish to turn off
+colors, such as in cases where the output is being written to a log file for
+debugging purposes. You can use the environment variable `COLORS` to explicitly
+force the colorization off or on. For example:
```
timdeschryver, line 47:
```suggestion
**Note**: The output of the DOM is colorized by default if your tests are
running in a node environment. However, you may sometimes want to turn off
colors, such as in cases where the output is written to a log file for
debugging purposes. You can use the environment variable `COLORS` to explicitly
force the colorization off or on. For example:
```

docs/dom-testing-library/api-debugging.mdx
```diff
@@ -40,6 +40,20 @@ This works on macOS/Linux, you'll need to do something else for Windows. If
...
+
+```
+COLORS=false npm test
+```
+
+As with `DEBUG_PRINT_LIMIT`, `COLORS` works in macOS/Linux but you'll need to do
+something else for Windows. If you'd like a solution that works for both, see
+[`cross-env`](https://www.npmjs.com/package/cross-env).
```
timdeschryver, line 55: Sorry for the additional comment.
I think we can just copy paste this note, what do you think?

```suggestion
This works on macOS/Linux, you'll need to do something else for Windows. If
you'd like a solution that works for both, see
[`cross-env`](https://www.npmjs.com/package/cross-env).
```

MynockSpit: Yeah, sure.

timdeschryver (approved):
Thanks!

eps1lon (approved):
Thanks!

eps1lon:
@all-contributors add @MynockSpit for docs

PR #1151, opened 2022-10-03
docs: add Vitest as alternative to Jest

rwieruch:
Vitest as alternative to Jest is becoming pretty popular these days. When I updated my book The Road to React, I couldn't find any guide for this setup though. So I wrote a brief guide as reference for my book and hopefully for this documentation as well.

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -12,7 +12,7 @@ there are some things you can do when configuring your testing framework to
 reduce some boilerplate. In these docs we'll demonstrate configuring Jest, but
 you should be able to do similar things with
 [any testing framework](#using-without-jest) (React Testing Library does not
-require that you use Jest).
+require that you use Jest). For example, using [React Testing Library with Vitest](https://www.robinwieruch.de/vitest-react-testing-library/) is a popular alternative to using it with Jest.
 
 ## Global Config
 


docs/react-testing-library/setup.mdx
```diff
@@ -12,7 +12,7 @@ there are some things you can do when configuring your testing framework to
 reduce some boilerplate. In these docs we'll demonstrate configuring Jest, but
 you should be able to do similar things with
 [any testing framework](#using-without-jest) (React Testing Library does not
-require that you use Jest).
+require that you use Jest). For example, using [React Testing Library with Vitest](https://www.robinwieruch.de/vitest-react-testing-library/) is a popular alternative to using it with Jest.
```
timdeschryver, line 15: Hi, the linked article is good and explains how to do it.
But I think we prefer to "inline" that content into our docs instead of referring to an external website/blog.
For example, similar to the [Svelte Testing Library docs](https://testing-library.com/docs/svelte-testing-library/setup#vitest).

Before closing this PR I'll wait until a second maintainer gives a review.

MatanBobi: I agree.. I think that blog posts tend to get outdated quite quickly and we aim to keep stuff inside or docs so we'll have the ability to update them :)

timdeschryver: I closed this, but feel free to open a PR with a "copy-paste" of the linked article.

PR #1157, opened 2022-10-13
chore: update deps

timdeschryver:
Update docusaurus
Also updated lint-staged (the result of running `npm audit fix`)
Replaced husky with a simple githook, because the `pre-commit` didn't execute


package.json
```diff
@@ -15,39 +15,24 @@
     "swizzle": "docusaurus swizzle",
     "clear": "docusaurus clear",
     "deploy": "docusaurus deploy",
-    "docusaurus": "docusaurus"
+    "docusaurus": "docusaurus",
+    "pre-commit": "lint-staged",
+    "prepare": "git config core.hookspath .githooks"
```
MatanBobi, line 20: I guess that this is here as a replacement for husky.. I'm just wondering why husky didn't work.. 
In fact, we were using version 1 and husky is already on version 8 I think.

MatanBobi (approved):
Nice, just a tiny comment :)

package.json
timdeschryver: Yea, if you want I can take a look into it.
IIRC husky v8 you also need to register the hooks in a `prepare` script, that's why I prefer to do it "manually" and loose the dependency.

MatanBobi: Yeah, I think you're right and Husky v7^ requires running a `prepare` script.
No problem :)

eps1lon, line 20: This is usually not required for git hooks. The problem here is that now I wouldn't know how to disable git hooks on a per repo basis. I usually change the hooks path to something I own and then symlink the hooks that I do find usefull. 

pre-commit hooks are something I will definitelly disable. So if you want to keep this as a pre-commit hook and not something else (e.g. pre-push), please add docs on how to disable this hook for this particular repository.

eps1lon, line 20: This is usually not required for git hooks. The problem here is that now I wouldn't know how to disable git hooks on a per repo basis. I usually change the hooks path to something I own and then symlink the hooks that I do find usefull. 

pre-commit hooks are something I will definitelly disable. So if you want to keep this as a pre-commit hook and not something else (e.g. pre-push), please add docs on how to disable this hook for this particular repository.

eps1lon (changes requested):

package.json
```diff
@@ -19,35 +19,24 @@
...
-      "*.mdx": [
-        "prettier --write",
-        "git add"
-      ],
-      "*.json": [
-        "prettier --write ",
-        "git add"
-      ]
-    }
+    "*.js": "prettier --write",
```
timdeschryver, line 27: Doing `git add` isn't required anymore - https://github.com/okonet/lint-staged#v10 

> From v10.0.0 onwards any new modifications to originally staged files will be automatically added to the commit. If your task previously contained a git add step, please remove this. The automatic behaviour ensures there are less race-conditions, since trying to run multiple git operations at the same time usually results in an error.

package.json
timdeschryver: 👍 That's fair - AFAIK this can't support your use case.
Updating husky also isn't an option then because they also register the hookspath this way (from the docs: "Powered by modern new Git feature (core.hooksPath)").

I reverted the husky update to keep everything like it was as I currently don't see an easy way to disable certain hooks.

timdeschryver: And reverted it back to the origin version

package.json
MatanBobi: Cool 👍

MatanBobi (approved):

package.json
timdeschryver: @eps1lon could you give this another review please?

PR #1158, opened 2022-10-18
feat: support zh-Hans

xupea:
1. support i18n
2. support zh-Hans（简体中文）


MatanBobi:
Hi @xupea!
Thanks for working on this.
It might take me some time to review this one but I'll put it on my list :)

timdeschryver (commented):
Before we review this, I would like to hear more opinions if we should add more languages to the docs or not.
Because this will also have an impact (read take time) in the future.

If the majority approves, some "ground rules" would be useful to add to the contributing guidelines.

MatanBobi:
I believe translations is a good idea and 100% agree that we should add ground rules for the contribution guidelines and talk about it a bit before merging :)

alexkrolick:
Docusaurus has good i18n support. Although it's likely there will be some divergence in the docs for different languages, it seems like a good thing to do. It would be nice to have a "translation lead" for each language that stays subscribed to this repo and tries to keep up with changes to the docs as the APIs change. Community contributions also welcome.

nazarepiedady:
@MatanBobi @alexkrolick

What happened to the translation initiative for the project?

I think this will increase the usage and access of the
non-native English speakers and programmers 
to the current project.

MatanBobi:
@nazarepiedady I didn't get the chance to review this one. I'll do my best to get to it soon :)

nazarepiedady:
@MatanBobi we need this feature as soon as possible,
because this will increase the number of its users, mainly
those that are non-English natives.

nickmccurdy:
Do we need to move English into `i18n` too?

nazarepiedady:
@nickmccurdy accordingly to the preview of the website, it seems to
work well without it, but to keep consistency in the project and also to
ease the understanding of the codebase I believe it is better to move
the English version to `i18h` too if it is possible to do it.

PR #1162, opened 2022-10-19
docs: add sveltekit update and alt config

MentalGear:
I struggled setting this up with sveltekit, and was only able to get it working with the added `mergeConfig`

### DIFF
--- a/docs/svelte-testing-library/setup.mdx
+++ b/docs/svelte-testing-library/setup.mdx
@@ -75,8 +75,28 @@ with any testing framework and runner you're comfortable with.
     ```  
 ### SvelteKit
 
-To use Vitest with SvelteKit install `vitest-svelte-kit`, which includes a preconfigured Vitest configuration for SvelteKit projects.
-You can take a look at the [`vitest-svelte-kit` configuration docs](https://github.com/nickbreaton/vitest-svelte-kit/tree/master/packages/vitest-svelte-kit#configuring) for further instructions.
+Previously `vitest-svelte-kit` could be used, but it is not necessary anymore and depreceated since `svelte-kit` is now a vite plugin.
+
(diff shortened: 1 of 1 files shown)


docs/svelte-testing-library/setup.mdx
```diff
@@ -75,8 +75,28 @@ with any testing framework and runner you're comfortable with.
     ```  
 ### SvelteKit
 
-To use Vitest with SvelteKit install `vitest-svelte-kit`, which includes a preconfigured Vitest configuration for SvelteKit projects.
-You can take a look at the [`vitest-svelte-kit` configuration docs](https://github.com/nickbreaton/vitest-svelte-kit/tree/master/packages/vitest-svelte-kit#configuring) for further instructions.
+Previously `vitest-svelte-kit` could be used, but it is not necessary anymore and depreceated since `svelte-kit` is now a vite plugin.
+
+If you struggle getting this working with the above vitest configuration, use the `mergeConfig` instead in `vitest.config.ts`:
```
timdeschryver, line 80: Could we reword this in a sentence that explains how to use `mergeConfig`?
Imho, mentioning `vitest-svelte-kit` is not important because it's deprecated, and it isn't needed.

Thoughts?

timdeschryver: Hey @MentalGear what do you think of this suggestion?

MentalGear: LGTM !

timdeschryver: Would you then like to update this PR with the proposed changes? 🙂

mcous:
@timdeschryver I was perusing the open PRs list for Svelte stuff, and I do not think this PR is necessary anymore after #1389

timdeschryver:
Thanks @mcous !

PR #1168, opened 2022-10-30
docs: fix api docs rtl

robertoms99:
Error in renderHooks example, must be return the props object instead of a string


eps1lon (approved):

eps1lon:
@all-contributors add @robertoms99 for docs

PR #1173, opened 2022-11-05
Update api-keyboard.mdx

lsoares:
https://testing-library.com/docs/user-event/keyboard/ still does not show the \ could have it been escaped by the markdown processor?


lsoares:
does not solve it

PR #1189, opened 2022-12-13
"findByRole" query over "waitFor and getByRole"

shawnesquivel:
By implementing tis change, the eslint suggestion "Prefer `findByRole` query over using `waitFor` + `getByRole" is removed. The change implements the recommended suggestions by these two articles, which is the same as the automatic command line "quickfix." [1] https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-find-by.md
 [2] https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

### DIFF
--- a/docs/react-testing-library/example-intro.mdx
+++ b/docs/react-testing-library/example-intro.mdx
@@ -93,7 +93,7 @@ test('loads and displays greeting', async () => {
 
   fireEvent.click(screen.getByText('Load Greeting'))
 
-  await waitFor(() => screen.getByRole('heading'))
+  await screen.findByRole('heading')
 
   expect(screen.getByRole('heading')).toHaveTextContent('hello there')
   expect(screen.getByRole('button')).toBeDisabled()


docs/react-testing-library/example-intro.mdx
```diff
@@ -93,7 +93,7 @@ test('loads and displays greeting', async () => {
 
   fireEvent.click(screen.getByText('Load Greeting'))
 
-  await waitFor(() => screen.getByRole('heading'))
+  await screen.findByRole('heading')
```
timdeschryver, line 96: The `waitFor` import on line 77 can be removed.

timdeschryver (commented):
Thanks!
Could you also update the other examples on this page?
For example on line 113, 139, and 213.

ledenis:
I think this could be closed in favor of https://github.com/testing-library/testing-library-docs/pull/1242

timdeschryver:
You're right, thanks @ledenis

PR #1194, opened 2022-12-26
docs: fix incorrect description of getByAltText

tnyo43:
issue: https://github.com/testing-library/testing-library-docs/issues/1190

`*ByAltText` query will not return for `` ([see sandbox](https://codesandbox.io/s/test-query-priority-mi4inx?file=/src/ByAltTest.test.tsx:329-436)).

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -62,11 +62,10 @@ oftentimes better to use `getByRole(expectedRole, { name: 'The name' })`. The
 accessible name query does not replace other queries such as `*ByAlt` or
 `*ByTitle`. While the accessible name can be equal to these attributes, it does
 not replace the functionality of these attributes. For example
-`<img aria-label="fancy image" src="fancy.jpg" />` will be returned for both
-`getByAltText('fancy image')` and `getByRole('img', { name: 'fancy image' })`.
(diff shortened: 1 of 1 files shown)


docs/queries/byrole.mdx
```diff
@@ -62,11 +62,10 @@ oftentimes better to use `getByRole(expectedRole, { name: 'The name' })`. The
 accessible name query does not replace other queries such as `*ByAlt` or
 `*ByTitle`. While the accessible name can be equal to these attributes, it does
 not replace the functionality of these attributes. For example
-`<img aria-label="fancy image" src="fancy.jpg" />` will be returned for both
-`getByAltText('fancy image')` and `getByRole('img', { name: 'fancy image' })`.
```
tnyo43, line 66: What I did is just removing this "`getByAltText('fancy image')`"

Maybe I could correct `` to `fancy image` as like [this comment](https://github.com/testing-library/testing-library-docs/issues/1190#issuecomment-1355409202). I made this change so that the following sentence makes sense.

eps1lon (approved):
Yep makes sense. We will probably keep the current implementation since `aria-label` and `alt` have different behavior so a dedicated query for just `alt` makes sense.

eps1lon:
@all-contributors add @tnyo43 for docs

PR #1195, opened 2022-12-28
docs: Update React setup.mdx with Vitest configuration instructions

pjchender:
Vitest is becoming a popular tool for starting a React project from scratch. However, there is still no documentation on how to use it. I felt it would be useful to update the React developer documentation with Vitest instructions.

Please let me know if the section needs to be reorganized or if further information is required.

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -491,3 +491,72 @@ And register it using mocha's `-r` flag:
 ```
 mocha -r ./mocha-watch-cleanup-after-each.js
 ```
+
+## Vitest
+
+When using the react-testing-library with [Vitest](https://vitest.dev/),
+you need to install jsdom and the @testing-library:
+
+```bash npm2yarn
+npm install --save-dev jsdom
+npm install --save-dev @testing-library/react @testing-library/jest-dom
+```
+
+Since the testing library uses jsdom, you must set the `environments`
+to `jsdom`.
+
+```js
+// vite.config.js
+export default defineConfig({
+  test: {
+    // highlight-next-line
+    environment: "jsdom",
+  },
+})
+```
+
+After installing the packages and setting the environment, you can start
+testing. However, some further steps might help you develop more efficiently.
+
+When utilizing testing-library APIs, you must first import
+`@testing-library/jest-dom` in every file. It's quite inconvenient.
+It is preferable to create a file that imports testing-library.
+
+```js
+// src/setupTests.js
+import '@testing-library/jest-dom'
+```
+
+By the config `setupFiles` in vitest, you can run it before each test file.
+
+```js
+// vite.config.js
+export default defineConfig({
+  test: {
+    environment: "jsdom",
+    // highlight-next-line
+    setupFiles: "src/setupTests.js"
+  },
+})
+```
+
+Now, you don't need to import testing-library in every test file.
+
+By default, vitest does not provide [global](https://vitest.dev/config/#globals) APIs
+for explicitness, so you have to import the `describe` or `test` from vitest
+when using them. If you want to use them globally without importing them,
+you can set `globals` to `true` in the `vite.config.js`.
+
+```js
+// vite.config.js
+export default defineConfig({
+  test: {
+    environment: "jsdom",
+    setupFiles: "src/setupTests.js"
+    // highlight-next-line
+    globals: true,
+  },
+})
+```
+
+Now, you don't need to import the `describe` and `test` manually from vitest.
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -340,7 +340,7 @@ using Create React App without TypeScript, save this to `jsconfig.json` instead.
 
 ### Jest 28
 
-If you're using Jest 28 or later, jest-environment-jsdom package now must be installed separately. 
+If you're using Jest 28 or later, jest-environment-jsdom package now must be installed separately.
 ```bash npm2yarn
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/setup.mdx
```diff
@@ -491,3 +491,72 @@ And register it using mocha's `-r` flag:
...
+
+When using the react-testing-library with [Vitest](https://vitest.dev/),
+you need to install jsdom and the @testing-library:
+
+```bash npm2yarn
+npm install --save-dev vitest jsdom
+npm install --save-dev @testing-library/react @testing-library/jest-dom
+```
+
+Since the testing library uses jsdom, you must set the `environments`
```
eps1lon, line 505: Neither `/react` nor `/dom` "use" JSDOM. But both libraries require **a** DOM implementation to be available. The config and recommendation for JSDOM is fine but the wording needs some tweaking to make sure people are aware that they could also use a Browser or other DOM implementations.

docs/react-testing-library/setup.mdx
```diff
@@ -491,3 +491,72 @@ And register it using mocha's `-r` flag:
...
+    environment: "jsdom",
+  },
+})
+```
+
+After installing the packages and setting the environment, you can start
+testing. However, some further steps might help you develop more efficiently.
+
+When utilizing testing-library APIs, you must first import
+`@testing-library/jest-dom` in every file. It's quite inconvenient.
```
eps1lon, line 522: For `/react` you don't **need** `@testing-library/jest-dom`. Are we also sure `@testing-library/jest-dom` library even works Vitest?

docs/react-testing-library/setup.mdx
```diff
@@ -491,3 +491,72 @@ And register it using mocha's `-r` flag:
...
+  },
+})
+```
+
+Now, you don't need to import testing-library in every test file.
+
+By default, vitest does not provide [global](https://vitest.dev/config/#globals) APIs
+for explicitness, so you have to import the `describe` or `test` from vitest
+when using them. If you want to use them globally without importing them,
+you can set `globals` to `true` in the `vite.config.js`.
```
eps1lon, line 548: Could you check if auto-cleanup works with Vitest with `globals`? We should document test cleanup for both config options.

eps1lon (commented):
Much appreciated! There are some details that we should clarify before merging though.

docs/react-testing-library/setup.mdx
pjchender: Thank you for pointing this out. I tweak the wording to make it more clear.

docs/react-testing-library/setup.mdx
pjchender: The auto-cleanup works as usual if we set the `globals` to `true`.

I also add a paragraph to describe how to do the cleanup and extend matchers manually when the `globals` is `false`.

docs/react-testing-library/setup.mdx
pjchender: Sorry, I am not sure if this is what you are asking. If we set the `globals` to `true`, then the matchers will be automatically extended when importing `@testing-library/jest-dom`. However, we must extend the matchers manually if we do not set it to `true`.

eps1lon: Sure but this document is about `@testing-library/react` not `@testing-library/jest-dom`. Let's split this out into a separate PR since I'm not sure jest-dom wants to support Vitest.

docs/react-testing-library/setup.mdx
```diff
@@ -491,3 +491,137 @@ And register it using mocha's `-r` flag:
...
+// vite.config.js
+export default defineConfig({
+  test: {
+    environment: "jsdom",
+    globals: true,
+    // highlight-next-line
+    setupFiles: "src/setupTests.js"
+  },
+})
+```
```
pjchender, line 604: @eps1lon do you suggest removing this paragraph from this PR?

eps1lon: Yep. jest-dom docs should handle this.

eps1lon: I think for Vitest you'd use [`vitest-dom`](https://github.com/chaance/vitest-dom) instead of `jest-dom`

pjchender: I don't think many people use `vitest-dom` for testing react components with Vitest. Additionally, without addressing the step of creating the `setupTests` file, it can be difficult to set up the testing environment for React. Please prove me wrong.

eps1lon: That's not really relevant. `jest-dom` is not for usage with Vitest. `vitest-dom` is. I don't see how the popularity of `vitest-dom` is relevant here?

pjchender: Thanks for your reply. I understand your points now. My documentation appears to be written for `@testing-library/jest-dom` but not for `@testing-library/react`, right?

Should I close this Pull Request here and create a new one for `@testing-library/jest-dom`? Since you mentioned that `jest-dom` is not meant to be used with Vitest, I'm considering whether I should create a Pull Request. I know that Vitest is compatible with Jest. Does this mean that Vitest is also compatible with jest-dom? What do you think?

eps1lon: Your PR is targetting `@testing-library/react`. There's no need to target `@testing-library/jest-dom` since that library is not supposed to be used with Vitest.

pjchender: @eps1lon thanks. I've removed the paragraph. Could you check it again?

PR #1204, opened 2023-01-14
Add docs for Web Testing Library

eps1lon:
Docs for the initial release. `waitFor` API is still in flux.


docs/web-testing-library/faq.mdx
```diff
@@ -0,0 +1,18 @@
...
+---
+
+<details>
+
+<summary>Why "Web" in "Web Testing Library</summary>
+
+This library implements testing utilities that are used on different clients
+implementing the
+[Minimum Common Web Platform API](https://proposal-common-min-api.deno.dev/) (or
+parts of it). This includes browsers, Node.js, Deno and React Native.
```
timdeschryver, line 13:
```suggestion
parts of it). This includes browsers, Node.js, Deno, and React Native.
```

docs/web-testing-library/api.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+`waitFor` may run the callback a number of times until the timeout is reached.
+Note that the number of calls is constrained by the `timeout` and `interval`
+options.
+
+This can be useful if you have a unit test that mocks API calls and you need to
+wait for your mock promises to all resolve.
+
+If you return a promise in the `waitFor` callback (either explicitly or
+implicitly with `async` syntax), then the `waitFor` utility will not call your
```
timdeschryver, line 46:
```suggestion
implicitly with the `async` syntax), then the `waitFor` utility does not call your
```

docs/web-testing-library/api.mdx
```diff
@@ -0,0 +1,57 @@
...
+callback again until that promise rejects. This allows you to `waitFor` things
+that must be checked asynchronously.
+
+The default `interval` is `50ms`. However it will run your callback immediately
```
timdeschryver, line 50:
```suggestion
The default `interval` is `50ms`. However, it runs your callback immediately
```

docs/web-testing-library/api.mdx
```diff
@@ -0,0 +1,57 @@
...
+before starting the intervals.
+
+The default `timeout` is `1000ms`.
+
+The default `onTimeout` takes the error and appends the `container`'s printed
+state to the error message which should hopefully make it easier to track down
+what caused the timeout.
```
timdeschryver, line 57: Isn't this specific for the DOM's implementation?

> and appends the `container`'s printed
state to the error message which should hopefully make it easier to track down
what caused the timeout.

docs/web-testing-library/install.mdx
```diff
@@ -0,0 +1,80 @@
...
+that is referenced in the `package.json` `engines.node` field. But the `module`
+and `umd:main` files are configured to compile down to support browsers as old
+as IE 10.
+
+<!--
+Links
+-->
+
+[npm]: https://www.npmjs.com/
+[node]: https://nodejs.org
```
timdeschryver, line 80: Is not used here, can be removed.

```suggestion
```

docs/web-testing-library/intro.mdx
```diff
@@ -0,0 +1,49 @@
...
+for which they are intended. As part of this, you want your testbase to be
+maintainable in the long run so refactors of your components (changes to
+implementation but not functionality) don't break your tests and slow you and
+your team down.
+
+## This solution
+
+The `Web Testing Library` is a very light-weight solution for testing code that
+runs on the Web Platform (a browser, [Node.js][node], [Deno](https://deno.land/)
+etc) or React Native. The main utilities it provides involve querying the DOM
```
timdeschryver, line 21:
```suggestion
etc), or React Native. The main utilities it provides involve querying the DOM
```

docs/web-testing-library/api.mdx
eps1lon: I think this was copy&pasted so we want to backport these changes.

docs/web-testing-library/intro.mdx
```diff
@@ -0,0 +1,49 @@
...
+> The more your tests resemble the way your software is used, the more
+> confidence they can give you.
+
+We try to only expose methods and utilities that encourage you to write tests
+that closely resemble how your code is used on the Web Platform or React Native.
+
+Utilities are included in this project based on the following guiding
+principles:
+
+1. They should be usable on both the Web Platform **and** React Native
```
alexkrolick, line 35: The name "web testing library" is a bit confusing if it also works for React Native, which is definitely not the "web". Is the goal of this library to provide a library- and platform- agnostic API to replace some parts of the existing ecosystem?

eps1lon: The goal is not to be platform agnostic (see e.g. ´setTimeout´) but offer shared utilities that are used on the most common platforms

alexkrolick: I see. In that case I'd suggest naming it something that is less similar to the framework wrappers that implement a full element traversal API, and more similar to the util libraries like user-event.

eps1lon: React Native is now official part of the web-interoperable runtimes: https://runtime-keys.proposal.wintercg.org/#react-native

Going to adjust the wording to use "web interoperable" instad of just "web" to make it less similar.

docs/web-testing-library/api.mdx
```diff
@@ -0,0 +1,74 @@
...
+This can be useful if you have a unit test that mocks API calls and you need to
+wait for your mock promises to all resolve.
+
+If you return a promise in the `waitFor` callback (either explicitly or
+implicitly with the `async` syntax), then the `waitFor` utility does not call
+your callback again until that promise rejects. This allows you to `waitFor`
+things that must be checked asynchronously.
+
+The default `interval` is `50ms`. However, it runs your callback
+immediatelybefore starting the intervals.
```
timdeschryver, line 53:
```suggestion
immediately before starting the intervals.
```

timdeschryver (approved):

docs/web-testing-library/api.mdx
```diff
@@ -0,0 +1,74 @@
+---
+id: api
+title: API
+---
+
+Several utilities are provided for dealing with asynchronous code. These can be
+useful to wait for an element to appear or disappear in response to an event,
+user action, timeout, or Promise. (See the
+[guide to testing disappearance](guide-disappearance.mdx).)
```
eps1lon, line 9: Does this link work?

timdeschryver: This works, it redirects to https://deploy-preview-1204--testing-library.netlify.app/docs/guide-disappearance

docs/web-testing-library/intro.mdx
alexkrolick: Any reason not to just put `waitFor` in its own package?

MatanBobi:
@eps1lon should we also rename the `@testing-library/web` repo to `@testing-library/web-interoperable`? Want me to do it?

PR #1205, opened 2023-01-19
remove 'will' from docs

timdeschryver:
Coming from #1204
This PR back-ports my feedback on the new web docs that were partly copy-pasted from this page.


eps1lon (approved):
nice

PR #1213, opened 2023-02-05
docs: be more specific about current Angular's rerender behavior

shaman-apprentice:
ref #653

### DIFF
--- a/docs/angular-testing-library/api.mdx
+++ b/docs/angular-testing-library/api.mdx
@@ -411,8 +411,8 @@ expect(screen.getByTestId('name-value').textContent).toBe('Sarah')
 
 ### `rerender`
 
-Create and render a new instance of the component.
-To update the existing component use [change](#change).
+Create and render a new instance of the component. Input properties that are not
+defined are cleared. To keep them use [change](#change).
 
 ```typescript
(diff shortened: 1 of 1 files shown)


docs/angular-testing-library/api.mdx
```diff
@@ -411,8 +411,8 @@ expect(screen.getByTestId('name-value').textContent).toBe('Sarah')
 
 ### `rerender`
 
-Re-render the same component with different props. Input properties that are not
-defined are cleared. This calls `detectChanges` after the props are updated.
+Create and render a new instance of the component.
```
timdeschryver, line 414: I would like to keep a note about "Input properties that are not
defined are cleared.".

Now that I see this, this is also the reason that `change` and `rerender` exist.
With change you can't clear input properties.

https://github.com/testing-library/angular-testing-library/issues/252

shaman-apprentice: I updated the PR. Please let me know what you think. I also created https://github.com/testing-library/angular-testing-library/issues/365 for further discussion. Thanks again for your great fast responses :)

timdeschryver (approved):

timdeschryver:
@all-contributors please add @shaman-apprentice  for docs

PR #1214, opened 2023-02-06
ByRole: Remove MatchOptions

eps1lon:
Don't merge before https://github.com/testing-library/dom-testing-library/pull/1211 is released in `latest`

Docs for https://github.com/testing-library/dom-testing-library/pull/1211

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -15,13 +15,11 @@ import TabItem from '@theme/TabItem'
 getByRole(
   // If you're using `screen`, then skip the container argument:
   container: HTMLElement,
-  role: TextMatch,
+  role: string,
   options?: {
-    exact?: boolean = true,
     hidden?: boolean = false,
     name?: TextMatch,
     description?: TextMatch,
-    normalizer?: NormalizerFn,
     selected?: boolean,
     checked?: boolean,
     pressed?: boolean,


docs/queries/byrole.mdx
```diff
@@ -15,13 +15,11 @@ import TabItem from '@theme/TabItem'
...
+  role: string,
   options?: {
-    exact?: boolean = true,
     hidden?: boolean = false,
     name?: TextMatch,
     description?: TextMatch,
-    normalizer?: NormalizerFn,
     selected?: boolean,
     checked?: boolean,
     pressed?: boolean,
```
timdeschryver, line 25: While I was looking at the PR I noticed that the docs are missing the `suggest` option.
Could we add it here, or should I create a PR for this?

```suggestion
    pressed?: boolean,
    suggest?: boolean,
```

eps1lon: can you do this in a separate one? I don't actually know what this option does :D

timdeschryver (approved):
Approving already so this can be merged whenever the release will be

docs/queries/byrole.mdx
MatanBobi: @timdeschryver Do you want me to take it? I have some free time :)

timdeschryver: Sure, go for it @MatanBobi

PR #1229, opened 2023-03-08
docs: update Vitest instructions for jest-dom

unikitty37:
Add instructions on how to configure Vitest to import @testing-library/jest-dom automatically, to match the Vite instructions at 6.2.

### DIFF
--- a/docs/svelte-testing-library/setup.mdx
+++ b/docs/svelte-testing-library/setup.mdx
@@ -66,6 +66,26 @@ with any testing framework and runner you're comfortable with.
     ```js
     import '@testing-library/jest-dom';
     ```
+    
+    To have Vitest import this automatically in _every_ test file, add the import statement to a new file, then add the file to the `setupFiles` array in the `vitest.config.ts` (in this case, the file is `vitest.include.jsdom.ts` in the project root)
+    
+    ```js
+    import { defineConfig } from 'vitest/config';
+    import { svelte } from '@sveltejs/vite-plugin-svelte';
+
+    export default defineConfig({
+        plugins: [svelte({ hot: !process.env.VITEST })],
+        test: {
+            include: ['src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
+            globals: true,
+            environment: 'jsdom',
+            setupFiles: [
+                './vitest.include.jsdom.ts'
+            ]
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):

docs/svelte-testing-library/setup.mdx
```diff
@@ -66,6 +66,26 @@ with any testing framework and runner you're comfortable with.
     ```js
     import '@testing-library/jest-dom';
     ```
+    
+    To have Vitest import this automatically in _every_ test file, add the import statement to a new file, then add the file to the `setupFiles` array in the `vitest.config.ts` (in this case, the file is `vitest.include.jsdom.ts` in the project root)
```
timdeschryver, line 70: Maybe this should be added under 5.3 ?

eps1lon (changes requested):
Why would you use a library specifically for Jest in Vitest? We already have an assertion library specifically for Vitest called `vitest-dom`. Could we clarify first if the folks at `@testing-library/jest-dom` want to support Vitest in addition to Jest?

unikitty37:
> Why would you use a library specifically for Jest in Vitest? We already have an assertion library specifically for Vitest called [`vitest-dom`](https://github.com/chaance/vitest-dom). Could we clarify first if the folks at `@testing-library/jest-dom` want to support Vitest in addition to Jest?

I was just updating the current documentation, which uses jest-dom. I didn't even know about `vitest-dom` — but it doesn't seem to be compatible with the current vitest version:

```
╰─▷ npm i -D vitest-dom
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: frontend@0.0.1
npm ERR! Found: vitest@0.25.8
npm ERR! node_modules/vitest
npm ERR!   dev vitest@"~0.25.3" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer vitest@"^0.16.0" from vitest-dom@0.0.4
npm ERR! node_modules/vitest-dom
npm ERR!   dev vitest-dom@"*" from the root project
```

unikitty37:
Since `vitest-dom` doesn't work with current vitest, and my comment pointing this out doesn't seem to have resulted in any response, I'm going to assume this contribution is unwanted at this point.

PR #1231, opened 2023-03-09
feat(ByRole): Allow filter by disabled state

eps1lon:
Docs for https://github.com/testing-library/dom-testing-library/pull/1221
Preview pending

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -25,6 +25,7 @@ getByRole(
     pressed?: boolean,
     suggest?: boolean,
     current?: boolean | string,
+    disabled?: boolean,
     expanded?: boolean,
     queryFallbacks?: boolean,
     level?: number,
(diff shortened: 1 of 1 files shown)


docs/queries/byrole.mdx
```diff
@@ -164,6 +165,28 @@ the "👎" by calling `getByRole('link', { current: false })`. To learn more abo
...
+You can filter elements by their disabled state by setting `disabled: true` or
+`disabled: false`.
+
+For example in
+
+```html
+<body>
+  <section>
+    <button disabled>Login</button>
+    <div role="alert" aria-disabled="false">Failed to login</button>
```
MatanBobi, line 179: Having the `aria-disabled` here seems a bit confusing to me. 
Are we also filtering by `aria-disabled` or just by the `disabled` attribute? I think that we should explain it here so people will have a clear understanding.

eps1lon: We're filtering by things that are disabled according to the a11y tree. This applies to every filter option in `ByRole`.

This here includes examples.

How would you explain it do avoid confusions?

docs/queries/byrole.mdx
```diff
@@ -164,6 +165,28 @@ the "👎" by calling `getByRole('link', { current: false })`. To learn more abo
...
+  </section>
+</body>
+```
+
+you can get the "Login" button by calling
```
MatanBobi, line 184:
```suggestion
You can get the "Login" button by calling `getByRole('button', { disabled: true })` and the alert by calling `getByRole('alert', { disabled: false })`.
```

docs/queries/byrole.mdx
MatanBobi: I'm thinking about something like this maybe?
https://github.com/testing-library/testing-library-docs/pull/1231/files#r1148318567

PR #1141, opened 2022-08-16
chore: update vue api doc to match implementation

davidnixon:
Fix for https://github.com/testing-library/vue-testing-library/issues/280

PR #1234, opened 2023-03-10
docs(vue): remove extra callbackFunction parameter

rydash:
Teeny tiny cleanup from the overhaul done in #1141.

This callback used to provide a way to install additional plugins, but now that's handled by [the `global.plugins` array from `@vue/test-utils` mount options.](https://test-utils.vuejs.org/api/#global)


timdeschryver (approved):

timdeschryver:
@all-contributors please add @rydash  for docs

PR #1244, opened 2023-04-06
doc: Add a hooks/rerender example with multiple props

sravan-s:
The rerender example for hooks was a bit hard to understand
Hope this example helps

### DIFF
--- a/docs/react-testing-library/api.mdx
+++ b/docs/react-testing-library/api.mdx
@@ -392,7 +392,6 @@ describe('useMyGreetingHook', () => {
     expect(result.current).toEqual('Hello sam and bob')
   })
 })
-
 ```
 
 > NOTE: When using `renderHook` in conjunction with the `wrapper` and `initialProps` options, the `initialProps` are not passed to the `wrapper` component.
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/api.mdx
```diff
@@ -370,6 +370,30 @@ test('returns logged in user', () => {
...
+If you have multiple props, try the following recipie
+```jsx
+function useMyGreetingHook(userA, userB) {
+  const greeting = React.useMemo(() => {
+    return 'Hello ' + userA + ' and ' + userB
+  }, [userA, userB])
+  return greeting
+}
+
+describe('useMyGreetingHook', () => {
```
timdeschryver, line 382: Instead of creating a new example, wouldn't it be simpler to add the example to the existing snippet?
E.g. with the following:

```
rerender({name: 'Bob'})
expect(result.current).toEqual({name: 'Bob'})
```

sravan-s: cleaning up original example is fine too

ps. Examples in hooks section are a bit difficult to discern :(  
https://github.com/testing-library/react-hooks-testing-library/issues/743

PR #1248, opened 2023-04-18
Move explanation about findBy combination to Single Elements

vinzid:
The explanation about `findBy` combination is more related to `Single Elements` instead of `Multiple Elements`, so move it upwards.


MatanBobi:
Hi @vinzid!
Thanks for taking the time to open this one.
Since `findBy` and `findAllBy` works the same and they both rely on `waitFor` as you can see [here](https://github.com/testing-library/dom-testing-library/blob/main/src/query-helpers.ts#L243), I think this change will be misleading and make people think this is relevant only for single elements.

I'm closing this one but thanks again :)

PR #1251, opened 2023-05-01
fix wrong role

lsoares:
value is not supported in headings
and the example is about role='spinbutton' anyway so it's a mistake.


timdeschryver (approved):
Thanks!

PR #1256, opened 2023-05-24
Add imports to `user-event` intro documentation

LLCampos:
It took me a while to understand how to import `userEvent` and even to understand that it was a separate library. 😶‍🌫️ 

I think adding these imports would make it clearer for users that get into the page and quickly scroll for the code examples.

### DIFF
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -59,6 +59,8 @@ itself - e.g. in a `before`/`after` hook - for reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
 
 ```js
+import userEvent from '@testing-library/user-event'
+
 // inlining
 test('trigger some awesome feature when clicking the button', async () => {
   const user = userEvent.setup()
@@ -71,6 +73,8 @@ test('trigger some awesome feature when clicking the button', async () => {
 ```
 
 ```js
+import userEvent from '@testing-library/user-event'
+
 // setup function
 function setup(jsx) {
   return {
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -61,6 +61,7 @@ itself - e.g. in a `before`/`after` hook - for reasons described in
 ```js
 import userEvent from '@testing-library/user-event'
 import { render, screen } from '@testing-library/vue'
+import MyComponent from './MyComponent.vue'
 
 // inlining
 test('trigger some awesome feature when clicking the button', async () => {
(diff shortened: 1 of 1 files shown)


docs/user-event/intro.mdx
```diff
@@ -59,6 +59,8 @@ itself - e.g. in a `before`/`after` hook - for reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
 
 ```js
```
timdeschryver, line 62: I agree with the comments.
On a similar note, I think adding imports to `screen` and `MyComponent` can also be included.

LLCampos: Added imports to render and screen, but unsure what to do about MyComponent - I'm not a React person :)

ph-fritsche: I'm not sure we should include the `render` import here, as it (re)introduces references to specific framework libraries.
A code comment might be better:
```jsx
  return {
    user: userEvent.setup(),
    // Import `render` from the framework library of your choice.
    // See https://testing-library.com/docs/dom-testing-library/install#wrappers
    ...render(jsx),
  }
```

LLCampos: If we do that for both `screen` and `render` we then end up with almost more comments than code? :)

Should we just have the import for `userEvent` since it is what this page is about?

ph-fritsche: > then end up with almost more comments than code? :)

Not necessarily a bad thing – especially in docs :grin:

LLCampos: Just added the suggested comments :)

docs/user-event/intro.mdx
```diff
@@ -60,33 +60,37 @@ itself - e.g. in a `before`/`after` hook - for reasons described in
...
+    // Import `render` from the framework library of your choice.
+    // See https://testing-library.com/docs/dom-testing-library/install#wrappers
     ...render(jsx),
   }
 }
 
 test('render with a setup function', async () => {
-  const {user} = setup(<MyComponent />)
+  // Import `setup` from the framework library of your choice.
+  // See https://testing-library.com/docs/dom-testing-library/install#wrappers
```
ph-fritsche, line 92: `setup` is declared above in the same snippet.

LLCampos: Oops. This is what happens when you make these edits in a rush. Fixed.

timdeschryver (approved):
Thanks!

ph-fritsche (approved):

timdeschryver:
@all-contributors please add @LLCampos  for docs

PR #1257, opened 2023-05-25
docs(user-event): add missing default value for `skipHover`

caiangums:
The value is already `false` in the source code. This change just specify it in the doc.
* respective file: https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/src/setup/setup.ts#LL31C3-L31C12


timdeschryver (approved):

timdeschryver:
@all-contributors please add @caiangums  for docs

PR #1267, opened 2023-06-26
Document mutations within waitFor not being recommended

frosas:
See https://github.com/testing-library/dom-testing-library/issues/808#issuecomment-721737410 for some context


eps1lon:
It's not quite clear from the thread why this is a bad idea. The docs should have at least one concrete example illustrating why. Is this only because we have a MutationObserver

MatanBobi:
@frosas Are you planning on updating this PR or can we close it?

frosas:
The original documentation is good enough so I'm closing it.

PR #1268, opened 2023-07-07
feat: add FAQ for svelte-testing-library

yanick:


MatanBobi:
Hi @yanick! Thanks for taking the time to open this one.
I'm not sure this is worth an FAQ page since it's probably data that can be found in an issue.
I'm closing this, thanks again and sorry.

yanick:
True. But not everybody will think of searching the issues, let alone the closed issues, of the repo. Is there any downside to put that answer in the FAQ? I mean, it's not like it's currently overflowing with other Qs....

MatanBobi:
A simple search in google will probably raise the answer. The downside to putting something in the docs is that we need to maintain it and keep it updated, at the moment it's only one FAQ, but if we don't see a reason to create an FAQ besides that, I don't think we'd want to burden of maintaining it :)

yanick:
I would expect that the maintainer of svelte-testing-library has the job to keep the documention up-to-date. And that'd be me.

But okay, FAQ in the project's README it is. :shrug:

MatanBobi:
> I would expect that the maintainer of svelte-testing-library has the job to keep the documention up-to-date. And that'd be me.
> 
> But okay, FAQ in the project's README it is. 🤷

Sorry, I missed the part that you're maintaining it :)

yanick:
> If you're good with maintaining this area too, I have no objections to push this.

That was my expectation, yup. :-)

MatanBobi:
@all-contributors please add @yanick for docs :)

PR #1271, opened 2023-07-11
Docs: add Sunhat to list of users

ali-kamalizade:
We at Sunhat have been using Testing Library (Angular) from the start.

PR #1274, opened 2023-07-12
Docs: fix Sunhat user image not shown

ali-kamalizade:
https://github.com/testing-library/testing-library-docs/pull/1271 added a new user. Unfortunately, I noticed that the SVG was not rendering since it was using relative width and height. This PR fixes my mistake from the previous PR.

image


timdeschryver (approved):
Looks good now, thanks for the follow-up @ali-kamalizade 
image

PR #1280, opened 2023-07-18
docs(angular): update rerender api

dzonatan:
based on changes made in https://github.com/testing-library/angular-testing-library/pull/366


timdeschryver (approved):
Thanks @dzonatan !
I also saw the GH issue conversation, which I will take a closer look at later this week.

PR #1283, opened 2023-07-21
docs: userEvent/convenience Example Code

ssi02014:
I wrote the example code for the "user-event/Convenience APIs".

[user-event v13](https://testing-library.com/docs/ecosystem-user-event)  I think it would be helpful for more people to have example code like on the page.

Except for "hover" and "unhover", this is based on the [user-event v13](https://testing-library.com/docs/ecosystem-user-event) example code.

The example code all passes the tests fine.

스크린샷 2023-07-21 오후 10 04 52

### DIFF
--- a/docs/user-event/api-convenience.mdx
+++ b/docs/user-event/api-convenience.mdx
@@ -182,13 +182,15 @@ keyboard('[/ShiftLeft][/ShiftRight]{Tab}')
 test('tab', async () => {
   render(
     <div>
-      <input data-testid="element" type="checkbox" />
-      <input data-testid="element" type="radio" />
-      <input data-testid="element" type="number" />
+      <input type="checkbox" />
+      <input type="radio" />
+      <input type="number" />
     </div>,
   )
 
-  const [checkbox, radio, number] = screen.getAllByTestId('element')
+  const checkbox = screen.getByRole('checkbox')
+  const radio = screen.getByRole('radio')
+  const number = screen.getByRole('spinbutton')
 
   expect(document.body).toHaveFocus()
 
--- a/docs/user-event/api-convenience.mdx
+++ b/docs/user-event/api-convenience.mdx
@@ -100,6 +100,7 @@ pointer({target: element})
 ```js
 const TestComponent = () => {
   const [isHover, setIsHover] = useState(false)
+
   return (
     <div
       style={{backgroundColor: isHover ? 'red' : 'green'}}
@@ -111,42 +112,7 @@ const TestComponent = () => {
   )
 }
 
-test('hover', async () => {
-  render(<TestComponent />)
-
-  const hoverBox = screen.getByText('Hover')
-
-  await userEvent.hover(hoverBox)
-
-  expect(hoverBox).toHaveStyle({backgroundColor: 'red'})
-})
-```
-
-### unhover()
-
-```ts
-unhover(element: Element): Promise<void>
-```
-
-```js
-pointer({target: element.ownerDocument.body})
-```
-
-```js
-const TestComponent = () => {
-  const [isHover, setIsHover] = useState(false)
-  return (
-    <div
-      style={{backgroundColor: isHover ? 'red' : 'green'}}
-      onMouseEnter={() => setIsHover(true)}
-      onMouseLeave={() => setIsHover(false)}
-    >
-      Hover
-    </div>
-  )
-}
-
-test('unhover', async () => {
+test('hover/unhover', async () => {
   render(<TestComponent />)
 
   const hoverBox = screen.getByText('Hover')
@@ -159,6 +125,16 @@ test('unhover', async () => {
 
   expect(hoverBox).toHaveStyle({backgroundColor: 'green'})
 })
+```
+
+### unhover()
+
+```ts
+unhover(element: Element): Promise<void>
+```
+
+```js
+pointer({target: element.ownerDocument.body})
 ```
 
 ## Keyboard
--- a/docs/user-event/api-convenience.mdx
+++ b/docs/user-event/api-convenience.mdx
@@ -20,13 +20,16 @@ pointer([{target: element}, {keys: '[MouseLeft]', target: element}])
 
 ```js
 test('click', async () => {
+  const onChange = jest.fn()
   const user = userEvent.setup()
-  render(<input type="checkbox" />)
+
+  render(<input type="checkbox" onChange={onChange} />)
 
   const checkbox = screen.getByRole('checkbox')
 
   await user.click(checkbox)
 
+  expect(onChange).toHaveBeenCalledTimes(1)
   expect(checkbox).toBeChecked()
 })
 ```
@@ -45,13 +48,16 @@ pointer([{target: element}, {keys: '[MouseLeft][MouseLeft]', target: element}])
 
 ```js
 test('double click', async () => {
+  const onChange = jest.fn()
   const user = userEvent.setup()
-  render(<input type="checkbox" />)
+
+  render(<input type="checkbox" onChange={onChange} />)
 
   const checkbox = screen.getByRole('checkbox')
 
   await user.dblClick(checkbox)
 
+  expect(onChange).toHaveBeenCalledTimes(2)
   expect(checkbox).not.toBeChecked()
 })
 ```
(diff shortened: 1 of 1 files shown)


ssi02014:
Hi, @timdeschryver 👋
I wanted to contribute to that repository, so I created this pull request.

Is my work appropriate for that repository? 
I'd appreciate it if you could take a look

docs/user-event/api-convenience.mdx
```diff
@@ -18,6 +18,20 @@ click(element: Element): Promise<void>
...
+  const onChange = jest.fn()
+  render(<input type="checkbox" onChange={onChange} />)
+
+  const checkbox = screen.getByRole('checkbox')
+
+  await userEvent.click(checkbox)
+
+  expect(onChange).toHaveBeenCalledTimes(1)
+  expect(checkbox).toBeChecked()
+})
```
timdeschryver, line 32: Instead of using `userEvent` directly, we should set it up first and create a `user`.
This was a change between v13 and v14, and these examples should also be updated with it.

```suggestion
test('click', async () => {
  const user = userEvent.setup()
  const onChange = jest.fn()
  render(<input type="checkbox" onChange={onChange} />)

  const checkbox = screen.getByRole('checkbox')

  await user.click(checkbox)

  expect(onChange).toHaveBeenCalledTimes(1)
  expect(checkbox).toBeChecked()
})
```

docs/user-event/api-convenience.mdx
```diff
@@ -55,6 +97,36 @@ hover(element: Element): Promise<void>
...
+}
+
+test('hover/unhover', async () => {
+  render(<TestComponent />)
+
+  const hoverBox = screen.getByText('Hover')
+
+  await userEvent.hover(hoverBox)
+
+  expect(hoverBox).toHaveStyle({backgroundColor: 'red'})
```
timdeschryver, line 122: Could you update this expectation please?
For example, to update the text of the `div` element instead.
The reason for this is that I personally don't like the `toHaveStyle` expectations, and IIRC some of the testing environments don't support this.

timdeschryver (changes requested):
Thanks for your contribution @ssi02014 !
What do you think of it @ph-fritsche ?

docs/user-event/api-convenience.mdx
ssi02014: @timdeschryver 
Thank you for your review. I modified the example code as you guided me.

https://testing-library.com/docs/user-event/utility
I initially referenced that page, and it's written as if it uses "userEvent" directly.
May I create a Pull Request after making modifications to this page?

docs/user-event/api-convenience.mdx
ssi02014: @timdeschryver 
Oh, thank you.
I have modified the example code according to your comments.

ph-fritsche:
I'm not sure about this. Examples bloat the docs and it is difficult to keep them concise *and* accurate.
Someone who understands all the implications without extra comments might not need an example at all, whereas someone who needs an example might be misled to believe everything in the example is best practice.
Misunderstandings will be perpetuated by quoting these examples on other platforms/channels like Stackoverflow.

```js
// The title doesn't matter for the example, but in a real project it should tell about the tested behavior.
test('click', async () => {
  // Yes, using `.setup()` is recommended and for a few tests this can be inlined,
  // but if certain steps like calling `render()` and `userEvent.setup()` are repeated throughout a test suite
  // we also recommend to use your own setup function that includes all these steps for better readability.
  const onChange = jest.fn()
  render(<input type="checkbox" onChange={onChange} />)

  const checkbox = screen.getByRole('checkbox')

  await user.click(checkbox)

  // Here we assert that UserEvent, the framework library and the DOM implementation do work.
  // The code excerpts we see on Stackoverflow, Discord, etc. often include many assertions like these
  // and often enough they lack assertions on behavior implemented in the subject under test.
  expect(onChange).toHaveBeenCalledTimes(1)
  expect(checkbox).toBeChecked()
})
```

If we want to include (more) examples, maybe we should create a small example project that really adheres to best practices.

ssi02014:
@ph-fritsche @timdeschryver 
Thank you for your review.

> but if certain steps like calling `render()` and `userEvent.setup()` are repeated throughout a test suite

I write independent test code for each case, which is why I had the problem you describe above. 

> If we want to include (more) examples, maybe we should create a small example project that really adheres to best practices.

This is actually the best.

I sympathize with your comments as well.

I found the [user-event v13](https://testing-library.com/docs/ecosystem-user-event) page very helpful when I first started using "userEvent", and I wrote the following pull request based on my experience.

I figured this might help others as much as it helped me.

https://testing-library.com/docs/user-event/utility/
Also, the example code was already written on that page, so I thought it would be a good fit. 

I'd like you to draw your own conclusions about the pull request. 🙏
I intend to follow your opinion.

timdeschryver:
Thanks for the feedback @ph-fritsche .

I agree with your points, and I think that having a small example in the docs that covers the usage is helpful (and acts as a starting point), that's why we should keep it at a minimum.

The proposed examples project is a good idea, but it's a bit out of scope for this PR (but can be a reason to close this PR).

Is this something that we can all agree with?
An example for `click` might look as follows, where it just verifies its behavior.

```ts
test('click', async () => {
  render(<input type="checkbox" />)
  
  const checkbox = screen.getByRole('checkbox')
  await user.click(checkbox)

  expect(checkbox).toBeChecked()
})
```

ssi02014:
@timdeschryver 
Yes, I agree. 🙏
I think it's good to keep the example code to a minimum, as it can help people like me who are just starting out.

ssi02014:
@timdeschryver @ph-fritsche 
I've modified the click example code to make it a bit more streamlined.

docs/user-event/api-convenience.mdx
```diff
@@ -55,6 +94,37 @@ hover(element: Element): Promise<void>
 pointer({target: element})
 ```
 
+```js
+const TestComponent = () => {
```
timdeschryver, line 98: Sorry I'm being a pain, but I think it would be better to revert this to the previous example.
The reason for this is to be framework/library-agnostic, this is now a react example.
I can't think of any other test for now 😅

The rest of the changes are looking good! Thanks for that.

ssi02014: @timdeschryver 
That's fine, thanks for your review.
I reverted to the previous code. 🙏

timdeschryver: I think I was not being clear on what I meant.
The test for hover can be reverted to use `toHaveStyle`.

The rest of the tests were good how they were, but I see that `onChange` was added back (which was not the intention.

ssi02014: @timdeschryver 
I'm sorry. I made a mistake. 🙏
I've modified the code again.

ssi02014: @timdeschryver 
I understood you to remove the "hover/unhover" example code, is that correct?

timdeschryver: Not really, I can't find it the full snippet anymore because the commits are overwritten with the force push.
But I meant the initial example that was using `toHaveStyle`, or was this example always using a react component?

From our earlier conversation:

```
expect(hoverBox).toHaveStyle({backgroundColor: 'red'})
```

ssi02014:
```
expect(hoverBox).toHaveStyle({backgroundColor: 'red'})
```
@timdeschryver 
I apologize for the confusion caused by the force push.
This example also uses a React Component.

ssi02014: @timdeschryver 
What about an example like this?
This is a test code for "hover/unhover" without using any React code.

```js
test('hover/unhover', async () => {
  render(<div>hover</div>)

  const hoverBox = screen.getByText('hover')
  let isHover = false;

  hoverBox.addEventListener('mouseover', () => {
    isHover = true;
  })
  hoverBox.addEventListener('mouseout', () => {
    isHover = false;
  })

  await user.hover(hoverBox)

  expect(isHover).toBeTruthy()

  await user.unhover(hoverBox)

  expect(isHover).toBeFalsy()
})
```

ssi02014:
@timdeschryver 
After thinking about it, I think it would be better to leave onChange for click, double click, and triple click and check how many times the onChange function has been called.

This would make the example code clearer as to whether it was a click, double click, or triple click.

"expect(checkbox).toBeChecked()" is not enough.

docs/user-event/api-convenience.mdx
timdeschryver: Oh my bad!
I'm fine with the second example using the event listener.
Thanks for following up!

ssi02014:
@timdeschryver
I have modified the code again.
Thanks again for your detailed review.

timdeschryver (approved):

ssi02014:
@timdeschryver 
I'm waiting for that pull request to be reflected 😂😂😂 
Maybe we need another maintainer to confirm?

timdeschryver:
@all-contributors please add @ssi02014  for docs

PR #1287, opened 2023-08-14
fix(a11y): reducing homepage accessibility issues by 64%

StephNathai:
## Description
- adding alt attributes to images for accessibility compliance
- tags that are for decorative purposes have an empty alt attribute

## Before
before

## After
after


MatanBobi (approved):
Hi @StephNathai!
Thanks for taking the time to doing this :)

MatanBobi:
@all-contributors please add @StephNathai for code :)

PR #1294, opened 2023-08-25
fix: use correct control modifier

Rafatcb:
I was creating a test and `{Ctrl}` wasn't working. I searched through the code and found that the correct modifier is`{Control}`, as you can see in [user-event/tests/keyboard/parseKeyDef.ts](https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/tests/keyboard/parseKeyDef.ts#L30) and [user-events/tests/utility/type.ts](https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/tests/utility/type.ts#L72).

### DIFF
--- a/docs/ecosystem-user-event.mdx
+++ b/docs/ecosystem-user-event.mdx
@@ -177,14 +177,14 @@ The following special character strings are supported:
 | `{home}`       | Home       | N/A                |                                                                                                                                                                     |
 | `{end}`        | End        | N/A                |                                                                                                                                                                     |
 | `{shift}`      | Shift      | `shiftKey`         | Does **not** capitalize following characters.                                                                                                                       |
-| `{ctrl}`       | Control    | `ctrlKey`          |                                                                                                                                                                     |
+| `{control}`    | Control    | `ctrlKey`          |                                                                                                                                                                     |
 | `{alt}`        | Alt        | `altKey`           |                                                                                                                                                                     |
 | `{meta}`       | OS         | `metaKey`          |                                                                                                                                                                     |
 | `{capslock}`   | CapsLock   | `modifierCapsLock` | Fires both keydown and keyup when used (simulates a user clicking their "Caps Lock" button to enable caps lock).                                                    |
 
-> **A note about modifiers:** Modifier keys (`{shift}`, `{ctrl}`, `{alt}`,
+> **A note about modifiers:** Modifier keys (`{shift}`, `{control}`, `{alt}`,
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @Rafatcb! Thanks for taking the time to open this one.
Just verifying, you're editing the user-event v13 page, are you using user-event v13?
The code references you've posted here are true for v14, I can't find an answer for v13 - @ph-fritsche is this pr correct? :)

Rafatcb:
Hi @MatanBobi.

Sorry, I didn't realize it was v13. I'm using the latest version (`14.4.3`). I didn't find a documentation for the special characters for v14. I found only the documentation that I suggested editing.

Although the code is for the current version (v14), I checked [Git Blame](https://github.com/testing-library/user-event/blame/2c5d9f1369301539826879f95c9f6093f676e2bb/tests/keyboard/parseKeyDef.ts#L24) and the [commit](https://github.com/testing-library/user-event/commit/f251d153d55e43ec84c6012fa09352b251133f20) is from [PR \#581](https://github.com/testing-library/user-event/pull/581), which went into the [v13 release](https://github.com/testing-library/user-event/releases?q=%23581&expanded=true).

MatanBobi:
Nice backtracking :)
Let's just wait for @ph-fritsche to approve that this indeed is correct for v13.

ph-fritsche:
The `keyboardMap` was introduced in `v13.0.0` per https://github.com/testing-library/user-event/pull/581 – this includes the support for `{Control}`.
The support for `{Ctrl}` was removed in `v14.0.0` per https://github.com/testing-library/user-event/pull/783.

MatanBobi:
Thanks @ph-fritsche :)
If that's the case, do you think that the v13 page should be with `{Ctrl}` as it is today or with `{Control}`?

ph-fritsche:
I think we should keep it as it is. The table is for the (legacy) modifiers which had separate implementations in earlier versions. 

The following sentence is only true for `{ctrl}` but not for `{control}`:
https://github.com/testing-library/testing-library-docs/blob/19ee5deb2375a3f18e0abd6a44fc2bb6bdfb62ff/docs/ecosystem-user-event.mdx#L185-L187

MatanBobi:
If that's the case, I think we can close this one. Thanks @Rafatcb for the effort.
Having said that, following this PR it looks like users are looking for a similar table for the current version of user-event :)

PR #1298, opened 2023-08-26
svelte: update docs

timdeschryver:


docs/svelte-testing-library/intro.mdx
```diff
@@ -40,5 +40,3 @@ Introduction][dom-solution-explainer] for a more in-depth explanation.
 
 1.  A test runner or framework.
 2.  Specific to a testing framework.
-
-We recommend Vitest as our preference.
```
timdeschryver, line 44: This felt odd to add it here under `This Solution - What this library is not` (this is also mentioned in `setup`).
I can revert this if that's preferable

docs/svelte-testing-library/setup.mdx
```diff
@@ -4,85 +4,83 @@ title: Setup
...
-            include: ['src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-            globals: true,
-            environment: 'jsdom'
-        }
-    });
+4. Add a `vitest.config.ts` configuration file to the root of your project
 
-    ```
+   ```js
+   import {defineConfig} from 'vite'
```
timdeschryver, line 52: Updated using https://github.com/vitest-dev/vitest/tree/main/examples/svelte as an example

docs/svelte-testing-library/setup.mdx
```diff
@@ -4,85 +4,83 @@ title: Setup
...
-6.  Create your component and a test file (checkout the rest of the docs to see how)
-    and run the following command to run the tests.
+   ```js
+   import '@testing-library/jest-dom'
+   ```
 
-    ```
-    npm run test
-    ```  
-### SvelteKit
```
timdeschryver, line 76: Removed this because the rest of the sections are about setting up test runners.
I added a sveltekit section to the FAQ.
With the current state, everything works per default with SvelteKit.
Thus, the steps that were added to install `vitest-svelte-kit` can be removed.

docs/svelte-testing-library/setup.mdx
```diff
@@ -4,85 +4,90 @@ title: Setup
...
-1.  Install Vitest and jsdom
 
-    ```
-    npm install --save-dev vitest jsdom
-    ```
+1. Install Vitest and jsdom
 
-2.  Add the following to your `package.json`
+We're using `jdom` here as the test environment, but you can use any other
+options e.g `happy-dom`.
```
timdeschryver, line 16: added `npm2yarn` to all bash scripts

docs/svelte-testing-library/setup.mdx
```diff
@@ -4,85 +4,90 @@ title: Setup
...
 
-    export default defineConfig({
-        plugins: [svelte({ hot: !process.env.VITEST })],
-        test: {
-            include: ['src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-            globals: true,
-            environment: 'jsdom'
-        }
-    });
+2. To compile the Svelte components before using them in Vitest, you need to
```
MatanBobi, line 40: The word "to" seems redundant to me here, but maybe that's just me :)

docs/svelte-testing-library/setup.mdx
```diff
@@ -4,85 +4,90 @@ title: Setup
...
+     plugins: [svelte({hot: !process.env.VITEST})],
+     test: {
+       globals: true,
+       environment: 'jsdom',
+     },
+   })
+   ```
 
-    5.2 import `@testing-library/jest-dom` at the start of your test files
+4. Optionally install [jest-dom](https://github.com/testing-library/jest-dom) to
```
MatanBobi, line 64: I'm hesitating about this phrasing since the go-to solution for `vitest` is probably `vitest-dom` and not `jest-dom`... Wdyt?

MatanBobi (approved):
This looks like a great improvement, thanks Tim :) I left two minor comments

As a general idea, it seems to me like we're covering some stuff here that should be covered in other library documentations (that's why we need to update it from time to time when API changes). 
I'm not sure I have a good idea on how to solve it. As I said, this PR is an improvement from what we have so anyways I think it can be merged.

docs/svelte-testing-library/setup.mdx
timdeschryver: Yea, I had the same thought and that's when I reached out to you earlier.
I didn't update it because it was originally `jest-dom`, and it isn't maintained by us.

docs/svelte-testing-library/setup.mdx
timdeschryver: I think it's fine.

PR #1306, opened 2023-09-13
fix: typo of numbering

snjro:


timdeschryver:
@all-contributors please add @snjro  for docs

PR #1314, opened 2023-09-28
fix documentation: fix angular challenge url

tomalaforge:
I created a site for my angular challenges. There is a page specific for testing purposes and nicer than a readme file.


timdeschryver (approved):
🤩 !

PR #1323, opened 2023-10-20
Docs: Updated UI of Page[Help]

JoyelJohny:
Before:
Screenshot 2023-10-20 at 19-08-46 Testing Library Testing Library

After:
Screenshot 2023-10-20 at 19-07-25 Testing Library Testing Library

### DIFF
--- a/src/pages/help.js
+++ b/src/pages/help.js
@@ -61,7 +61,9 @@ export default function Help(props) {
         <div className="mainContainer documentContainer postContainer helpContainer">
           <div>
             <header>
-              <h1>Need help?</h1>
+              <h1>
+                <p className="headings">Need help?</p>
+              </h1>
             </header>
             <GridBlock
               contents={supportLinks.slice(0, 3)}
@@ -74,7 +76,9 @@ export default function Help(props) {
               align="left"
             />
             <section>
-              <h2>Buy a Course</h2>
+              <h1>
+                <p className="headings">Buy a course</p>
+              </h1>
               <p>
                 Learn how to test JavaScript with{' '}
                 <ExternalLink href="https://kentcdodds.com">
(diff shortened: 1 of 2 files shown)


alexkrolick (approved):

alexkrolick:
@MatanBobi looks pretty good to me, what do you think?

src/pages/help.js
```diff
@@ -61,7 +61,9 @@ export default function Help(props) {
         <div className="mainContainer documentContainer postContainer helpContainer">
           <div>
             <header>
-              <h1>Need help?</h1>
+              <h1>
+                <p className="headings">Need help?</p>
```
MatanBobi, line 65: AFAIR, putting a `p` inside `h1` is not recommended based on the [spec](https://www.w3.org/TR/2014/REC-html5-20141028/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements). Any chance to remove this? :)

src/pages/help.js
```diff
@@ -74,7 +76,9 @@ export default function Help(props) {
               align="left"
             />
             <section>
-              <h2>Buy a Course</h2>
+              <h1>
```
MatanBobi, line 79: As an idea, having more than 1 `h1` per page isn't recommended for accessibility reasons. I think this should remain an `h2` and if we have another `h1` somewhere, we should also change it to something else.

MatanBobi (changes requested):
Hi @JoyelJohny, thanks for taking the time to work on this.
As an idea, I like this change but I left a few comments regarding the implementation.

src/pages/help.js
JoyelJohny: Yes I can remove that, Is it okay to change `fontsize` of `h1` element.

MatanBobi: Yeah, that's alright :) thanks.

JoyelJohny:
If there is any need for changes ,Please tell me 😄

MatanBobi (approved):

MatanBobi:
That's great, thanks!

MatanBobi:
@all-contributors please add @JoyelJohny for code.

PR #1327, opened 2023-10-26
Update MSW usage in "example-intro"

kettanaito:
The only thing to watch out is that MSW requires Node.js 18 or later. If RTL requirements contradict this, we shouldn't update the examples.

### DIFF
--- a/docs/react-testing-library/example-intro.mdx
+++ b/docs/react-testing-library/example-intro.mdx
@@ -72,15 +72,15 @@ See the following sections for a detailed breakdown of the test
 
 ```jsx title="__tests__/fetch.test.jsx"
 import React from 'react'
-import {rest} from 'msw'
+import {httpm, HttpResponse} from 'msw'
 import {setupServer} from 'msw/node'
 import {render, fireEvent, screen} from '@testing-library/react'
 import '@testing-library/jest-dom'
 import Fetch from '../fetch'
 
 const server = setupServer(
-  rest.get('/greeting', (req, res, ctx) => {
-    return res(ctx.json({greeting: 'hello there'}))
+  http.get('/greeting', () => {
+    return HttpResponse.json({greeting: 'hello there'})
   }),
 )
 
@@ -101,8 +101,8 @@ test('loads and displays greeting', async () => {
 
 test('handles server error', async () => {
   server.use(
-    rest.get('/greeting', (req, res, ctx) => {
-      return res(ctx.status(500))
+    http.get('/greeting', () => {
+      return new HttpResponse(null, { status: 500 })
     }),
   )
 
@@ -132,7 +132,7 @@ test('handles server error', async () => {
 import React from 'react'
 
 // import API mocking utilities from Mock Service Worker
-import {rest} from 'msw'
+import {http, HttpResponse} from 'msw'
 import {setupServer} from 'msw/node'
 
 // import react-testing methods
@@ -161,9 +161,9 @@ component makes.
 // declare which API requests to mock
 const server = setupServer(
   // capture "GET /greeting" requests
-  rest.get('/greeting', (req, res, ctx) => {
+  http.get('/greeting', (req, res, ctx) => {
     // respond using a mocked JSON body
-    return res(ctx.json({greeting: 'hello there'}))
+    return HttpResponse.json({greeting: 'hello there'})
   }),
 )
 
@@ -181,8 +181,8 @@ test('handles server error', async () => {
   server.use(
     // override the initial "GET /greeting" request handler
     // to return a 500 Server Error
-    rest.get('/greeting', (req, res, ctx) => {
-      return res(ctx.status(500))
+    http.get('/greeting', (req, res, ctx) => {
+      return new HttpResponse(null, { status: 500 })
     }),
   )
 
--- a/docs/react-testing-library/example-intro.mdx
+++ b/docs/react-testing-library/example-intro.mdx
@@ -72,7 +72,7 @@ See the following sections for a detailed breakdown of the test
 
 ```jsx title="__tests__/fetch.test.jsx"
 import React from 'react'
-import {httpm, HttpResponse} from 'msw'
+import {http, HttpResponse} from 'msw'
 import {setupServer} from 'msw/node'
 import {render, fireEvent, screen} from '@testing-library/react'
 import '@testing-library/jest-dom'


kettanaito:
@MatanBobi, would it make sense to upgrade RTL itself? Node.js v14 and v16 are [EOL](https://nodejs.org/en/download/releases). I think it's best RTL migrates to the LTS, which is v18, which is also compatible with MSW 2.0.

MatanBobi:
> @MatanBobi, would it make sense to upgrade RTL itself? Node.js v14 and v16 are [EOL](https://nodejs.org/en/download/releases). I think it's best RTL migrates to the LTS, which is v18, which is also compatible with MSW 2.0. 

We have that planned as part of DTL major upgrade that's currently in alpha, once we'll merge it, we'll also bump this one :)

docs/react-testing-library/example-intro.mdx
```diff
@@ -72,15 +72,15 @@ See the following sections for a detailed breakdown of the test
 
 ```jsx title="__tests__/fetch.test.jsx"
 import React from 'react'
-import {rest} from 'msw'
+import {httpm, HttpResponse} from 'msw'
```
timdeschryver, line 75: Should this be `http`?

```suggestion
import {http, HttpResponse} from 'msw'
```

timdeschryver (commented):
Thanks for the update, and congrats with the release @kettanaito !

MatanBobi:
@kettanaito do you want me to put the old examples too? Since we're not yet enforcing Node 18, we should probably keep both examples there.

kettanaito:
@MatanBobi, I'm concerned that will be confusing for the reader. If RTL doesn't support Node.js 18 yet, leave the old examples but add a note that recommends using MSW 1.x because that's the last version compatible with Node.js 14-16. 

I wouldn't feature new examples because people can't use them anyway (RTL locks them to old versions of Node.js incompatible with MSW).

timdeschryver:
Because there are already some requests to update this I think we should merge this.
We can add a warning on this page that MSW requires Node.js v18+.
What do you think?

MatanBobi:
> Because there are already some requests to update this I think we should merge this. We can add a warning on this page that MSW requires Node.js v18+. What do you think?

Sounds like a good idea to me :)

kettanaito:
Great news! I've heard from people they were confused about the example featuring MSW 1. I'm so excited to see this updated! 🚀 

Also, thanks for the pointers on Node 18 support. I will keep that in mind if I get any questions about RTL + Node 18.

docs/react-testing-library/example-intro.mdx
```diff
@@ -72,15 +72,15 @@ See the following sections for a detailed breakdown of the test
```
timdeschryver, line 72: Initially I wanted to create a warning, but I noticed that we don't explicitly mention that the example uses MSW so I used a note.

```suggestion

:::note
The example uses [MSW (Mock Service Worker)](https://mswjs.io/) to mock HTTP requests.
MSW requires Node.js 18 or later.
:::
```

timdeschryver: Nevermind, it does mention it below the example.
Because I can't add a comment there, the lines at https://github.dev/kettanaito/testing-library-docs/blob/dbd489c8fa229dd43473daf7fd2d205c66fc3288/docs/react-testing-library/example-intro.mdx#L120-L122 should be removed.

```suggestion

:::note
We recommend using the [Mock Service Worker (MSW)](https://github.com/mswjs/msw) library
to declaratively mock API communication in your tests instead of stubbing
`window.fetch`, or relying on third-party adapters.

MSW requires Node.js 18 or later.
:::
```

MatanBobi:
Closing in favor of #1378. Thanks @kettanaito for working on this :)

kettanaito:
@skondrashov, we are trying to make the experience with Jest better through https://github.com/mswjs/jest-fixed-jsdom. Feel free to try it and let me know if it fixes the problems. This environment is meant to replace all the suggestions in the current docs, afaik. 

> My personal experience has been that even though msw seems like a very cool project, they more or less state that you shouldn't use it with jest for this reason, while RTL recommends jest, which leads to a pretty confusing experience when msw is recommended in the docs here, and I wish that the example had used something I could get up and running in a shorter amount of time, even if the old way of doing mocks isn't as good.

The confusion is understandable. I don't recommend Jest because it's an old tool that fails to adapt to the pace with which JavaScript evolves. That doesn't necessarily means MSW doesn't work in Jest. In fact, you have examples with [plain Jest](https://github.com/mswjs/examples/tree/main/examples/with-jest) as well as [Jest+JSDOM](https://github.com/mswjs/examples/tree/main/examples/with-jest-jsdom) working. Take a look at those and see what you have to adjust in your Jest config. 

> PS The polyfill doesn't even work, so I feel like msw is prohibitively complicated to use even if all of this was detailed in the docs.

Polyfill certainly works, and the `jest-fixed-jsdom` is largely based on it. The Jest issue you've linked is unlikely to be resolved. Jest and JSDOM are broken by design, as they will never rely on Node.js global APIs even if those are present in both the browser and Node.js. A valid JavaScript code fails in JSDOM. That looks like a broken library to me but when I reported it to the maintainers I was kindly told that "I don't understand the goal behind the project." 

I'm saying this to stress that neither of those tools considers this a bug so don't count on it ever getting resolved. This is why I recommend tools like Vitest and HappyDOM that manage to respect your globals better. They aren't perfect but they are years ahead in the testing framework field than alternatives.

kettanaito:
This also brings forth a far larger discussion of browser-like environments and how, imo, they have no place in 2024. You should be testing browser code in the browser. Browsers are far nicer to automate and cheaper to spawn than ever before. JSDOM was created 14 years ago to tackle terrible browser automation experience. It's not terrible anymore. In fact, it's quite fantastic. With both Playwright and Cypress investing in component-level testing, I hope to see the day when everybody will move away from hacks to actual in-browser testing.

skondrashov:
@kettanaito Thank you for the work you've been doing, and sorry for my late reply. I found the time to give RTL with MSW another shot using the new test environment, and I'm still unable to get the basic example to work. Both `import { setupServer } from 'msw/node'` as suggested by the RTL docs and `import { setupServer } from 'msw/browser'` as suggested by the msw github readme are invalid imports after `npm i msw` (which installs msw@2.3.1 as of this writing).

I agree largely with your points on using browsers rather than fake environments.

I disagree with the way you see the projects involved in this problem, because jsdom do in fact view not having every polyfill as a bug - their whole thing is trying to recreate a browser environment. Jest gives you a base javascript environment, and then the responsibility of jsdom is to make it browser-like. This makes perfect sense to me, and I feel like I'm aligned with the way the Jest developers see it and the way the jsdom developers see it. Yes, jsdom should provide definitions for browser features that msw relies on, but the jsdom developers agree with that: https://github.com/jsdom/jsdom/issues/2524#issuecomment-1935121424

The best thing I think I can offer you is my perspective as a mostly lazy person who only codes for work at my job. The overwhelming majority of people who are going to be trying to set this up are lazy people who only code for work at their job. They don't care about any of the things we're talking about, they want a document where they can follow the steps and then have tests that work at the end. Currently, it is impossible to use the RTL testing setup to get tests in a project. It was easy to do in 2020 before you touched the page. The work you have done might be philosophically sound, might be pushing the ecosystem ever forwards, is probably really good code, but none of that matters at all when you have to be a one in a thousand developer to even get the thing to run a basic test. Even if I'm doing it wrong somehow, I'm already using at least 3 separate documents to try to make this work and it simply isn't going to budge. My perspective is that I am never going to be able to see msw work, but that if the page was reverted to its 2020 form, I would have several days of my life back that I threw into the void because you replaced the working instructions with broken instructions. No matter how good the work you've been doing, or whether this new environment you created is good, the docs still don't work. I need the docs to work. That's all 99% of people are going to care about. In 2020, they worked. They should be reverted to that version. I will never find out if msw is any good unless they work, and any time anyone asks me about msw, I will tell them "there is no way to set up msw with jest, but if you look at the docs from 2020 you will find an example you can work with". My suggestion is to add "usable" to the list of things you want your library to be, on top of "clean", and "philosophically sound", and "cutting edge", and all of the other things that it likely is. And my suggestion to RTL is to remove any mention of msw from the docs until msw is usable on top of its other good qualities.

kettanaito:
@skondrashov, please use https://github.com/mswjs/jest-fixed-jsdom if you want to use MSW in Jest+JSDOM combo. MSW works fine in raw Jest. Here are a few reference examples:

- [MSW + Jest](https://github.com/mswjs/examples/tree/main/examples/with-jest)
- [MSW + Jest/JSDOM](https://github.com/mswjs/examples/tree/main/examples/with-jest-jsdom) (I believe this hasn't been updated to `jest-fixed-jsom` yet so you can drop `jest.polyfills.js` from this). 

> import { setupServer } from 'msw/browser' as suggested by the msw github readme

I don't believe this code snippet is ever mentioned anywhere. It's `from 'msw/node'`, that's the correct import. 

> Yes, jsdom should provide definitions for browser features that msw relies on, but the jsdom developers agree with that:

I have hard feelings about this. If you get `X is undefined` when referring to a standard API that has been available everywhere for years, it's a serious issue with your technology, regardless of its conceptual mission. 

> I think the overwhelming majority of people who are going to be trying to set this up are lazy people who only code for work at their job.

That's what we've published `jest-fixed-jsdom`. A one-liner for people who want to focus on doing their job.

skondrashov:
@kettanaito `msw/browser` import example is from the readme here: https://github.com/mswjs/msw

I did in fact use `jest-fixed-jsdom`, and though it got past all the polyfill errors, I just wasn't able to get the setupWorker import to work. I think the package addresses my concerns (the name could be more descriptive but that's minor), it's just that to make it work I have to reference the RTL setup docs, this thread, and several other documents across the internet along the way. I don't want to spend the time to figure out why I'm not able to import setupWorker right now but it could easily be something nobody else would run into. I'm just saying that ultimately all of these tools become useful to most people only once there is documentation to explain how to use them. It would be nice for the docs to be updated, and I'd be happy to try the steps that are laid out if you make a PR to the docs, I just don't want to jump around github and stackoverflow threads to debug this when the 2020 docs have already solved all my problems.

kettanaito:
> I just wasn't able to get the setupWorker import to work. I think the package addresses my concerns

Something is seriously mixed here. You should _never_ `setupWorker` in a Node.js process. Jest runs a Node.js process. Please use `setupServer` as it is advised [here](https://mswjs.io/docs/integrations/node). I think that will resolve all your issues. 

If you found that some information in MSW led you to believe you can use `setupWorker` in Node.js, please send it to me, I will fix it. I'm trying my best to make the environmental distinction as clear as possible. Thanks.

skondrashov:
You're right, I didn't notice that the examples use two different names. I went to the msw readme after being unable to get the RTL docs to work, and thought maybe that usage example would work. I certainly didn't think about whether setupWorker would work in Node.js. I believe I was getting the same "`setupWorker` is not an export" error for both, so I never got to the point where I could get runtime exceptions. Again, I'm sure this process works if you know what you're doing, but now you're linking me to yet another doc which isn't on the RTL site, and I just want the RTL docs to tell me how to use RTL.

PR #1329, opened 2023-10-28
docs: don't install `@testing-library/dom`

nickmccurdy:


MatanBobi:
AFAIU from the [code](https://github.com/testing-library/user-event/blob/main/package.json#L69), `@testing-library/dom` is still a peer dependency, doesn't that mean we still need that? I understand that if a user installs `@testing-library/react` for example they'll have it installed but if a user only installs `user-event`, this might still be needed don't you think?

nickmccurdy:
AFAIK the main reason we did this was because npm 3 stopped installing peer dependencies by default, but this has been reverted as of npm 7, so it should be equivalent in most cases. In other cases the package manager should will warn. I don't think we should be making this assumption about our users' dependencies anyway.

docs/user-event/install.mdx
```diff
@@ -25,12 +25,3 @@ yarn add --dev @testing-library/user-event
...
 </Tabs>
-
-Note that `@testing-library/user-event` requires `@testing-library/dom`.
-
-If you use one of the
-[framework wrappers](../dom-testing-library/install.mdx#wrappers), it is
-important that `@testing-library/dom` is resolved to the same installation
-required by the framework wrapper of your choice.  
-Usually this means that if you use one of the framework wrappers, you should not
-add `@testing-library/dom` to your project dependencies.
```
ph-fritsche, line 36: I think we should keep this note.
When people upgrade e.g. their version of `@testing-library/react` the `@testing-library/dom` package isn't always deduplicated. I'd rather add examples how to diagnose and fix this to further reduce retyped text when I suspect this to be the root of someeone's problem (e.g. "user-event only types the first N characters").

MatanBobi:
Since people don't really care about the warnings from the `npm i` script, maybe we can add a note about that for old users?

nickmccurdy:
A note about what?

docs/user-event/install.mdx
nickmccurdy: Are you suggesting we revert this deletion and add examples of specific use cases? Could you please give some examples? I'm a bit concerned about the complexity of supporting all this, and now I believe peer dependency errors are in a much better state than they used to be, so I'm not sure it's worth documenting.

MatanBobi:
> A note about what? Can you give an example?

That explicitly installing `@testing-library/dom` is still needed if you're using npm < 7. Since `user-event` doesn't have an engines definition and `@testing-library/react` for example still support node 14 which originally was bundled with npm 6.
That's just an idea, you can ignore it if you don't think it's needed :)

nickmccurdy:
Nobody is using `npm < 7`, we don't support this use case anymore, and it's also not documented in User Event's own readme.

nickmccurdy:
I'm also willing to discuss this on Discord if that would be easier for either of you.

docs/user-event/install.mdx
ph-fritsche: Yes, I would keep this paragraph on duplicate versions of `/dom`.
I'm under the impression this ranks as the second most common reason for `act` warnings and wrong input with UserEvent after missing `await`. Both isn't our fault and should not happen, but it does and we still have to deal with people running into these problems.
Therefore I'd like to have a section I can point to if I suspect this problem. The exact combinations of packages and package managers change over time and maintaining a list of them is out of scope of our docs. But if we feel that this paragraph isn't perspicuous, maybe we should add some explanation/example.

MatanBobi:
> Our `engines` are out of date because Node 16 support was recently dropped, but I'll make sure this is fixed soon and users should have already upgraded Node for security reasons anyway.
> 
> I'd really prefer to keep the original revert of my change (the first commit in this pull ignoring formatting changes) as it's now outdated advice and it's not even consistent with User Event's readme. If you have any strong opinions against this please say so, preferably with explicit examples of potentially problematic use cases.

We'll probably update those once the major version for DTL will merged.
I understand, I'm good with that :)

ph-fritsche (approved):

PR #1330, opened 2023-10-29
Highlights necessary step in Svelte setup

kyle-n:
@testing-library/svelte requires `globals: true` to run `cleanup()` after each test, but this is [a bit easy to miss in the docs](https://github.com/testing-library/testing-library-docs/issues/1331). This PR highlights this required step.

### DIFF
--- a/docs/preact-testing-library/api.mdx
+++ b/docs/preact-testing-library/api.mdx
@@ -54,6 +54,7 @@ Unmounts the component from the container and destroys the container.
 
 > This is called automatically if your testing framework (such as mocha, Jest or
 > Jasmine) injects a global `afterEach()` function into the testing environment.
+> If not, you will need to call `cleanup()` after each test.
 
 If you'd like to disable this then set `process.env.PTL_SKIP_AUTO_CLEANUP` to
 true when running your tests.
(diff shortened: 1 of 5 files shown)


MatanBobi:
Hi @kyle-n!
Thanks for taking the time to write this.
Any chance to align this explanation with what we have in `testing-libarry/react`?
You can see it [here](https://testing-library.com/docs/react-testing-library/api#cleanup).
It also feels like a rephrasing there can help to state that `globals: true` is needed if you want to take it :)

kyle-n:
@MatanBobi Thanks for the feedback! 

I rephrased the message under `cleanup()` to match the React documentation. I also shortened the message, which hopefully makes it easier to read, and copied it to the other frameworks that use `cleanup` and only run it with a global `afterEach`.

Does this phrasing work for `cleanup`?

> This is called automatically if your testing framework (such as mocha, Jest or Jasmine) injects a global `afterEach()` function into the testing environment.

kyle-n:
I thought that sentence was redundant, since the previous sentence implies if `cleanup()` is only run automatically with a global `afterEach()`, it is not run without one. However, on second thought, it is better to be explicit and clear. Updated.

MatanBobi (approved):
Thanks @kyle-n!

MatanBobi:
@all-contributors please add @kyle-n for docs :)

PR #1334, opened 2023-10-31
Update user event's delay option to say milliseconds instead of seconds

### LINKED ISSUE #1335: User event's delay value is incorrect documented as seconds
Delay is incorrectly documented as seconds instead of milliseconds. [See doc here.](https://testing-library.com/docs/user-event/options#delay)

bsheps:
The delay value gets passed into setTimeout which takes a value in milliseconds (not seconds).

fixes #1335


ph-fritsche (approved):

MatanBobi:
Thanks!

MatanBobi:
@all-contributors please add @bsheps for docs :)

PR #1344, opened 2023-11-24
chore: Improve docs with removing misleading statement on query API i…

vadimshvetsov:
Since in new versions of `@testing-library/cypress` we don't have `queryBy` and `queryAllBy` APIs its reasonable to remove them from docs.


MatanBobi (approved):
Wow, it has been like that since 2020, that's sad for us.
Thanks for picking it up!

MatanBobi:
@all-contributors please add @vadimshvetsov for docs :)

MatanBobi:
@all-contributors please add @vadimshvetsov for docs :)

PR #1346, opened 2023-12-02
docs: add redirect to new rtl template

MatanBobi:
This PR adds a new redirect to RTL template for reproductions
IMO, we can create more templates (DTL, ATL and more).
I can definitely take the DTL if we want.


timdeschryver (approved):
I like the idea! 🤩

PR #1349, opened 2023-12-06
docs: add query ByRole performance notes

agentdylan:
Adds notes about performance of *ByRole queries to the Queries > ByRole page.

`getByRole` and the likes are known to be slow at times and documentation has been requested[^1][^2][^3] to help make this more visible.

This has been known for several years[^1] and performance remains a concern[^4] so I believe it is worthwhile adding to the docs. 

I think it's important to have a balance of conveying the importance of `getByRole` while still sharing this information about the performance so users can make an informed decision if performance is a problem.

This relates to issues
- https://github.com/testing-library/dom-testing-library/issues/820
- https://github.com/testing-library/dom-testing-library/issues/698

These issues have great discussion around this topic, which I've attempted to consolidate.

[^1]: https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-726812393

[^2]: https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-726980009

[^3]: https://github.com/testing-library/dom-testing-library/issues/820#issuecomment-726994803

[^4]: https://github.com/testing-library/dom-testing-library/issues/698

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -462,3 +462,20 @@ You can query a specific element like this
 ```js
 getByRole('alertdialog', {description: 'Your session is about to expire'})
 ```
+
+## Performance
+
+`getByRole` is the most preferred query to use as it most closely resembles the
+user experience, however the calculations it must perform to provide this
+confidence can be expensive (particularly with large DOM trees).
+
+Where test performance is a concern it may be desirable to trade some of this
+confidence for improved performance.
+
+`getByRole` performance can be improved by setting the option
+[`hidden`](#hidden) to `true` and thereby avoid expensive visibility checks.
+Note that in doing so inaccessible elements will now be included in the result.
+
+Another option may be to substitute `getByRole` for simpler `getByLabelText` and
+`getByText` queries which can be significantly faster though less robust
+alternatives.


docs/queries/byrole.mdx
```diff
@@ -462,3 +462,20 @@ You can query a specific element like this
 ```js
 getByRole('alertdialog', {description: 'Your session is about to expire'})
 ```
+
+## Performance
```
timdeschryver, line 466: This looks good to me, but I'd like to keep it open for second reviewer.
One additional comment we might encourage is to use happy-dom as that should be faster.

timdeschryver (approved):

docs/queries/byrole.mdx
agentdylan: Sounds good, I'd be happy to add a note about happy-dom as I agree it is likely the best option if it works for people (I've seen some users report it isn't a viable option for them _yet_).

What do we think about adding something like this?

"Firstly, if you are using jsdom we'd encourage trying happy-dom as it is much more efficient. If this works for you it is your best option, improving performance while keeping test confidence.

Otherwise the suggestions below may help."

timdeschryver: Let's leave it like this for now, as this will likely break something.
We can always revisit it later.

timdeschryver:
@all-contributors please add @agentdylan for docs

docs/queries/byrole.mdx
cmacdonnacha: From my research, `happy-dom` is faster but lacks the fetureset of `JSDOM`, and that is reason folks often revert back to `JSDOM` after attempting a migration.

PR #1354, opened 2024-01-02
angular: document deferrable views

timdeschryver:
This PR adds documentation for a new API introduced in ATL.

### DIFF
--- a/docs/angular-testing-library/api.mdx
+++ b/docs/angular-testing-library/api.mdx
@@ -424,7 +424,7 @@ await rerender({componentInputs: {count: 7}, partialUpdate: true})
 
 // count is updated to 7
 expect(screen.getByTestId('count-value').textContent).toBe('7')
-// name is rendered because of the partial update
+// name is still rendered as "Sarah" because of the partial update
 expect(screen.getByTestId('name-value').textContent).toBe('Sarah')
 ```
 
--- a/docs/angular-testing-library/api.mdx
+++ b/docs/angular-testing-library/api.mdx
@@ -395,18 +395,37 @@ properties that are not defined are cleared.
 
 ```typescript
 const {rerender} = await render(Counter, {
-  componentProperties: {count: 4, name: 'Sarah'},
+  componentInputs: {count: 4, name: 'Sarah'},
 })
 
 expect(screen.getByTestId('count-value').textContent).toBe('4')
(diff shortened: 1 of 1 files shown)


docs/angular-testing-library/api.mdx
```diff
@@ -395,20 +395,39 @@ properties that are not defined are cleared.
 
 ```typescript
 const {rerender} = await render(Counter, {
-  componentProperties: {count: 4, name: 'Sarah'},
+  componentInputs: {count: 4, name: 'Sarah'},
```
MatanBobi, line 398: Is there an implementation of the `Counter` component somewhere in this page? I wasn't able to find it. Maybe it will help to add it?

docs/angular-testing-library/api.mdx
```diff
@@ -395,20 +395,39 @@ properties that are not defined are cleared.
...
+})
+
+expect(screen.getByTestId('count-value').textContent).toBe('4')
+expect(screen.getByTestId('name-value').textContent).toBe('Sarah')
+
+await rerender({componentInputs: {count: 7}, partialUpdate: true})
+
+// count is updated to 7
+expect(screen.getByTestId('count-value').textContent).toBe('7')
+// name is rendered because of the partial update
```
MatanBobi, line 427: Do we mean "Name is still rendered as `Sarah` because of the partial update"?

docs/angular-testing-library/api.mdx
timdeschryver: No, the implementation is "hidden" inside of the ATL repo.
I'll leave this for now. This comment makes me think if I should add a link to a working example (most of these already exist within ATL).
What's your opinion?

docs/angular-testing-library/api.mdx
timdeschryver: Correct. I'll update the description to make this more clear.
Thanks for the review!

timdeschryver, line 427:
```suggestion
// name is still rendered as "Sarah" because of the partial update
```

docs/angular-testing-library/api.mdx
MatanBobi: For the `Counter` component? Isn't it just for the example? I think it might help to have it here but if the component is big then a link might also work.

MatanBobi (approved):

PR #1375, opened 2024-03-23
docs: add FAQ for "not wrapped in act(...)" warning

JaKXz:
When I'm helping folks with react testing library, this is one of the most frequently asked questions I get 😅 

apologies if this is a duplicate request.


MatanBobi:
Hi @JaKXz, thanks for taking the time to open this one.
This question is already available in our React Testing Library FAQ 😀 

https://testing-library.com/docs/react-testing-library/faq

PR #1380, opened 2024-04-08
react: legacyRoot is only supported in React <= 18

eps1lon:
Docs for https://github.com/testing-library/react-testing-library/pull/1294

### DIFF
--- a/docs/react-testing-library/api.mdx
+++ b/docs/react-testing-library/api.mdx
@@ -107,6 +107,8 @@ your components.
 
 ### `legacyRoot`
 
+**WARNING**: This option is only available when tests run with React 18 and earlier.
+
 By default we'll render with support for concurrent features (i.e.
 [`ReactDOMClient.createRoot`](https://reactjs.org/docs/react-dom-client.html#createroot)).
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/api.mdx
```diff
@@ -107,6 +107,8 @@ your components.
 
 ### `legacyRoot`
 
+**WARNING**: This option is only available when tests run with React 18 and earlier.
```
MatanBobi, line 110: AFAIR and please correct me if I'm wrong, using `legacyRoot` wasn't available for versions before React 18. Also, I think a note here will be nicer in terms of UX.
```suggestion
:::note
This option is only available when tests run with React 18.
:::
```

eps1lon: Legacy root is everything before React 18. We don't officially support React 17 with latest RTL though. But this section is specifically for the React version you have installed not for the RTL version.

eps1lon: Also: this is not a note but a warning. Feel free to follow-up with an appropriate styling that's specific to our deploy setup.

PR #1391, opened 2024-05-08
docs: clarify fake timer usage with user-event

domnantas:
fixes https://github.com/testing-library/testing-library-docs/issues/1180

The docs do not clarify how to use fake timers together with `user-event`. Added a section with an example how to properly set it up.

### DIFF
--- a/docs/guides-using-fake-timers.mdx
+++ b/docs/guides-using-fake-timers.mdx
@@ -65,7 +65,8 @@ const user = userEvent.setup({advanceTimers: jest.advanceTimersByTime})
 ```
 
 :::caution
-You may find suggestions to set `delay: null` in `userEvent.setup()` in
-order to solve this issue. That is not recommended, as it may cause unexpected
-behaviour.
+
+You may find suggestions to set `delay: null` in `userEvent.setup()` in order to
+solve this issue. That is not recommended, as it may cause unexpected behaviour.
+
 :::
--- a/docs/guides-using-fake-timers.mdx
+++ b/docs/guides-using-fake-timers.mdx
@@ -13,11 +13,6 @@ code, most testing frameworks offer the option to replace the real timers in
 your tests with fake ones. This has a side effect - when using fake timers in
 your tests, _all_ of the code inside your test uses fake timers.
 
-This becomes apparent when using [`user-event`](user-event/intro.mdx) to test
-user interactions. Internally, it uses `setTimeout` to
-[`delay`](user-event/options.mdx#delay) subsequent actions and allow
-asynchronous code to run between the inputs.
-
 The common pattern to setup fake timers is usually within the `beforeEach`, for
 example:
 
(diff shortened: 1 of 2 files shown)


docs/guides-using-fake-timers.mdx
```diff
@@ -51,3 +51,22 @@ afterEach(() => {
   jest.useRealTimers()
 })
 ```
+
+## Using fake timers together with `user-event`
```
timdeschryver, line 55: Thanks for the PR @domnantas !
I do think that we shouldn't mention anything about user-event on this page though.
The reason for this is that this a "core" page.

For this PR the changes on this page can be reverted, and

- "This has a side effect - when using fake timers in
your tests, _all_ of the code inside your test uses fake timers." can be a note on this page
- the caution can be moved to https://testing-library.com/docs/user-event/options#advancetimers

Or do you see this differently @MatanBobi ?

MatanBobi: I 100% agree with you @timdeschryver :)

domnantas: Should I remove
````
`user-event` internally uses `setTimeout` to delay subsequent actions. Using
fake timers in such tests will cause timeouts, since the `setTimeout` callbacks
do not get executed. In order to allow `user-event` to advance the fake timers,
it is necessary to set [`advanceTimers`](user-event/options.mdx#advancetimers)
option in `userEvent.setup()`:

```js
const user = userEvent.setup({advanceTimers: jest.advanceTimersByTime})
```
````
section entirely, or move it somewhere else?

I understand "user-event" and "core" are separate scopes, but when tests break due to fake timers, the `Using fake timers` page in docs is the first thing I checked. Maybe there should at least a note in this page linking to the `advanceTimers` option?
```
:::note

Using fake timers together with `user-event` may cause test timeouts. See [`advanceTimers`](user-event/options.mdx#advancetimers) option for clarification.

:::
```

Comment (timdeschryver) [reply docs/guides-using-fake-timers.mdx:55]: I think it can be removed, or moved to the user-event docs at the advance timers option (but I think it already mentions it).

I like the note 👍.

Comment (domnantas) [reply docs/guides-using-fake-timers.mdx]: Simplified and moved the explanation to `user-event/options.mdx`. Left a short note on `guides-using-fake-timers`

File: docs/guides-using-fake-timers.mdx [hunk 2]
```diff
@@ -10,11 +10,8 @@ flaky.
 
 To solve these problems, or if you need to rely on specific timestamps in your
 code, most testing frameworks offer the option to replace the real timers in
-your tests with fake ones. This should be used sporadically and not on a regular
-basis since using it contains some overhead.
-
-When using fake timers in your tests, all of the code inside your test uses fake
-timers.
```
domnantas, line 17: I still stand by removing the `This should be used sporadically and not on a regular
basis since using it contains some overhead` sentence. Fake timers are nothing to be scared of when used correctly. And in some cases (let's say you're developing a calendar module) they need to be used in nearly all of your tests :)

docs/guides-using-fake-timers.mdx
```diff
@@ -26,12 +23,10 @@ beforeEach(() => {
 })
 ```
 
-When using fake timers, you need to remember to restore the timers after your
-test runs.
-
-The main reason to do that is to prevent 3rd party libraries running after your
-test finishes (e.g cleanup functions), from being coupled to your fake timers
-and use real timers instead.
```
domnantas, line 34: And here, I would argue 3rd party cleanup functions are not the main reason (though a valid one). Restoring timers is necessary to prevent fake timers leaking into other tests, where you don't expect them (especially with parallel and sharded tests!). So I clarified both reasons without singling out either of them as the "main" reason

docs/user-event/options.mdx
```diff
@@ -8,8 +8,19 @@ can be applied per [`setup()`](setup.mdx).
 
 ### advanceTimers
 
-If you are using fake timers, you need to advance your timers when we internally
-[delay](#delay) subsequent code.
+`user-event` adds a [delay](#delay) between some subsequent inputs. When using
```
MatanBobi, line 11: I think we're missing the exact example we had in the commit before because currently, it doesn't make a lot of sense. I'd remove all of the examples and only keep what we had before :)
```js
const user = userEvent.setup({advanceTimers: jest.advanceTimersByTime})
```

domnantas: I agree

MatanBobi: This looks much better to me, thanks.
I'll wait for @timdeschryver to respond as he also reviewed this one :)

timdeschryver (approved):
Thanks!

timdeschryver:
@all-contributors please add @domnantas for docs

PR #1392, opened 2024-05-14
feat/fix layout shifts

denisx:
### What you did:

Adds image dimensions to fix layout shifts

### What happened:

When images don't have dimensions, layout can shift at slow connection


src/pages/index.js
```diff
@@ -55,7 +60,11 @@ const HomeSplash = props => {
 
   return (
     <SplashContainer>
-      <Logo img_src={`${baseUrl}img/logo-large.png`} />
+      <Logo
+        img_src={`${baseUrl}img/logo-large.png`}
+        img_width={128}
```
MatanBobi, line 65: I prefer refactoring all of these to be the same as the other component (camelCased) instead of adding more props that don't follow normal conventions. Wdyt?

denisx: i make pr via local file conventions. because 1 pr = 1 feature. if you want, i can add refactoring, too. 
but it is not usual way. it is better to do new pr with refactoring (change to camel, and add linter to keep it in control)

MatanBobi: Since we're adding another prop here, I prefer seeing it done in this PR to follow the "boy scout rule". No need for a lint rule though :)

denisx: fix only my new code, or small refactoring for all this part?

denisx: "I prefer refactoring all of these" - ok

denisx: make update at pr

MatanBobi (approved):
Thanks @denisx :)

MatanBobi:
@all-contributors please add @denisx for code :)

PR #1394, opened 2024-05-21
fix: replace the TabItem values with general ones

Efim-Kapliy:
It will be easier for the user if you replace the TabItem values with general ones, because if you select a tab in one of the code examples, TypeScript, then other examples change asynchronously (i.e. the code examples will remain in JavaScript, or vice versa).

### DIFF
--- a/docs/react-testing-library/setup.mdx
+++ b/docs/react-testing-library/setup.mdx
@@ -31,10 +31,10 @@ make your test util file accessible without using relative paths.
 The example below sets up data providers using the [`wrapper`](api.mdx#wrapper)
 option to `render`.
 
-  <Tabs groupId="test-utils" defaultValue="jsx" values={[ {label: 'JavaScript',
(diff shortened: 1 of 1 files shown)


MatanBobi:
Thanks @Efim-Kapliy but I'm not sure I'm following. Currently if a user changes a tab, the rest change accordingly. What is this PR aiming to solve?

MatanBobi:
Thanks, I was under the impression that these already work.
What's the reason it's not working? The values were the same before (e.g. jsx/tsx), your change refactors it to be js/ts.

Efim-Kapliy:
Yes, that's right.

timdeschryver (approved):
I had to take 3-4 looks to see what changed, but this looks good to me! 
Thanks for the PR.

> Do you want to align it across the entire docs?

I checked and this page is the only place where we use these tab items (with ts(x) and js(x)) ~ the other pages use the name of the TL (react, angular, ...)

timdeschryver:
@all-contributors please add @Efim-Kapliy  for docs

PR #1410, opened 2024-08-19
docs(byrole.mdx): add tips for input type password

moeyashi:
from https://github.com/testing-library/dom-testing-library/issues/1128

### DIFF
--- a/docs/queries/byrole.mdx
+++ b/docs/queries/byrole.mdx
@@ -75,9 +75,7 @@ this functionality in your test or not is up to you.
 
 :::tip input type password
 
-Use [ByLabelText](queries/bylabeltext.mdx) to select `<input type="password" />`.
-
-This input type doesn't have the textbox role.
+Unfortunately, the spec defines that `<input type="password" />` has no implicit role. This means that in order to query this type of element we must fallback to a less powerful query such as `[ByLabelText](queries/bylabeltext.mdx)`.
 
 :::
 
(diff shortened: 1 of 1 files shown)


docs/queries/byrole.mdx
```diff
@@ -73,6 +73,14 @@ not replace the functionality of these attributes. For example
 its description if `fancy.jpg` could not be loaded. Whether you want to assert
 this functionality in your test or not is up to you.
 
+:::tip input type password
+
+Use [ByLabelText](queries/bylabeltext.mdx) to select `<input type="password" />`.
```
MatanBobi, line 78: Thanks!
I think it's worth to first explain the reason and then come with alternatives.
Something like:
```
Unfortunately, the spec defines that `<input type="password" />` has no implicit role. This means that in order to query this type of element we must fallback to a less powerful query such as `[ByLabelText](queries/bylabeltext.mdx)`.
```
Wdyt?

moeyashi: Thank you for the review.

To conclude, I don't find either approach particularly jarring.

I considered whether to put the reasoning before or after the conclusion, but in Japan, it is customary to present the conclusion first, so I followed that practice. (However, I’m not entirely confident that this is the best approach for documentation.)
Additionally, when I searched within the repository using :::tip, I found that imperative statements were used here, which also influenced my decision.
https://github.com/testing-library/testing-library-docs/blob/main/docs/svelte-testing-library/api.mdx?plain=1#L113-L117

moeyashi: > ByLabelText is only one solution.

> it's not just a decision we took as part of the library

I 100% agree.

I also think that the expression you proposed effectively includes these points and is very appropriate.

Thank you for your explanation.

MatanBobi: Thank you for taking the time to work on this.

MatanBobi:
@all-contributors please add @moeyashi for docs :)

PR #1412, opened 2024-08-20
fix(byrole.mdx): broken link

moeyashi:
I fixed broken the link that was added in https://github.com/testing-library/testing-library-docs/pull/1410 .

And applied formatting using Prettier.

## before

image

## after

image


timdeschryver (approved):
Good catch, thanks!

PR #1413, opened 2024-08-22
Update angular api

timdeschryver:
Update the Angular API:
- mark properties as deprecated
- document new properties

Closes https://github.com/testing-library/angular-testing-library/issues/476
Closes https://github.com/testing-library/angular-testing-library/issues/467

### DIFF
--- a/docs/angular-testing-library/api.mdx
+++ b/docs/angular-testing-library/api.mdx
@@ -76,62 +76,6 @@ export async function render<WrapperType = WrapperComponent>(
 
 ## Component RenderOptions
 
-### ~~`componentInputs`~~ (deprecated)
-
-An object to set `@Input` properties of the component. Uses `setInput` to set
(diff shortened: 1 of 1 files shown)


MatanBobi (approved):
Looks good 😀 
Do you think we should maybe add the version number from where the new APIs are available or is that common knowledge?

timdeschryver:
> Looks good 😀 Do you think we should maybe add the version number from where the new APIs are available or is that common knowledge?

It's new API's in Angular Testing Library. I don't want to add version numbers and keep the docs to the latest version. Otherwise it can become messy imho 😅.

MatanBobi:
> It's new API's in Angular Testing Library. I don't want to add version numbers and keep the docs to the latest version. Otherwise it can become messy imho 😅.

No worries, it was just an idea 😀

PR #1430, opened 2024-09-24
Remove contradiction from documentation

skonky:
I found something confusing in the documentation regarding testing-library-dom

[Here](https://testing-library.com/docs/user-event/install/) you say that testing-library-dom should be resolved via the framework wrapper 

And when you then go to the [actual framework wrapper](https://testing-library.com/docs/react-testing-library/intro) (in my case react) it states the following:

> To get started with React Testing Library, you'll need to install it together with its peerDependency @testing-library/dom:
`npm install --save-dev @testing-library/react @testing-library/dom`

but this contradicts what is said earlier right?

### DIFF
--- a/docs/react-testing-library/intro.mdx
+++ b/docs/react-testing-library/intro.mdx
@@ -9,11 +9,10 @@ APIs for working with React components.
 
 ## Installation
 
-To get started with `React Testing Library`, you'll need to install it together
-with its peerDependency `@testing-library/dom`:
+To get started with `React Testing Library`, you'll need to install it.
 
 ```bash npm2yarn
-npm install --save-dev @testing-library/react @testing-library/dom
+npm install --save-dev @testing-library/react
 ```
 
 ### With TypeScript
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/intro.mdx
```diff
@@ -9,11 +9,10 @@ APIs for working with React components.
 
 ## Installation
 
-To get started with `React Testing Library`, you'll need to install it together
-with its peerDependency `@testing-library/dom`:
+To get started with `React Testing Library`, you'll need to install it.
```
skonky, line 12: maybe this text can even be removed as it's kind of stating the obvious

MatanBobi:
Hi @skonky, thanks for this.
This is not true anymore for all of our framework wrappers. RTL for examples does not include DTL anymore.
I believe that this should be fixed in user-event.

skonky:
@MatanBobi Cool, I will update the docs to say the opposite of what I did in the current commit :)

PR #1432, opened 2024-10-05
feat(qwik-testing-library): add documentation for Qwik Testing Library

ianlet:
This PR adds documentation for the new [Qwik Testing Library](https://github.com/ianlet/qwik-testing-library) to provide more testing tools for the Qwik Framework.

The library was tested with a few projects running in production, but any additional feedback is greatly appreciated as well!

We will keep updating the documentation with more examples and upcoming APIs (such as better mocking utilities and `renderHook`) when they become available.

_Note: the easy way to setup the library with `npm run qwik add testing-library` is a [PR in the Qwik framework](https://github.com/QwikDev/qwik/pull/6924) and will soon be available. You need to follow the manual setup for now._

### DIFF
--- a/docs/qwik-testing-library/api.mdx
+++ b/docs/qwik-testing-library/api.mdx
@@ -14,14 +14,20 @@ sidebar_label: API
 
 ## `render`
 
-Render your component to the DOM with Qwik. By default, the component will be
-rendered into a `<host>` appended to `document.body`.
+Render your component to the DOM with Qwik. By default, when no options are
+provided, the component will be rendered into a `<host>` appended to
+`document.body`.
 
 ```tsx
 import {render} from '@noma.to/qwik-testing-library'
+import {MockProvider} from './MockProvider'
 import {MyComponent} from './MyComponent'
 
-const result = await render(<MyComponent />, renderOptions)
+const result = await render(<MyComponent />, {
+  baseElement: document.body,
+  container: document.createElement('host'),
+  wrapper: MockProvider,
+})
 ```
 
 ### Render Options
--- a/docs/qwik-testing-library/api.mdx
+++ b/docs/qwik-testing-library/api.mdx
@@ -1,0 +1,197 @@
+---
+id: api
+title: API
+sidebar_label: API
+---
+
+`@noma.to/qwik-testing-library` re-exports everything from
+[`@testing-library/dom`][@testing-library/dom], as well as:
+
+- [`render`](#render)
+- [`cleanup`](#cleanup)
+
+[@testing-library/dom]: ../dom-testing-library/api.mdx
+
+## `render`
+
+Render your component to the DOM with Qwik. By default, the component will be
+rendered into a `<host>` appended to `document.body`.
+
+```tsx
+import {render} from '@noma.to/qwik-testing-library'
+import {MyComponent} from './MyComponent'
+
+const result = await render(<MyComponent />, renderOptions)
+```
+
+### Render Options
+
+You may also customize how Qwik Testing Library renders your component. Most of
+the time, you shouldn't need to modify these options.
+
+| Option        | Description                                         | Default                          |
+| ------------- | --------------------------------------------------- | -------------------------------- |
+| `container`   | The container in which the component is rendered.   | `document.createElement('host')` |
+| `baseElement` | The base element for queries and [`debug`](#debug). | `document.body`                  |
+| `queries`     | [Custom Queries][custom-queries].                   | N/A                              |
+| `wrapper`     | The wrapper to provide a context to the component.  | N/A                              |
+
+[custom-queries]: ../dom-testing-library/api-custom-queries.mdx
+
+#### `wrapper`
+
+You can wrap your component into a wrapper to provide a context and other
+functionalities needed by the component under test.
+
+```tsx
+import {render} from '@noma.to/qwik-testing-library'
+import {QwikCityMockProvider} from '@builder.io/qwik-city'
+
+await render(<MyComponent />, {wrapper: QwikCityMockProvider})
+```
+
+### Render Results
+
+| Result                        | Description                                                |
+| ----------------------------- | ---------------------------------------------------------- |
+| [`baseElement`](#baseelement) | The base DOM element used for queries.                     |
+| [`container`](#container)     | The DOM element the component is mounted to.               |
+| [`asFragment`](#asFragment)   | Convert the DOM element to a `DocumentFragment`.           |
+| [`debug`](#debug)             | Log elements using [`prettyDOM`][pretty-dom].              |
+| [`unmount`](#unmount)         | Unmount and destroy the component.                         |
+| [`...queries`](#queries)      | [Query functions][query-functions] bound to `baseElement`. |
+
+[pretty-dom]: ../dom-testing-library/api-debugging.mdx#prettydom
+[query-functions]: ../queries/about.mdx
+
+#### `baseElement`
+
+The base DOM element that queries are bound to. Corresponds to
+`renderOptions.baseElement`. If you do not use `renderOptions.baseElement`, this
+will be `document.body`.
+
+#### `container`
+
+The DOM element the component is mounted in. Corresponds to
+`renderOptions.container`. If you do not use `renderOptions.container`, this
+will be `document.createElement('host')`. In general, avoid using `container`
+directly to query for elements; use [testing-library queries][query-functions]
+instead.
+
+#### `asFragment`
+
+Returns a `DocumentFragment` of your rendered component. This can be useful if
+you need to avoid live bindings and see how your component reacts to events.
+
+```tsx
+import {component$} from '@builder.io/qwik';
+import {render} from '@testing-library/react';
+import {userEvent} from "@testing-library/user-event";
+
+const TestComponent = component$(() => {
+  const count = useSignal(0);
+
+  return (
+    <button onClick$={() => (count.value++))}>
+      Click to increase: {count}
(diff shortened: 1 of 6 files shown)


timdeschryver (commented):
I'm not really up-to-date with this PR.

Did you verify bringing the documentation over to testing-library.com with someone @ianlet ?
I'm asking because the other libraries are all part of the @testing-library org (with SolidJS as an exception).

ianlet:
Thanks for the follow-up @timdeschryver! No, I did not verify bringing the documentation over to tesing-library.com before with someone as I was thinking this PR would open up the conversation. Sorry if I should have contacted someone before 😶

How do you usually proceed to bring in a new port of Testing Library to another framework?

timdeschryver:
Sorry for my late response @ianlet 
We've discussed this and are currently in the process of reviewing the library/docs

cc @MatanBobi

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,101 @@
...
+    await render(<LatestPostList />)
+
+    await waitFor(() =>
+      expect(screen.queryAllByRole('listitem')).toHaveLength(2),
+    )
+  })
+})
+```
+
+Notice how the mocked function is ending with `Qrl` instead of `$`, despite
```
MatanBobi, line 95: This seems to me like an implementation detail that we should maybe think about hiding from the user? Also, is this aligned with the qwik team? If this will change the entire mocking solution will break.

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,101 @@
...
+being named as `getLatestPosts$`. This is caused by the Qwik optimizer renaming
+it to `Qrl`. So, we need to mock the `Qrl` function instead of the original `$`
+one.
+
+If your `server$` function doesn't end with `$`, the Qwik optimizer might not
```
MatanBobi, line 100: That's a problem if we can't predict the behavior ahead of time and recommend how to mock this function. Do we have any contact in the qwik team to discuss this?

docs/qwik-testing-library/api.mdx
```diff
@@ -0,0 +1,197 @@
...
+## `render`
+
+Render your component to the DOM with Qwik. By default, the component will be
+rendered into a `<host>` appended to `document.body`.
+
+```tsx
+import {render} from '@noma.to/qwik-testing-library'
+import {MyComponent} from './MyComponent'
+
+const result = await render(<MyComponent />, renderOptions)
```
MatanBobi, line 24: I prefer seeing the entire options here and not an object which isn't defined yet (this is my preference at least). Wdyt?

MatanBobi (commented):
Thanks for this Ian.
I've left a few comments :)

Also, regarding the library itself, I recommend adding qwik (with the supported version) and DTL as peer dependencies.
I'd also think about automatic cleanup but that might be tricky to support in both vitest and jest.

docs/qwik-testing-library/example.mdx
ianlet: I double-checked with a core team member and actually the Qwik optimizer (what's doing the qwik magic behind the scenes) will always transform a function ending with the `$` dollar sign to `Qrl` and will leave the others untouched. This is true for both Qwik V1 and V2 (the upcoming major release).

So I can simply update the doc to state that the optimizer changes a function _only_ if it ends with `$`.

docs/qwik-testing-library/example.mdx
ianlet: I agree that it is inconvenient for the end user to have to think about this when mocking those kind of function.

For the context: this is how Qwik works and qwik devs are [aware of this](https://qwik.dev/docs/advanced/dollar/). In Qwik, this is not really considered as an implementation detail as devs can consciously choose one approach (`Qrl`) or the other `$`. It's just that the `$` is way more convenient and less boilerplate.

I'll try to think of a way to apply the same magic (`$` -> `Qrl`) when mocking entire modules. I'll keep you updated.

MatanBobi: Thanks.. This of course is not a blocker, just a comment :)

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,101 @@
...
+describe('<Counter />', () => {
+  // describe the test case
+  it('should increment the counter', async () => {
+    // render the component into the DOM
+    await render(<Counter />)
+
+    // retrieve the 'increment count' button
+    const incrementBtn = screen.getByRole('button', {name: /increment count/})
+    // click the button twice
+    await userEvent.click(incrementBtn)
```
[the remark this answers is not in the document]
ianlet: Considering [this note](https://testing-library.com/docs/user-event/setup#direct-apis), would you recommend as well to use `setup` explicitly?

docs/qwik-testing-library/setup.mdx
```diff
@@ -0,0 +1,159 @@
...
+If you prefer to set it up manually or want to understand what the setup script
+does, read along...
+
+## Manual Setup
+
+This module is distributed via [npm][npm] which is bundled with [node][node] and
+should be installed as one of your project's `devDependencies`:
+
+```bash npm2yarn
+npm install --save-dev @noma.to/qwik-testing-library
```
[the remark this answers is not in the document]
ianlet: I'll add `@testing-library/dom` :smile: About `qwik`, considering installing it manually requires quite some boilerplate, do you think it's worth documenting that approach as well?

ianlet:
> I'd also think about automatic cleanup but that might be tricky to support in both vitest and jest.

Would you consider this as a blocker for this PR? In any case I think we can reuse the same approach as RTL where they check if globals are set and use that to autocleanup (+ an env var to opt out). It should work for both vitest and jest as they use the same global API.

docs/qwik-testing-library/setup.mdx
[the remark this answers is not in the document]
MatanBobi: Nah, you're probably right here :) DTL will be enough

docs/qwik-testing-library/example.mdx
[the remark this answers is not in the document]
MatanBobi: Yeah... This is also mentioned at the bottom - 
> Using the methods on the instances returned by userEvent.setup() is recommended.

So we prefer following this recommendation in our docs :)

MatanBobi:
> @MatanBobi Would you consider this as a blocker for this PR? In any case I think we can reuse the same approach as RTL where they check if globals are set and use that to autocleanup (+ an env var to opt out). It should work for both vitest and jest as they use the same global API.

Definitely not a blocker :)

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,101 @@
...
+Below are some examples of how to use the Qwik Testing Library to tests your
+Qwik components.
+
+You can also learn more about the [**queries**][tl-queries-docs] and [**user
+events**][tl-user-events-docs] to help you write your tests.
+
+[tl-queries-docs]: ../queries/about.mdx
+[tl-user-events-docs]: ../user-event/intro.mdx
+
+## Qwikstart
```
timdeschryver, line 16: Is this pun intentional? (I love it 😂)

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,101 @@
...
+    await render(<Counter />)
+
+    // retrieve the 'increment count' button
+    const incrementBtn = screen.getByRole('button', {name: /increment count/})
+    // click the button twice
+    await userEvent.click(incrementBtn)
+    await userEvent.click(incrementBtn)
+
+    // assert that the counter is now 2
+    await waitFor(() => expect(screen.getByText(/count:2/)).toBeInTheDocument())
```
timdeschryver, line 42: We prefer using `findByText` in this case.
The result will be the same, but the intention is clearer.

timdeschryver (commented):
I think everything is covered 👍

docs/qwik-testing-library/example.mdx
ianlet: wow I wasn't aware `findBy` queries actually had a retry strategy! good to know, I'll update it :) thanks!

ianlet:
@MatanBobi @timdeschryver updated the PR with your comments

I also made a new release in the library itself to add support for auto cleanup and the ability to opt out with `QTL_SKIP_AUTO_CLEANUP` like the other libraries offer.

MatanBobi (approved):
This looks great to me, thanks Ian :) I'll leave it open for a couple of days to see if @timdeschryver wants to approve it too.

timdeschryver:
I will go over this tomorrow.
Thanks for your patience @ianlet , and your thorough review @MatanBobi

docs/qwik-testing-library/example.mdx
```diff
@@ -0,0 +1,103 @@
+---
+id: example
+title: Example
+sidebar_label: Example
+---
+
+Below are some examples of how to use the Qwik Testing Library to tests your
+Qwik components.
```
timdeschryver, line 8:
```suggestion
Below are some examples of how to use the Qwik Testing Library to test your
Qwik components.
```

timdeschryver (approved):
LGTM 👍
I just encountered a small typo, which I committed and resolved myself.

timdeschryver:
@all-contributors please add @ianlet for docs

ianlet:
Thank you @timdeschryver & @MatanBobi for your reviews and for merging! :raised_hands:

PR #1446, opened 2024-11-04
Update intro.mdx import

minsoo-web:
This page is based on v14, so I think it should use named import.

### DIFF
--- a/docs/user-event/intro.mdx
+++ b/docs/user-event/intro.mdx
@@ -64,7 +64,7 @@ itself - e.g. in a `before`/`after` hook - for reasons described in
 ["Avoid Nesting When You're Testing"](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).
 
 ```js
-import userEvent from '@testing-library/user-event'
+import { userEvent } from '@testing-library/user-event'
 
 // inlining
 test('trigger some awesome feature when clicking the button', async () => {


MatanBobi:
@ph-fritsche Any chance to get your look at this one? I remember we had some issues with that named import?

ahayes91:
Might be worth using this PR to update a few other places referencing v14 - https://github.com/testing-library/testing-library-docs/pull/1476/files has some examples!

ph-fritsche:
I'm not sure…

* Direct APIs will be deprecated and removed at some point.
* Preparing the document is necessary and giving users the option to call this (and cleanup callbacks) at any point might solve compatibility issues with other frameworks/tools. https://github.com/testing-library/user-event/pull/1270
* A test that reuses old pointer/keyboard state might be confusing. Therefore we should keep on recommending to call `setup()` inside the test. Using a setup function that combines this with render function of the respecting framework gives a clean and concise test code.

```ts
// current recommendation according to docs
import userEvent, {PointerEventsCheckLevel} from '@testing-library/user-event'
await user.setup({pointerEventsCheck: PointerEventsCheckLevel.Never}).click()

// also possible, but not officially recommended
import { userEvent, PointerEventsCheckLevel} from '@testing-library/user-event'
```

If we expose `prepare`, `reset` and `detach` with the next version, we could also expose `setup` as named export.
```ts
import {setup, PointerEventsCheckLevel} from '@testing-library/user-event'
// or if the user prefers
import * as userEvent from '@testing-library/user-event'
```

PR #1449, opened 2024-11-11
fix(a11y): underline links in a text block as pur WCAG 2.0

brianlu2610:
As per WCAG 2.0 guidelines, links must be visually distinguishable from normal text when they are in a text block without relying on color. This PR remediates that issue for the "Help" page.

Current Help Page in production: https://testing-library.com/help 

https://dequeuniversity.com/rules/axe/4.10/link-in-text-block?application=AxeChrome
https://www.w3.org/TR/WCAG20-TECHS/F73.html

image

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -3,15 +3,15 @@
 }
 
 [data-theme='dark']:root {
-  --ifm-color-primary: #7FA8F1;
-  --ifm-link-color: #7FA8F1;
+  --ifm-color-primary: #7fa8f1;
+  --ifm-link-color: #7fa8f1;
   --docsearch-muted-color: white;
 }
 
 [data-theme='light']:root {
-  --ifm-color-primary: #1554B7;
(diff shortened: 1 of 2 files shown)


MatanBobi:
Thanks :)
Shouldn't we just add the underline to every anchor element? Why did you choose to do it for a specific class? We can resolve this issue for the entire site that way.

brianlu2610:
We can definitely do that.

I chose to add a class because not all anchor elements need to be permanently underlined for the site to be accessible. ONLY those surrounded by a block of text would require it. Anchor elements by themselves only require to be underlined on hover, which is already default behavior.

I don't think there is any harm in underlining every link in the docs, but wanted to keep that optional.

MatanBobi:
Got you. Thanks! :)

MatanBobi:
@all-contributors please add @brianlu2610 for code :)

PR #1459, opened 2024-12-19
docs: Update Vue Testing Library Sidebar Order

aburdiss:
Match React Testing Library for better organization.

I noticed this today while I was reading through the Vue Testing Library docs for the first time, and noticed that the order of the docs in the sidebar was not the same as the React Testing Library docs that I look at all the time. I believe the order of the sidebar links should match across projects, and the rest of the projects and links do match as far as I can tell!


timdeschryver (approved):

timdeschryver:
@all-contributors please add @aburdiss  for docs

PR #1463, opened 2024-12-24
Add TypeScript Examples to vue-testing-library

aburdiss:
On the React Testing Library Docs, there are TypeScript examples, but these do not exist on most other pages.

I believe that it's useful for TypeScript examples to exist on other pages as well, especially examples, and even if they're extremely similar to the JavaScript examples, because this makes it more accessible to newer developers.

Disclosure: I'm newer to VueJS and TypeScript (coming from JS/React), and have been reading the docs myself to get this set up on my end, so if anyone notices better ways of handling these examples I'm not attached to the code inside.

### DIFF
--- a/docs/vue-testing-library/examples.mdx
+++ b/docs/vue-testing-library/examples.mdx
@@ -3,7 +3,15 @@ id: examples
 title: Example
 ---
 
+import Tabs from '@theme/Tabs'
+import TabItem from '@theme/TabItem'
+
 ## Basic example
+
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):
This looks good to me 👍
I'll leave this open a while longer so others can also review, otherwise i'll merge this next week.

MatanBobi:
Thanks @aburdiss!
@all-contributors please add @aburdiss for docs :)

PR #1476, opened 2025-03-19
Update docs to indicate named export should be used for `userEvent`

ahayes91:
As per https://github.com/testing-library/user-event/discussions/1205


ahayes91:
d'oh, looks like https://github.com/testing-library/testing-library-docs/pull/1446 has been open to do this on the intro page for a while too

timdeschryver:
@ahayes91 that for the contribution, but as you pointed out this is a duplicate.

PR #1479, opened 2025-04-02
Update docs with `reactStrictMode` in `render`

bernardobelchior:
Update docs with `reactStrictMode` in `render`, introduced in https://github.com/testing-library/react-testing-library/pull/1390.

### DIFF
--- a/docs/react-testing-library/api.mdx
+++ b/docs/react-testing-library/api.mdx
@@ -16,6 +16,7 @@ as these methods:
   - [`onRecoverableError`](#onrecoverableerror)
   - [`wrapper`](#wrapper)
   - [`queries`](#queries)
+  - [`reactStrictMode`](#render-options-reactstrictmode)
 - [`render` Result](#render-result)
   - [`...queries`](#queries-1)
   - [`container`](#container-1)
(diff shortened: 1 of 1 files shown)


docs/react-testing-library/api.mdx
```diff
@@ -31,7 +32,8 @@ as these methods:
   - [`initialProps`](#initialprops)
   - [`onCaughtError`](#oncaughterror)
   - [`onRecoverableError`](#onrecoverableerror)
-  - [`wrapper`](#wrapper-1)
```
bernardobelchior, line 34: This link was wrong

docs/react-testing-library/api.mdx
```diff
@@ -161,6 +163,11 @@ utility functions to create custom queries.
 Custom queries can also be added globally by following the
 [custom render guide](setup.mdx#custom-render).
 
+### `render` Options `reactStrictMode`
```
eps1lon, line 166:
```suggestion
### `reactStrictMode`
```

Isn't this redundant?

eps1lon (commented):
I'd also cross reference this option from `configure#reactStrictMode` to help discoverability.

docs/react-testing-library/api.mdx
bernardobelchior: Naming it only `reactStrictMode` would conflict with `reactStrictMode` from `configure`. If you prefer it another way, let me know.

bernardobelchior:
> I'd also cross reference this option from `configure#reactStrictMode` to help discoverability.

Done. Let me know if wording is fine or if there's any suggestion for improvement.

docs/react-testing-library/api.mdx
eps1lon: It's nested below "render Options" already so there shouldn't be a conflict. Or do you mean the anchors conflict? Can we decouple the anchor ID from the heading name?

This just seems noisy at the moment:
```
- render Options
   - container
   - baseElement
   - [...]
   - render Options reactStrictMode
```

eps1lon: Then let's keep it for now. Could even argue the repeated context is good on nested list with lots of siblings since it reduces cognitive overhead.

PR #1487, opened 2025-06-13
docs(svelte-testing-library): simplify examples with link to STL repository

mcous:
I just got through porting the Svelte examples to v5, and I went ahead and did it in the [STL repository itself](https://github.com/testing-library/svelte-testing-library/pull/440) so I could ensure that examples would be automatically tested (which caught [a bug](https://github.com/testing-library/svelte-testing-library/pull/439), which was cool).

In order to minimize the maintenance burden of the Svelte docs, this PR removes most examples from the `testing-library.com` docs in favor of linking to the STL repository for more complex examples.

I kept the basic example in place, because I think it's helpful to have _something_ here, and updated it to the same Svelte 5 example that now also lives in the STL repo.

### DIFF
--- a/docs/svelte-testing-library/example.mdx
+++ b/docs/svelte-testing-library/example.mdx
@@ -4,48 +4,51 @@ title: Example
 sidebar_label: Example
 ---
 
-For additional resources, patterns, and best practices about testing Svelte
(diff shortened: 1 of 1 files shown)


mcous:
Hey @timdeschryver! Yup this one is still good. Will tag you next time I need to put something up

PR #1494, opened 2025-08-14
Update about.mdx

megaaleksandar:
The Chrome extension is not available anymore, per `This extension is no longer available because it doesn't follow best practices for Chrome extensions.`

Hope this helps


MatanBobi:
Thanks @megaaleksandar!
We have an open issue about it and we do aim to resolve it but we currently don't have someone to handle the manifest change to v3. I'm closing this as we do want to keep the extension alive.

PR #1495, opened 2025-08-19
Update User Event docs on RTL requirement

jacklaurencegaray:


MatanBobi:
Thanks @jacklaurencegaray!
Actually, keeping the comment in user-event makes sense to me, maybe it's just worth emphasizing that we should make sure if DTL is already installed by one of the framework wrappers or not (and give an example of RTL which isn't installing it for you).

PR #1500, opened 2025-08-27
Add author name to recipes documentation

jvpiires:


timdeschryver (changes requested):
Please open an issue first with a description and why this change is needed before creating a PR.

PR #1502, opened 2025-09-04
Angular testing library docs update

alvaromartmart:
- Updates documentation links to point to the new Angular docs site
- Updates misleading example of `on` syntax
- Suggests using `imports` instead of `declarations` since standalone components/directives/etc. are the new default


timdeschryver (approved):
Looks great, thanks @alvaromartmart !

PR #1524, opened 2026-01-22
untangle syntax in intro.mdx

brandonzylstra:
The simplest correction would be this:

> [The Guiding Principles](/docs/guiding-principles) of this library emphasize a focus on tests that closely resemble how your web pages are interacted with by the users.

but I'm guessing that the "with" was avoided since some people prefer not to use that sort of syntax (even though it's actually perfectly normal and conventional).  So I'm proposing another alternative here that avoids putting two prepositions next to each other.


timdeschryver (approved):
Thanks @brandonzylstra !

timdeschryver:
@all-contributors please add @brandonzylstra  for docs

brandonzylstra:
> Thanks @brandonzylstra !

you're very welcome!  and thanks for accepting the PR!

PR #1532, opened 2026-04-01
docs: use queryByText in guide-disappearance.mdx

davidjb:
This adjusts the first example on this page to use the `queryBy` finder instead of `getBy` to avoid throwing in the query itself.

This gotcha gets explained a few times later in the doc (https://testing-library.com/docs/guide-disappearance#waiting-for-disappearance) but this is the first example you come to, and the use of `getByText` means the `expect(...).toBeInTheDocument()` will never be called.

### DIFF
--- a/docs/guide-disappearance.mdx
+++ b/docs/guide-disappearance.mdx
@@ -31,8 +31,10 @@ test('movie title appears', async () => {
   // element is initially not present...
 
   // wait for appearance inside an assertion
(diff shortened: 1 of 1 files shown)


davidjb:
Self-closing on account of the explanation at https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-presence-queries.md - the docs would benefit from distinguishing explicit choices on using `getBy`, assuming that applies here.

PR #1535, opened 2026-04-18
Add @accesslint/jest and @accesslint/vitest to the Ecosystem

ckundo:
Adds two new Ecosystem entries for the `@accesslint/*` accessibility-matcher packages. Both provide a `toBeAccessible()` matcher that works with any DOM produced by Testing Library (`container`, `screen.getByRole(...)`, etc.) — the same matcher composes across `@testing-library/react`, `@testing-library/vue`, `@testing-library/svelte`, and `@testing-library/angular`.

The libraries are synchronous, run under `jsdom` and `happy-dom`, and cover WCAG 2.2 Level A and AA via [`@accesslint/core`](https://github.com/AccessLint/accesslint/tree/main/core).

Follows the same page structure as `ecosystem-jasmine-dom` and `ecosystem-jest-native`.

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -218,6 +218,8 @@ module.exports = {
         'ecosystem-rtl-simple-queries',
         'ecosystem-testing-library-selector',
         'ecosystem-cli-testing-library',
+        'ecosystem-accesslint-jest',
+        'ecosystem-accesslint-vitest',
       ],
     },
   ],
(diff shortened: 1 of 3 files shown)


sidebars.js
```diff
@@ -218,6 +218,8 @@ module.exports = {
         'ecosystem-rtl-simple-queries',
         'ecosystem-testing-library-selector',
         'ecosystem-cli-testing-library',
+        'ecosystem-accesslint-jest',
+        'ecosystem-accesslint-vitest',
```
timdeschryver, line 222: What about creating a single page to add the general info + vitest and jest setups

